PR #25, opened 2024-09-17
Fix incorrect character in standard chars

bennetthardwick:
Spotted this little typo. I wonder if it is worth adding in a test that all of a certain ASCII code range is present.


coderdan:
Wow! Eagle eyes! Yes, good suggestion. I plan to add some property tests before this crate is released.

coderdan (approved):

PR #27, opened 2024-09-24
Aead trait

coderdan:
# Vitamin C powered encryption basics

AEAD: Authenticating Encryption with Associated Data

## Adds `vitaminc-aead`

* Aad types (Authenticated Associated Data) - `IntoAad` trait, impls for common types
* `Nonce` type and a `NonceGenerator` to mitigate against IV/Nonce reuse
* `Cipher` trait - `encrypt_vec`, `encrypt_slice` and `decrypt_vec` methods
* `CipherText` read/write "monads" and a `LocalCiphertext` type
* `Encrypt` and `Decrypt` traits and impls for common types

These traits are designed with inspiration from serde.
`Cipher` is loosely like `serde::Serializer` and `Encrypt`/`Decrypt` kinda like `Serialize`, `Deserialize`.

```rust
// Implements `Cipher`
let cipher = CoolCipher::new(...);
// Implements `Encrypt`
let plaintext = String::from("secret");
plaintext.encrypt(&cipher);
```

## Adds `vitaminc-encrypt`

* `Cipher` implementation using `aws-lc-rs`
* Super simple interface for common usage
* Property tests

```rust
let key = Key::from([0u8; 32]);
let encrypted = vitaminc_encrypt::encrypt(&key, "message").unwrap();
```

## Security notes

* Care taken to Zeroize sensitive values if an error occurs or if values are copied before encryption
* Limit any debugging of potentially sensitive values
* Using `Protected` types wherever possible/appropriate


packages/random/src/safe_rand.rs
```diff
@@ -6,8 +6,11 @@
 //! However, this implementation does not perform any zeroization and the authors of the `rand` crate
 //! have explictly [stepped away](https://github.com/rust-random/rand/issues/1358) from making it a "cryptographically secure" random number generator.
 use rand::{CryptoRng, RngCore, SeedableRng};
+use vitaminc_protected::Controlled;
+use zeroize::Zeroize;
 
 /// A secure random number generator that is safe to use for cryptographic purposes.
+// FIXME: RandChaCha20Rng is not zeroized and the crate doesn't support that feature
```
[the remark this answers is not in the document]
coderdan: Added https://github.com/cipherstash/vitaminc/issues/31

packages/aead/src/aad.rs
```diff
@@ -0,0 +1,173 @@
...
+    fn extend<T: IntoIterator<Item = u8>>(&mut self, iter: T) {
+        self.0.to_mut().extend(iter);
+    }
+}
+
+pub trait IntoAad<'a> {
+    fn into_aad(self) -> Aad<'a>
+    where
+        Self: Sized;
+}
```
freshtonic, line 51: What's the reason behind this trait instead of using Rust's `From` & `Into`?

packages/aead/src/ciphertext/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        let ciphertext = builder.ciphertext_mut().unwrap();
+        ciphertext.copy_from_slice(&encrypted);
+        let ciphertext = builder.append_tag(&[3u8; 16]).build();
+
+        assert_eq!(ciphertext.0.len(), 38);
+        assert_eq!(&ciphertext.0[..12], &[1u8; 12]);
+        assert_eq!(&ciphertext.0[12..22], &[2u8; 10]);
+        assert_eq!(&ciphertext.0[22..], &[3u8; 16]);
+    }*/
```
freshtonic, line 129: Commented out tests (or code in general) should include an explanation.

packages/aead/src/aad.rs
coderdan: 1. Ability to work in no-std environments (don't rely on `std::convert::From`
2. Keep some flexibility as the API is still in flux - it may not remain exactly the same as `From`
3. To be explicit about the lifetime

freshtonic (approved):

PR #39, opened 2025-08-09
feat: add protected-derive crate and opaque Debug implementation

coderdan:
Adds `debug` module to `protected` for opaque debugging of types.
Adds a trait, derive macro and convenience wrapper type.

```rust
#[derive(OpaqueDebug)]
struct Foo([u8; 32]);
let x = Foo([0; 32]);

assert_eq!(format("{x:?}"), "Foo (***)");
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1164,6 +1164,15 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


coderdan:
This has been superseded by #40.

PR #129, opened 2026-02-19
fix: remove circular dev-dependency between vitaminc-protected and vitaminc

coderdan:
## Summary

- The `TimingSafeEq` derive macro was generating code referencing `::vitaminc::protected::` paths, which forced `vitaminc-protected` to dev-depend on the `vitaminc` umbrella crate
- Since `vitaminc` depends on `vitaminc-protected`, this created a circular dependency that made publishing to crates.io impossible — `cargo publish` tries to resolve dev-deps against the registry and fails
- Changed the derive macro to emit `::vitaminc_protected::` paths and removed the unnecessary dev-dependency

### DIFF
--- a/packages/protected/Cargo.toml
+++ b/packages/protected/Cargo.toml
@@ -28,8 +28,6 @@ vitaminc-protected-derive = { path = "../protected-derive", version = "0.1.0-pre
 rmp-serde = "1.3.1"
 serde_json = "1.0.149"
 sha2 = "0.10.9"
-# NOTE: we need this for doctests (particularly the derive macros)
-vitaminc = { path = "../vitaminc", version = "0.1.0-pre4.1", features = [ "protected"] }
 
 [features]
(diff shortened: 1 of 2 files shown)


packages/protected/Cargo.toml
```diff
@@ -28,8 +28,6 @@ vitaminc-protected-derive = { path = "../protected-derive", version = "0.1.0-pre
 rmp-serde = "1.3.1"
 serde_json = "1.0.149"
 sha2 = "0.10.9"
-# NOTE: we need this for doctests (particularly the derive macros)
-vitaminc = { path = "../vitaminc", version = "0.1.0-pre4.1", features = [ "protected"] }
```
coderdan: No longer needed because doctests refer to `vitaminc_protected` instead.

CDThomas (approved):

PR #130, opened 2026-02-19
fix(aead): remove circular dev-dependency on vitaminc-encrypt

coderdan:
## Summary

- Removes the circular dev-dependency between `vitaminc-aead` and `vitaminc-encrypt` that prevented `cargo publish`
- Adds a feature-gated `test-utils` module with `TestCipher` — a no-op cipher that embeds AAD for round-trip verification
- Updates all doctests in `context.rs` and `README.md` to use `TestCipher` instead of `Aes256Cipher`

### DIFF
--- a/packages/aead/src/test_utils.rs
+++ b/packages/aead/src/test_utils.rs
@@ -1,0 +1,78 @@
+//! Test utilities for `vitaminc-aead`.
+//!
+//! Provides [`TestCipher`], a no-dependency cipher for use in doctests and
+//! examples. It is **not** cryptographically secure — it simply stores the
+//! plaintext alongside the AAD so that decryption can verify the AAD matches.
+
+use crate::{Cipher, IntoAad, LocalCipherText, Unspecified};
+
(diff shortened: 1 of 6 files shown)


packages/aead/src/test_utils.rs
```diff
@@ -0,0 +1,78 @@
+//! Test utilities for `vitaminc-aead`.
+//!
+//! Provides [`TestCipher`], a no-dependency cipher for use in doctests and
+//! examples. It is **not** cryptographically secure — it simply stores the
+//! plaintext alongside the AAD so that decryption can verify the AAD matches.
```
freshtonic, line 5: Nice!

I think I introduced the circular dependency initially, so I'm glad you came up with a workaround.

freshtonic (approved):
Good solution.

packages/aead/src/test_utils.rs
coderdan: Not sure but I definitely introduced a couple myself! The new release process picks up any issues immediately.

PR #148, opened 2026-03-28
feat: revised Cipher/Decipher traits with visitor pattern

coderdan:
## Summary

- Redesigns the `Cipher` and `Encrypt` traits with associated types and AAD support
- Adds `Decipher` trait with serde-style visitor pattern (`DecipherVisitor`, `SeqAccess`, `MapAccess`) and GAT `type Ok<T>` for future async support
- Adds `Decrypt` trait (counterpart to `Encrypt`) with impls for `Vec<u8>`, `String`, `u32`, `[u8; N]`, `Vec<T>`, `HashMap<String, T>`, and `Protected<T>`
- Introduces `AesCipherText` enum (`Single` | `Sequence` | `Map`) as the recursive encrypted output type
- Implements full encrypt/decrypt pipeline for `Aes256Cipher` including AAD support on both paths
- Adds `map_ok` to `Decipher` trait enabling `Decrypt` impls for wrapper types like `Protected<T>`

## Design notes

- **Visitor pattern**: Follows serde's `Deserializer`/`Visitor` separation — the `Decipher` reads encrypted data, the visitor constructs values
- **GAT on Decipher**: `type Ok<T>` allows sync (`Result<T, E>`) and async (`BoxFuture<'c, Result<T, E>>`) implementations
- **`map_ok`**: Required because GATs prevent generic transformation of `D::Ok<T>` to `D::Ok<U>` — acts as a functor `fmap`
- **Compound type**: `AesDecipher` (private) holds cipher ref + ciphertext + AAD; exposed via `Aes256Cipher::decrypt()` convenience method (like `serde_json::from_str`)
- **Error handling**: Visitor methods return `Result<Self::Value, Unspecified>` — no nesting with the GAT for sync impls

## Not yet implemented

- `encrypt_none` / `passthrough` / `Option<T>` (commented out, needs design)
- README examples (marked `ignore`, reference old API)

## Test plan

- [x] Property tests (quickcheck) for roundtrip: String, &str, u32, [u8; N], Vec<String>, Protected<String>
- [x] Property tests for AAD roundtrip and wrong-AAD/wrong-key rejection
- [x] `decrypt_fails_with_wrong_key` uses `DifferingKeyPair` — deterministic, cannot flake on a key collision
- [x] HashMap roundtrip test, plus HashMap-with-AAD roundtrip and wrong-AAD rejection
- [x] `cargo clippy` passes (including tests), on both the `aws-lc-rs` and RustCrypto backends
- [x] `cargo test --doc` passes for `vitaminc-aead` and `vitaminc-encrypt`

## Review follow-ups (PR #148)

Addressing @freshtonic's review — see the inline reply for the two items deliberately deferred (positional/key AAD binding is undesirable for the row-encryption use case; stack-array zeroize is tracked in #170). The rest:

- Dropped the dead `Decipher::Error` associated type — failures are always reported via `Unspecified` inside the `Ok` container
- Documented the `MapCipher::encrypt_key` `&'static str` restriction (`HashMap<String, T>` decrypts but does not encrypt) in the trait docs and the `aead` README
- `AesMapCipher::encrypt_key` now errors on a double call instead of silently dropping the first key
- Deleted the stale, uncompiled `context.rs` / `decrypt.rs` / `test_utils.rs` modules and the now-unused `test-utils` feature; re-implementing `ContextTag` on the revised traits is tracked in #178
- Restored `#![deny(unused_imports)]` on `vitaminc-encrypt`
- `TAG_LEN` is now `cfg`-gated to the RustCrypto backend instead of `#[allow(dead_code)]`
- `AesMapCipher::current_key` holds `&'static str`; removed redundant `Send` bounds on the `Vec`/`HashMap` decrypt visitors


freshtonic (commented):
## Review: Revised Cipher/Decipher traits with visitor pattern

Solid, well-documented redesign with a clear migration path and tracked follow-up issues. The serde mirroring is a sensible choice. Below are the items I'd want addressed before merge — not approving yet given the security points in particular.

### Security considerations

**1. Sequence/map structure is not authenticated — reordering & swapping attacks**

`Vec<T>::encrypt_with_aad` encrypts every element with the *identical* AAD, and `HashMap` encrypts every value with the *identical* AAD. Each element gets a fresh nonce (good), but nothing binds an element to its **position** or a value to its **key**.

For an attacker who can tamper with an `AesCipherText`:
- **Reorder** elements of a `Sequence` — every element still authenticates, decryption succeeds silently.
- **Swap values between map keys** (e.g. trade the ciphertext under `"name"` and `"city"`) — undetected.
- **Relabel map keys** — keys are stored in cleartext *and* never fed into AAD, so they are neither encrypted nor authenticated.
- **Truncate** a `Sequence` or drop `Map` entries — `VecVisitor`/`HashMapVisitor` loop until exhaustion, so a shorter collection comes back with no error.

Recommendation: fold positional/key context into the per-element AAD. The `IntoAad` tuple support exists for exactly this — encrypt sequence element *i* with `(user_aad, i)` and map values with `(user_aad, key)`, making reorder/swap/relabel detectable. Worth deciding explicitly whether truncation must also be rejected (would need an authenticated length/count).

**2. Stack key material / arrays no longer zeroized (regression)**

The old `[u8; N]` `Encrypt` impl explicitly `self.zeroize()`'d the source array after `to_vec()`. The new path (`Encrypt for [u8; N]` → `Cipher::encrypt_bytes_array` default → `data.to_vec()`) copies the bytes into a `Vec` and drops the original `[u8; N]` without zeroizing. This also hits `Key::encrypt_with_aad`, which copies raw 32-byte key material to a heap `Vec` with the stack copy left un-zeroized.

The code comments acknowledge this and link #170, so it's tracked — but it's a real regression from current `main` behaviour and matters for key-wrapping. Flagging so the merge decision is deliberate.

### Correctness / design

- **`Decipher::Error` is effectively dead.** Visitor methods hardcode `Result<_, Unspecified>` and `SeqAccess`/`MapAccess` callers do `.map_err(|_| Unspecified)`. Either drop the associated type or wire it through. Note the visitor API is fundamentally synchronous (`visit_*` returns a plain `Result`), which may constrain the async story the `Ok<T>` GAT is being added to support — worth confirming the async impl can thread through this.
- **`MapCipher::encrypt_key` only accepts `&'static str`.** `HashMap<String, T>` (runtime-derived keys) can be *decrypted* but not *encrypted* — only `HashMap<&'static str, T>` implements `Encrypt`. The README advertises `HashMap<String, T>: Decrypt` without noting the encrypt-side restriction. Intentional? If so, document it.
- **`encrypt_key` called twice silently drops the first key** — it overwrites `current_key` without checking for a pending un-valued key. A debug assertion or error would catch trait-contract misuse.
- **`decrypt_fails_with_wrong_key` is theoretically flaky.** The old code used `DifferingKeyPair` to guarantee distinct keys; this relies on two random keys differing. The inline comment ("if they happen to match... which is fine") is misleading — the test returns a bool and *would* fail on a collision. Probability is negligible, but correct the comment or retain `DifferingKeyPair`.

### Code quality / conventions

- **Commented-out `mod` declarations in `aead/src/lib.rs`** (`//mod context;`, `//mod decrypt;`, `//mod test_utils;`) leave stale uncompiled source files in the tree. `encrypt.rs` was properly deleted — these should be too, or kept compiling.
- **`#![deny(... unused_imports)]` was dropped** from `encrypt/src/lib.rs`. If transitional, prefer a targeted `#[allow]` over weakening the crate-wide lint.
- **`TAG_LEN` is now `#[allow(dead_code)]`** — if unused, remove it rather than annotating.
- Minor: `AesMapCipher::current_key` stores `String` though the key is `&'static str` (unnecessary allocation); the `Vec<T>`/`HashMap` decrypt visitors add a redundant `Send` bound already implied by the `Decrypt` supertrait.

### Test coverage

Good property-test coverage of happy-path roundtrips plus AAD and wrong-key negatives. Gaps:

- **Doc-tests not in the test plan** (lists property tests + `cargo clippy` only). Several README examples became `ignore`/`no_run`, but the `aead` README "Custom Types" `User` example appears to remain a live ` ```rust ` doctest with substantially rewritten content. Given the repo's recent run of "doctest fixes" commits, please confirm `cargo test --doc` passes for both crates.
- No negative test for the structural-integrity gap above — once AAD binding is added, add tests that reorder a `Sequence` / swap `Map` values and assert decryption fails.
- No HashMap-with-AAD roundtrip test.

coderdan:
Thanks @freshtonic. A few comments inline:

> **1. Sequence/map structure is not authenticated — reordering & swapping attacks**
> 
> `Vec<T>::encrypt_with_aad` encrypts every element with the _identical_ AAD, and `HashMap` encrypts every value with the _identical_ AAD. Each element gets a fresh nonce (good), but nothing binds an element to its **position** or a value to its **key**.

This is in fact desirable for the target use-case: `Vec<T>` where `T` is a row in a database table. Encoding insertion order in the AAD would prevent decryptions when rows are retrieved (inevitably!) out of order.

I can see value in adding an `StrictOrder<T>(Vec<T>)` wrapper type which implements the per-element AAD suggested but that would serve a different goal and I think could be done as a follow-up.

> **2. Stack key material / arrays no longer zeroized (regression)**
> 
> The old `[u8; N]` `Encrypt` impl explicitly `self.zeroize()`'d the source array after `to_vec()`. The new path (`Encrypt for [u8; N]` → `Cipher::encrypt_bytes_array` default → `data.to_vec()`) copies the bytes into a `Vec` and drops the original `[u8; N]` without zeroizing. This also hits `Key::encrypt_with_aad`, which copies raw 32-byte key material to a heap `Vec` with the stack copy left un-zeroized.
> 
> The code comments acknowledge this and link #170, so it's tracked — but it's a real regression from current `main` behaviour and matters for key-wrapping. Flagging so the merge decision is deliberate.

That's valid and was left as a follow-up to limit PR size.

The remaining concerns are valid and I'll address now.

freshtonic (approved):
@coderdan has addressed the relevant feedback and justified the mis-diagnosed issues. Approved.

PR #149, opened 2026-03-31
feat(aead): add IntoAad impls for [u8; N] and &[u8; N]

coderdan:
## Summary
- Add `IntoAad` impl for owned `[u8; N]` (uses `Aad::new_owned`)
- Add `IntoAad` impl for borrowed `&[u8; N]` (zero-copy via `Aad::from_slice`)
- `Option<[u8; N]>` works automatically via the existing blanket impl

## Test plan
- [x] Added unit tests for `[u8; N]`, `&[u8; N]`, and `Option<[u8; N]>`
- [x] All 32 aead tests pass

### DIFF
--- a/packages/aead/src/aad/mod.rs
+++ b/packages/aead/src/aad/mod.rs
@@ -76,6 +76,18 @@ impl<'a> IntoAad<'a> for Vec<u8> {
     }
 }
 
+impl<'a, const N: usize> IntoAad<'a> for [u8; N] {
+    fn into_aad(self) -> Aad<'a> {
+        Aad::new_owned(self)
+    }
+}
+
+impl<'a, const N: usize> IntoAad<'a> for &'a [u8; N] {
+    fn into_aad(self) -> Aad<'a> {
+        Aad::from_slice(self.as_slice())
+    }
+}
+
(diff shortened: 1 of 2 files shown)


packages/aead/src/aad/mod.rs
```diff
@@ -191,6 +203,30 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_option_byte_array_aad() {
+        let some_aad = Some([1u8, 2, 3]).into_aad();
+        let expected = Aad::pae(&[&[1, 2, 3]]);
+        assert_eq!(some_aad.as_bytes(), expected.as_bytes());
+
+        let none_aad = Option::<[u8; 3]>::None.into_aad();
+        assert_ne!(none_aad.as_bytes(), some_aad.as_bytes());
```
yujiyokoo, line 229: No big deal but it might help the reader if there's another assertion here to check that the result is `&[]` (which I am assuming is the case but not 100% certain, hence this comment)

yujiyokoo (dismissed):

packages/aead/src/aad/mod.rs
coderdan: Good suggestion!

PR #163, opened 2026-05-10
feat(encrypt): add wasm32 support via RustCrypto backend

coderdan:
`vitaminc-encrypt` has been unusable on `wasm32-unknown-unknown` because its only backend is `aws-lc-rs`, whose `aws-lc-sys` C deps don't cross-compile to wasm. That blocks any downstream crate from building for wasm even when it only consumes the `Aad`/`IntoAad` types — which is currently making cipherstash-suite take a workaround feature-gate detour.

## Approach: cfg-based dual backend, no feature flag

| Target | Backend |
|---|---|
| `cfg(not(target_arch = "wasm32"))` | `aws-lc-rs` — current behaviour, FIPS-validated, AES-NI accelerated. **No native perf change.** |
| `cfg(target_arch = "wasm32")` | `aes-gcm` from RustCrypto — pure Rust, builds without a C toolchain |

Both implement RFC 5116 AES-256-GCM with the same wire format (`ciphertext \|\| 16-byte tag`), so existing ciphertexts decrypt cleanly under either backend. No data migration.

## Shape

A small `backend/` module exposes a single `CipherKey` type with `seal_in_place_append_tag` / `open_in_place`. `cipher.rs` and `key.rs` were thinned to use it — they no longer reference any specific crypto library.

`vitaminc-random` gained a wasm32-only `getrandom = { features = ["wasm_js"] }` target dep plus a `.cargo/config.toml` rustflag (`--cfg getrandom_backend="wasm_js"`). Without these, `rand 0.10` → `getrandom 0.4` fails to compile on wasm32 because no entropy backend is selected.

## Verification

- Native test suite (`cargo test -p vitaminc-encrypt`) — 10 unit + 5 doc tests pass unchanged
- `cargo check --target wasm32-unknown-unknown -p vitaminc-encrypt` — clean
- `cargo clippy --target wasm32-unknown-unknown -p vitaminc-encrypt --tests` — clean
- `mise run clippy` (native, all features) — clean

Wasm32 *runtime* round-trip testing isn't part of this PR — that's most cleanly exercised downstream where wasm targets are actually deployed.

## Why now

cipherstash-suite is adding wasm support so `cipherstash-client` can run in Supabase Edge Functions. The first PR there has a workaround feature-gate (`cts-common/aead`) that exists purely because vitaminc-encrypt didn't build on wasm. Once a release containing this lands, that workaround can be reverted.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,3 +69,14 @@ jobs:
 
       - name: build check (vitaminc-encrypt, wasm32)
         run: cargo check --target wasm32-unknown-unknown -p vitaminc-encrypt
+
+      - name: install wasm-pack
+        uses: jetli/wasm-pack-action@v0.4.0
+        with:
+          version: 'latest'
+
+      - name: test (vitaminc-encrypt, wasm32 in Node)
+        # Compiles to wasm32-unknown-unknown and runs tests inside Node — this
+        # is what gates the *actual* wasm codegen path (the native
+        # `_test-rust-crypto-backend` step above is just a proxy).
+        run: wasm-pack test --node packages/encrypt
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,6 +34,7 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy, rustfmt
+          targets: wasm32-unknown-unknown
 
       - name: Start LocalStack
         uses: LocalStack/setup-localstack@v0.3.2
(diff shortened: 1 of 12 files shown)


packages/encrypt/src/backend.rs
```diff
@@ -0,0 +1,30 @@
...
+//! - **Native** (`cfg(not(target_arch = "wasm32"))`): `aws-lc-rs` —
+//!   FIPS-validated, hardware-accelerated.
+//! - **Wasm32** (`cfg(target_arch = "wasm32")`): `aes-gcm` from RustCrypto —
+//!   pure Rust, builds without a C toolchain.
+//!
+//! Both implementations conform to RFC 5116 AES-256-GCM and produce
+//! byte-identical ciphertext+tag output for the same inputs.
+//!
+//! The internal `_test-rust-crypto-backend` feature forces the RustCrypto
+//! backend on native targets so CI can exercise it without wasm tooling.
```
tobyhede, line 14: Are there any tests that validate that output is byte-identical

.github/workflows/test.yml
```diff
@@ -43,11 +44,28 @@ jobs:
...
+      - name: test (default — aws-lc-rs backend)
         run: cargo test
+
+      - name: test (vitaminc-encrypt, RustCrypto backend on native)
+        # Forces the RustCrypto path that wasm32 uses, so the same property
+        # tests run against both backends in CI.
+        run: cargo test -p vitaminc-encrypt --features _test-rust-crypto-backend
+
+      - name: build check (vitaminc-encrypt, wasm32)
+        run: cargo check --target wasm32-unknown-unknown -p vitaminc-encrypt
```
tobyhede, line 71: The proxying does not quite test that actual wasm codegen runs the tests

Might be worth testing the full generation: `wasm-pack test --node`

compiles your Rust crate to wasm32-unknown-unknown and runs its tests inside Node.js

packages/encrypt/Cargo.toml
```diff
@@ -12,13 +12,33 @@ keywords.workspace = true
...
 
 zeroize = { workspace = true }
 
+# Native: aws-lc-rs (FIPS-validated, AES-NI accelerated).
+# `aes-gcm` is normally only used on wasm32, but is exposed here as an
+# optional dep so the `_test-rust-crypto-backend` feature can pull it in for
+# CI testing of the wasm32 backend on a native runner.
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+aws-lc-rs = "1.16.3"
+aes-gcm = { version = "0.10", default-features = false, features = ["aes", "alloc"], optional = true }
```
tobyhede: Claude tells me `aes-gcm` has a `zeroize` flag that might be worth turning on

tobyhede (changes requested):
Looks good so far.
Adding some cross-backend parity tests would validate the primary goal.

.github/workflows/test.yml
coderdan: Good idea.

packages/encrypt/src/backend.rs
coderdan: I misunderstood when I first read this. I realise now you're mean a test that validates a ciphertext encrypted via the WASM backend decrypts using the AWS-LC backend and vice verse.

Added a test with Known-Answer values to verify.

tobyhede (approved):

PR #164, opened 2026-05-10
fix(permutation): close cache-timing side-channel in permute_array

coderdan:
## Summary

- `permute_array`/`depermute_array` indexed the array by a secret permutation-key byte (`out[i] = input[k]`, `out[k] = input[i]`), so CPU cache-line accesses leak key bits to a co-resident observer — same class of attack that broke software AES T-table implementations.
- Replaced the secret-indexed load/store with a constant-time scan that touches every input/output position in fixed order, using `subtle::ConditionallySelectable` for per-element select. The memory access pattern is now driven only by public `N`.
- Trait bound on `T` widens to include `ConditionallySelectable`; all current `IsPermutable` element types (`u8`/`u16`/`u32`) already implement it via subtle.
- Secret locals (`input`, `kv`, `selected`, `src`) wrapped in `Zeroizing<_>` so wipes run via `Drop` on every exit including unwind.

## Discovery

Found via Trail of Bits' `constant-time-analysis` skill. The instruction-level analyzer passed cleanly on both arm64 and x86_64 (no `DIV`/branch-on-secret), but it explicitly does not detect cache-line access patterns — this leak surfaced in the manual review pass that follows the analyzer.

## Performance impact

Measured at `-O3` on arm64 (M-series), `[u8; N]`, key bytes pre-extracted, 500 K iterations:

| N | old (variable-time) | new (constant-time scan) | overhead |
|---|---|---|---|
| 8 | 6.5 ns | 66 ns | 10× |
| 16 | 4.8 ns | 159 ns | 33× |
| 32 | 7.5 ns | 733 ns | 98× |
| 64 | 17 ns | 3.2 µs | 189× |
| 128 | 39 ns | 12.6 µs | 325× |

Expected: `O(N)` indexed loads → `O(N²)` conditional moves. Acceptable for current call sites (key inversion, complement, single-block permutes); the absolute cost is sub-microsecond up to N=32.

## Follow-up

Not in this PR — `tbl`/`pshufb` SIMD specialization for performance-sensitive sizes (single-instruction byte shuffle, documented constant-time on every CPU that ships them) would close most of the gap.

The same secret-indexed-lookup pattern exists in `bitwise.rs:21` (`arr.get_unchecked(k)`) but is benign: bit-arrays there are 1–16 bytes (≤1 cache line). Hardened in #165 as defense-in-depth.

## Test plan

- [x] `cargo test -p vitaminc-permutation` — 6 unit + 3 doctests pass (round-trip, depermutation, associativity, key inversion, complement at N=8/16/32/64)
- [x] `cargo build --workspace` — clean
- [x] `cargo fmt --check` — clean
- [x] ct_analyzer on release asm (arm64 + x86_64) — PASSED, 0 errors / 0 warnings
- [x] zeroize-audit — wipes intact in LLVM IR at O3 (10 volatile-zero stores, vs 8 at O0)
- [x] CI green

### DIFF
--- a/packages/permutation/src/elementwise.rs
+++ b/packages/permutation/src/elementwise.rs
@@ -1,4 +1,5 @@
 use crate::{private::IsPermutable, PermutationKey};
+use subtle::{ConditionallySelectable, ConstantTimeEq};
 use vitaminc_protected::{Controlled, Zeroed};
 use zeroize::Zeroize;
 
@@ -15,7 +16,7 @@ pub trait Depermute<T> {
 /// Implement permutation for Protected type containing a permutable array.
 impl<const N: usize, T> Permute<[T; N]> for PermutationKey<N>
 where
-    T: Zeroize + Default + Copy,
+    T: Zeroize + Default + Copy + ConditionallySelectable,
     [T; N]: IsPermutable + Zeroed,
 {
     fn permute(&self, input: [T; N]) -> [T; N] {
@@ -25,7 +26,7 @@ where
 
 impl<const N: usize, T> Depermute<[T; N]> for PermutationKey<N>
 where
-    T: Zeroize + Default + Copy,
+    T: Zeroize + Default + Copy + ConditionallySelectable,
     [T; N]: IsPermutable + Zeroed,
 {
     fn depermute(&self, input: [T; N]) -> [T; N] {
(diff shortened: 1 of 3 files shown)


coderdan:
Folded in `9bdec01`: wrap secret locals (`input`, `kv`, `selected`, `src`) in `Zeroizing<_>` so the wipes are performed by `Drop` and survive unwind paths. Removes the redundant explicit `.zeroize()` calls.

Verified with the [trailofbits zeroize-audit skill](https://github.com/trailofbits/skills): volatile-zero stores in LLVM IR are intact at O3 (10 stores, vs 8 at O0 — i.e. not reduced). The ct_analyzer still passes on arm64. Permutation correctness tests pass.

packages/permutation/src/elementwise.rs
```diff
@@ -34,42 +35,49 @@ where
...
+    // `ConditionallySelectable` so the access pattern is independent of `kv`,
+    // preventing cache-line timing leaks of the secret key bytes. Secret
+    // locals are wrapped in `Zeroizing` so they are wiped on any unwind path.
+    let input = Zeroizing::new(input);
+    let mut out: [T; N] = Zeroed::zeroed();
+    for (i, k) in key.iter().enumerate() {
+        let kv = Zeroizing::new(k.risky_unwrap());
+        let mut selected = Zeroizing::new(input[0]);
+        for (j, src) in input.iter().enumerate().skip(1) {
+            let mask = (j as u8).ct_eq(&*kv);
```
tobyhede, line 60: `j as u8` is correct for al current `IsPermutable` sizes, but would silently wrap.

Consider adding assertion `const _: () = assert!(N <= 256, "...")`

packages/permutation/src/elementwise.rs
```diff
@@ -34,42 +35,49 @@ where
...
-
-    // We copied all elements to the output so we should zeroize the input
-    input.zeroize();
+    // Constant-time scan: for each output position `i`, the key byte `kv`
+    // selects which input element to copy. We scan all `j` in 0..N and use
+    // `ConditionallySelectable` so the access pattern is independent of `kv`,
+    // preventing cache-line timing leaks of the secret key bytes. Secret
+    // locals are wrapped in `Zeroizing` so they are wiped on any unwind path.
+    let input = Zeroizing::new(input);
+    let mut out: [T; N] = Zeroed::zeroed();
```
tobyhede: `out` is out is not re-zeroized on unwind. 
Mostly theoretical - loop operations would need to trigger panic/unwind.

tobyhede (approved):
Looks good.
Worthing considering a regression test - no existing test would fail if the constant-time fix were silently reverted.

coderdan:
> Looks good. Worthing considering a regression test - no existing test would fail if the constant-time fix were silently reverted.

Not really possible to do timing checks in regular tests. Have added #166 to run constant time analyzer in CI. Its a bigger piece of work but worth doing.

Issue #173, opened 2026-05-15
aead: implement Encrypt for Option<T>


coderdan:
`impl<T: Encrypt> Encrypt for Option<T>` is currently a commented-out stub in `packages/aead/src/encrypt/impls.rs`. It is blocked on `Cipher::encrypt_none` / `Cipher::passthrough` (see cipherstash/vitaminc#171).

When unblocked, `Some(v)` should encrypt the inner value and `None` should produce a distinguishable, authenticated "no value" output so a decryptor can recover an `Option<T>` unambiguously.

Tracked from PR [#148](<https://github.com/cipherstash/vitaminc/issues/148>).

Issue #172, opened 2026-05-15
encrypt: implement encrypt_none and passthrough on Aes256Cipher


coderdan:
Once `Cipher::encrypt_none` and `Cipher::passthrough` are introduced (see cipherstash/vitaminc#171), the concrete `Aes256Cipher` impl in `packages/encrypt/src/cipher.rs` needs to provide both. Commented-out signatures are already in place.

Tracked from PR [#148](<https://github.com/cipherstash/vitaminc/issues/148>).

Issue #171, opened 2026-05-15
aead: implement encrypt_some, encrypt_none, and passthrough on Cipher


coderdan:
`Cipher` in `packages/aead/src/cipher.rs` should grow three additional entry points:

- `encrypt_some<T: Encrypt>(self, value: T)` — the optional-present case
- `encrypt_none(self)` — the optional-absent case (must still authenticate "no value" distinguishably from passthrough)
- `passthrough<T: 'static>(self, data: T)` — pass an unencrypted value through the cipher's output container

The commented-out sketch lives in the same file. The first two are required to unblock `impl Encrypt for Option<T>` (currently a commented-out stub in `packages/aead/src/encrypt/impls.rs`).

Tracked from PR #148.

PR #174, opened 2026-05-16
feat(aead, encrypt): add encrypt_some/encrypt_none/passthrough

### LINKED ISSUE #170: aead: zeroize byte-array plaintext when forwarded through encrypt_bytes_array default
The default impl of `Cipher::encrypt_bytes_array` in `packages/aead/src/cipher.rs` calls `.to_vec()` on the input array before forwarding to `encrypt_bytes_vec`. We should verify what `to_vec` does internally and ensure the original stack array is zeroized before drop — otherwise sensitive bytes may linger on the stack after the cipher returns.

See `packages/aead/src/cipher.rs` (in `Cipher::encrypt_bytes_array`).

Tracked from PR #148.

coderdan:
Builds on top of #148. Closes the three deferrals tracked in #171, #172, #173.

## Summary

- `Cipher::encrypt_some` (default forwarder), `Cipher::encrypt_none` (required), `Cipher::passthrough<T: Any + Send + 'static>` (required)
- `SeqCipher::passthrough_next`, `MapCipher::passthrough_entry`
- `Decipher::decrypt_option<T>` and `decrypt_passthrough<T>` — typed direct accessors that bypass the visitor (target type is already concrete at the call site)
- `Encrypt` / `Decrypt` for `Option<T>`
- `AesCipherText` gains `None(LocalCipherText)` (sealed empty plaintext, AAD bound to the tag) and `Passthrough(Box<dyn Any + Send + 'static>)`

## Design notes

- **`None` is cryptographically authenticated**, not just a structural marker — seals an empty plaintext under the supplied AAD so the absent case cannot be forged or migrated between contexts
- **Visitor trait left unchanged** — `decrypt_option` and `decrypt_passthrough` are direct typed methods on `Decipher` because the target type is concrete at the call site, so dispatch via a visitor would only add stub methods to every existing visitor
- **`Box<dyn Any>` for passthrough** is documented as static-at-the-API-boundary with a runtime downcast on decryption. The earlier spike at `Spikes/protect-rs-spike` tried `erased_serde` and a fully-static cons-list — both were shelved. Trade-off: `AesCipherText` containing a `Passthrough` is not `Serialize` — a follow-up will be needed before it can be persisted across processes
- **Type-laundering guards**: `decrypt_option` rejects `Passthrough`, and `decrypt_passthrough` rejects every non-passthrough variant. The existing `_ => Err` arms in `decrypt_bytes` / `decrypt_seq` / `decrypt_map` already reject `None` and `Passthrough`

## Deviation from the design

- `Cipher::encrypt_none` and `Cipher::passthrough` are **required** rather than defaulted to `Err`. `Self::Error` is an associated type so a generic default can't construct one. Future mock ciphers will need stubs

## Protected<T> chain of custody (def6668)

Pushed `Protected<T>` — the workspace's chain-of-custody primitive — through the `Cipher` / `DecipherVisitor` trait surface so the protection guarantee is enforced at the choke-point rather than at every leaf. Driven by the zeroize audit (see Validation below).

- `Cipher::encrypt_bytes_vec` / `encrypt_bytes_array` take `Protected<Vec<u8>>` / `Protected<[u8; N]>`
- `DecipherVisitor::visit_bytes_vec` takes `Protected<Vec<u8>>`
- `AesDecipher::decrypt_local_ciphertext` returns `Protected<Vec<u8>>` — no trailing `risky_unwrap`
- `Key::encrypt_with_aad` passes `self.0` straight through — the 32-byte key never exists as a bare `[u8; 32]` at the trait boundary
- The default `encrypt_bytes_array` impl uses `risky_ref` so the outer `Protected<[u8; N]>` stays owned and is wiped on drop (resolves #170)

Passthrough's `T: Any + Send + 'static` bound is **deliberately not tightened** to add `Zeroize` — passthrough is for non-sensitive data (schema tags, format flags) by design. Doc comments strengthened to make this explicit and point secret-bearing callers at `Encrypt` / `Protected`.

Map-state guards (765ffac): `AesMapCipher::passthrough_entry` and `AesMapCipher::end` now reject a pending `current_key`, matching the check `encrypt_key` already had. `Encrypt for u32` routes via `encrypt_bytes_array(Protected::new(self.to_le_bytes()))` to skip a heap hop.

### Breaking

Two `Cipher` trait method signatures change shape. External implementers must update `encrypt_bytes_vec` / `encrypt_bytes_array`.

## Validation

- **Zeroize audit (`/zeroize-audit` skill)** — initial run on `vitaminc-aead` + `vitaminc-encrypt` surfaced 12 findings driving the refactor above. Re-run after `def6668`: 8 prior findings resolved (1 high, 7 medium), passthrough correctly not re-flagged (matches design intent). One **new cross-crate issue** surfaced: `vitaminc_protected::Protected<T>` derives `Zeroize` and asserts `ZeroizeOnDrop` but **has no `Drop` impl**, so the leaf zeroize threaded by this PR is not yet enforced at runtime. The type-level chain of custody is correct; the leaf wipe will start firing automatically once `Protected<T>` gains its `Drop`. Tracked in #181.
- **Copilot review** — 6 inline comments. Two addressed in 765ffac (`AesMapCipher::passthrough_entry` and `end` were missing the `current_key` contract check; plus a `u32` `Encrypt` impl simplification). Three deferred to #181 with explicit replies on each thread — the three `visit_bytes_vec` error-path comments are subsumed by the Protected `Drop` fix. One non-actionable.
- **Claude Code review (in-loop)** — planning agent designed the trait surface change before code was written; explore agent mapped all `risky_unwrap` call sites across the workspace; the plan was refined against a "prefer `Protected` over `Zeroizing`" preference and a "passthrough is intentionally non-sensitive" design intent saved as project memory.
- **Manual review** — by the PR author.

## Test plan

- [x] `roundtrip_option_some_string` (quickcheck)
- [x] `roundtrip_option_none_string`
- [x] `roundtrip_option_some_with_aad` / `roundtrip_option_none_with_aad`
- [x] `decrypt_option_none_fails_with_wrong_aad`
- [x] `decrypt_string_rejects_none_ciphertext` (empty-plaintext masking guard)
- [x] `decrypt_option_rejects_passthrough_ciphertext` (type-laundering guard)
- [x] `roundtrip_passthrough_u32` / `roundtrip_passthrough_string`
- [x] `passthrough_type_mismatch_returns_err`
- [x] `roundtrip_vec_of_option_string` (Option inside SeqAccess)
- [x] `encrypt_key_twice_without_value_fails` — MapCipher state-machine
- [x] `encrypt_value_without_pending_key_fails`
- [x] `passthrough_entry_with_pending_key_fails`
- [x] `end_with_pending_key_fails`
- [x] `passthrough_entry_succeeds_with_no_pending_key`
- [x] `cargo test --workspace` green (one unrelated `vitaminc-kms` test fails locally on missing localstack; CI-side `Test` job green)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean

## Follow-ups

- #181 `Protected<T>` missing `Drop` impl (also affects `Equatable<T>` / `Exportable<T>`)
- Wire-format serialization of `AesCipherText` containing `Passthrough` (`Box<dyn Any>` is not `Serialize`)
- `SeqAccess::next_passthrough` / `MapAccess::next_passthrough_entry` — add only when the future derive macro needs them

### DIFF
--- a/packages/aead/src/decipher/impls.rs
+++ b/packages/aead/src/decipher/impls.rs
@@ -111,3 +111,12 @@ where
         D::map_ok(T::decrypt(decipher), Protected::init_from_inner)
     }
 }
+
+impl<'c, T> Decrypt<'c> for Option<T>
+where
+    T: Decrypt<'c> + 'c,
+{
+    fn decrypt<D: Decipher<'c>>(decipher: D) -> D::Ok<Self> {
+        decipher.decrypt_option::<T>()
+    }
+}
--- a/packages/aead/src/decipher/impls.rs
+++ b/packages/aead/src/decipher/impls.rs
@@ -9,10 +9,7 @@ impl<'c> Decrypt<'c> for Vec<u8> {
         struct BytesVisitor;
         impl<'c> DecipherVisitor<'c> for BytesVisitor {
             type Value = Vec<u8>;
-            fn visit_bytes_vec(
-                self,
-                data: Protected<Vec<u8>>,
-            ) -> Result<Self::Value, Unspecified> {
+            fn visit_bytes_vec(self, data: Protected<Vec<u8>>) -> Result<Self::Value, Unspecified> {
                 // The caller asked for a bare `Vec<u8>` — this is the
                 // explicit extraction boundary where ownership leaves the
                 // cipher pipeline.
@@ -28,10 +25,7 @@ impl<'c> Decrypt<'c> for String {
         struct StringVisitor;
         impl<'c> DecipherVisitor<'c> for StringVisitor {
             type Value = String;
-            fn visit_bytes_vec(
-                self,
-                data: Protected<Vec<u8>>,
-            ) -> Result<Self::Value, Unspecified> {
+            fn visit_bytes_vec(self, data: Protected<Vec<u8>>) -> Result<Self::Value, Unspecified> {
                 String::from_utf8(data.risky_unwrap()).map_err(|_| Unspecified)
             }
         }
@@ -44,10 +38,7 @@ impl<'c, const N: usize> Decrypt<'c> for [u8; N] {
         struct ArrayVisitor<const N: usize>;
         impl<'c, const N: usize> DecipherVisitor<'c> for ArrayVisitor<N> {
             type Value = [u8; N];
-            fn visit_bytes_vec(
-                self,
-                data: Protected<Vec<u8>>,
-            ) -> Result<Self::Value, Unspecified> {
+            fn visit_bytes_vec(self, data: Protected<Vec<u8>>) -> Result<Self::Value, Unspecified> {
                 data.risky_unwrap().try_into().map_err(|_| Unspecified)
             }
         }
@@ -60,10 +51,7 @@ impl<'c> Decrypt<'c> for u32 {
         struct U32Visitor;
         impl<'c> DecipherVisitor<'c> for U32Visitor {
             type Value = u32;
-            fn visit_bytes_vec(
-                self,
-                data: Protected<Vec<u8>>,
-            ) -> Result<Self::Value, Unspecified> {
+            fn visit_bytes_vec(self, data: Protected<Vec<u8>>) -> Result<Self::Value, Unspecified> {
                 let bytes: [u8; 4] = data.risky_unwrap().try_into().map_err(|_| Unspecified)?;
                 Ok(u32::from_le_bytes(bytes))
             }
--- a/packages/aead/src/decipher/impls.rs
+++ b/packages/aead/src/decipher/impls.rs
@@ -9,8 +9,14 @@ impl<'c> Decrypt<'c> for Vec<u8> {
         struct BytesVisitor;
         impl<'c> DecipherVisitor<'c> for BytesVisitor {
             type Value = Vec<u8>;
-            fn visit_bytes_vec(self, data: Vec<u8>) -> Result<Self::Value, Unspecified> {
-                Ok(data)
+            fn visit_bytes_vec(
+                self,
+                data: Protected<Vec<u8>>,
+            ) -> Result<Self::Value, Unspecified> {
+                // The caller asked for a bare `Vec<u8>` — this is the
+                // explicit extraction boundary where ownership leaves the
+                // cipher pipeline.
+                Ok(data.risky_unwrap())
             }
         }
         decipher.decrypt_bytes(BytesVisitor)
@@ -22,8 +28,11 @@ impl<'c> Decrypt<'c> for String {
         struct StringVisitor;
         impl<'c> DecipherVisitor<'c> for StringVisitor {
             type Value = String;
-            fn visit_bytes_vec(self, data: Vec<u8>) -> Result<Self::Value, Unspecified> {
-                String::from_utf8(data).map_err(|_| Unspecified)
+            fn visit_bytes_vec(
+                self,
+                data: Protected<Vec<u8>>,
+            ) -> Result<Self::Value, Unspecified> {
+                String::from_utf8(data.risky_unwrap()).map_err(|_| Unspecified)
             }
         }
         decipher.decrypt_bytes(StringVisitor)
@@ -35,8 +44,11 @@ impl<'c, const N: usize> Decrypt<'c> for [u8; N] {
         struct ArrayVisitor<const N: usize>;
         impl<'c, const N: usize> DecipherVisitor<'c> for ArrayVisitor<N> {
             type Value = [u8; N];
-            fn visit_bytes_vec(self, data: Vec<u8>) -> Result<Self::Value, Unspecified> {
-                data.try_into().map_err(|_| Unspecified)
+            fn visit_bytes_vec(
+                self,
+                data: Protected<Vec<u8>>,
+            ) -> Result<Self::Value, Unspecified> {
+                data.risky_unwrap().try_into().map_err(|_| Unspecified)
             }
         }
         decipher.decrypt_bytes(ArrayVisitor::<N>)
@@ -48,8 +60,11 @@ impl<'c> Decrypt<'c> for u32 {
         struct U32Visitor;
         impl<'c> DecipherVisitor<'c> for U32Visitor {
             type Value = u32;
-            fn visit_bytes_vec(self, data: Vec<u8>) -> Result<Self::Value, Unspecified> {
-                let bytes: [u8; 4] = data.try_into().map_err(|_| Unspecified)?;
+            fn visit_bytes_vec(
+                self,
+                data: Protected<Vec<u8>>,
+            ) -> Result<Self::Value, Unspecified> {
+                let bytes: [u8; 4] = data.risky_unwrap().try_into().map_err(|_| Unspecified)?;
                 Ok(u32::from_le_bytes(bytes))
             }
         }
--- a/packages/encrypt/src/cipher.rs
+++ b/packages/encrypt/src/cipher.rs
@@ -708,4 +708,64 @@ mod test {
         let decrypted: Vec<Option<String>> = cipher.decrypt(ciphertext).expect("Decryption failed");
         decrypted == items
     }
+
+    // --- MapCipher state-machine contract ---
+    //
+    // `AesMapCipher` enforces a strict key→value pairing: every `encrypt_key`
+    // must be followed by exactly one `encrypt_value` (or `passthrough_entry`
+    // for the unencrypted variant) before the next key or `end`. The four
+    // tests below pin down each branch where the contract can be violated.
+
+    #[test]
+    fn encrypt_key_twice_without_value_fails() {
+        let key = Key::from([42u8; 32]);
+        let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
+        let map = (&cipher).encrypt_map().encrypt_key("first").unwrap();
+        assert!(map.encrypt_key("second").is_err());
+    }
+
+    #[test]
+    fn encrypt_value_without_pending_key_fails() {
+        let key = Key::from([42u8; 32]);
+        let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
+        let map = (&cipher).encrypt_map();
+        assert!(map.encrypt_value("orphan-value", ()).is_err());
+    }
+
+    #[test]
+    fn passthrough_entry_with_pending_key_fails() {
+        let key = Key::from([42u8; 32]);
+        let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
+        let map = (&cipher).encrypt_map().encrypt_key("pending").unwrap();
+        // Adopting the new key here would silently drop "pending".
+        assert!(map.passthrough_entry("other", 42u32).is_err());
+    }
+
+    #[test]
+    fn end_with_pending_key_fails() {
+        let key = Key::from([42u8; 32]);
+        let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
+        let map = (&cipher).encrypt_map().encrypt_key("pending").unwrap();
+        assert!(map.end().is_err());
+    }
+
+    #[test]
+    fn passthrough_entry_succeeds_with_no_pending_key() {
+        let key = Key::from([42u8; 32]);
+        let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
+        // Sanity: the new guard does not break the happy path.
+        let ciphertext = (&cipher)
+            .encrypt_map()
+            .passthrough_entry("version", 1u32)
+            .and_then(|m| m.end())
+            .expect("passthrough_entry should succeed without a pending key");
+        match ciphertext {
+            AesCipherText::Map(entries) => {
+                assert_eq!(entries.len(), 1);
+                assert_eq!(entries[0].0, "version");
+                assert!(matches!(entries[0].1, AesCipherText::Passthrough(_)));
+            }
+            _ => panic!("expected Map ciphertext"),
+        }
+    }
 }
--- a/packages/encrypt/src/cipher.rs
+++ b/packages/encrypt/src/cipher.rs
@@ -6,7 +6,7 @@ use vitaminc_aead::{
     LocalCipherText, MapAccess, MapCipher, NonceGenerator, RandomNonceGenerator, SeqAccess,
     SeqCipher, Unspecified,
 };
-use vitaminc_protected::Controlled;
+use vitaminc_protected::Protected;
 
 /// The recursive ciphertext container produced by [`Aes256Cipher`].
 ///
@@ -61,7 +61,11 @@ impl<'c> Cipher for &'c Aes256Cipher {
     type SeqCipher = AesSeqCipher<'c>;
     type MapCipher = AesMapCipher<'c>;
 
-    fn encrypt_bytes_vec<'a, A>(self, data: Vec<u8>, aad: A) -> Result<Self::Ok, Self::Error>
+    fn encrypt_bytes_vec<'a, A>(
+        self,
+        data: Protected<Vec<u8>>,
+        aad: A,
+    ) -> Result<Self::Ok, Self::Error>
     where
         A: IntoAad<'a>,
     {
@@ -251,14 +255,13 @@ impl AesDecipher<'_> {
         cipher: &Aes256Cipher,
         ct: LocalCipherText,
         aad: &[u8],
-    ) -> Result<Vec<u8>, Unspecified> {
+    ) -> Result<Protected<Vec<u8>>, Unspecified> {
         let (nonce, reader) = ct.into_reader().read_nonce::<NONCE_LEN>();
         let nonce_bytes = nonce.into_inner();
 
         reader
             .accepts_plaintext_ok(|data| cipher.key.open(&nonce_bytes, aad, data))
             .read()
-            .map(|data| data.risky_unwrap())
     }
 }
 
--- a/packages/encrypt/src/cipher.rs
+++ b/packages/encrypt/src/cipher.rs
@@ -1,5 +1,6 @@
 use crate::backend::{CipherKey, NONCE_LEN};
 use crate::Key;
+use std::any::Any;
 use vitaminc_aead::{
     Cipher, CipherTextBuilder, Decipher, DecipherVisitor, Decrypt, Encrypt, IntoAad,
     LocalCipherText, MapAccess, MapCipher, NonceGenerator, RandomNonceGenerator, SeqAccess,
@@ -22,6 +23,15 @@ pub enum AesCipherText {
     /// A map of (cleartext key, ciphertext value) pairs produced from a
     /// `HashMap`-shaped plaintext. Keys are not encrypted.
     Map(Vec<(String, AesCipherText)>),
+    /// The authenticated absent marker produced by [`Cipher::encrypt_none`].
+    /// Stores a sealed empty plaintext whose tag binds the supplied AAD.
+    None(LocalCipherText),
+    /// A typed value passed through unencrypted via [`Cipher::passthrough`].
+    /// Not serializable to bytes — see
+    /// `packages/aead/src/cipher.rs` for the API-level type bound and the
+    /// runtime downcast performed by
+    /// [`Decipher::decrypt_passthrough`](vitaminc_aead::Decipher::decrypt_passthrough).
+    Passthrough(Box<dyn Any + Send + 'static>),
 }
 
 /// Implements AES-256-GCM. Backend is selected at compile time:
@@ -86,9 +96,32 @@ impl<'c> Cipher for &'c Aes256Cipher {
         }
     }
 
-    // Tracked: https://github.com/cipherstash/vitaminc/issues/172
-    // fn encrypt_none(self) -> Result<Self::Ok, Self::Error> { }
-    // fn passthrough<T: 'static>(self, data: T) -> Result<Self::Ok, Self::Error> { }
+    fn encrypt_none<'a, A>(self, aad: A) -> Result<Self::Ok, Self::Error>
+    where
+        A: IntoAad<'a>,
+    {
+        let nonce = self.nonce_generator.generate()?;
+        let nonce_bytes: [u8; NONCE_LEN] = nonce.as_ref().try_into().map_err(|_| Unspecified)?;
+        let aad = aad.into_aad();
+
+        CipherTextBuilder::new()
+            .append_nonce(nonce)
+            .append_target_plaintext(Vec::<u8>::new())
+            .accepts_ciphertext_and_tag_ok(|mut buf| {
+                self.key
+                    .seal(&nonce_bytes, aad.as_bytes(), &mut buf)
+                    .map(|()| buf)
+            })
+            .build()
+            .map(AesCipherText::None)
+    }
+
+    fn passthrough<T>(self, value: T) -> Result<Self::Ok, Self::Error>
+    where
+        T: Any + Send + 'static,
+    {
+        Ok(AesCipherText::Passthrough(Box::new(value)))
+    }
 }
 
 /// [`SeqCipher`] driver for [`Aes256Cipher`]. Encrypts each element under its
@@ -110,6 +143,14 @@ impl<'c> SeqCipher for AesSeqCipher<'c> {
     {
         let encrypted = data.encrypt_with_aad(self.cipher, aad)?;
         self.items.push(encrypted);
+        Ok(self)
+    }
+
+    fn passthrough_next<T>(mut self, value: T) -> Result<Self, Self::Error>
+    where
+        T: Any + Send + 'static,
+    {
+        self.items.push(AesCipherText::Passthrough(Box::new(value)));
         Ok(self)
     }
 
@@ -150,6 +191,15 @@ impl<'c> MapCipher for AesMapCipher<'c> {
         let key = self.current_key.take().ok_or(Unspecified)?;
         let encrypted = value.encrypt_with_aad(self.cipher, aad)?;
         self.entries.push((key.to_string(), encrypted));
+        Ok(self)
+    }
+
+    fn passthrough_entry<T>(mut self, key: &'static str, value: T) -> Result<Self, Self::Error>
+    where
+        T: Any + Send + 'static,
+    {
+        self.entries
+            .push((key.to_string(), AesCipherText::Passthrough(Box::new(value))));
         Ok(self)
     }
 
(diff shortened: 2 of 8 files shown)


packages/aead/src/decipher/impls.rs
```diff
@@ -22,8 +25,8 @@ impl<'c> Decrypt<'c> for String {
         struct StringVisitor;
         impl<'c> DecipherVisitor<'c> for StringVisitor {
             type Value = String;
-            fn visit_bytes_vec(self, data: Vec<u8>) -> Result<Self::Value, Unspecified> {
-                String::from_utf8(data).map_err(|_| Unspecified)
+            fn visit_bytes_vec(self, data: Protected<Vec<u8>>) -> Result<Self::Value, Unspecified> {
+                String::from_utf8(data.risky_unwrap()).map_err(|_| Unspecified)
             }
```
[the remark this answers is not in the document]
coderdan: Deferred — once #181 (`Protected<T>` missing `Drop` impl) lands, the `Vec<u8>` extracted by `risky_unwrap` and dropped on the failing conversion is the same allocation that `Protected<Vec<u8>>` previously owned, so its `Drop` will wipe it without an explicit zeroize here.

packages/aead/src/decipher/impls.rs
```diff
@@ -35,8 +38,8 @@ impl<'c, const N: usize> Decrypt<'c> for [u8; N] {
         struct ArrayVisitor<const N: usize>;
         impl<'c, const N: usize> DecipherVisitor<'c> for ArrayVisitor<N> {
             type Value = [u8; N];
-            fn visit_bytes_vec(self, data: Vec<u8>) -> Result<Self::Value, Unspecified> {
-                data.try_into().map_err(|_| Unspecified)
+            fn visit_bytes_vec(self, data: Protected<Vec<u8>>) -> Result<Self::Value, Unspecified> {
+                data.risky_unwrap().try_into().map_err(|_| Unspecified)
             }
```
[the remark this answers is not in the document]
coderdan: Deferred — once #181 (`Protected<T>` missing `Drop` impl) lands, the `Vec<u8>` extracted by `risky_unwrap` and dropped on the failing conversion is the same allocation that `Protected<Vec<u8>>` previously owned, so its `Drop` will wipe it without an explicit zeroize here.

packages/aead/src/decipher/impls.rs
```diff
@@ -48,8 +51,8 @@ impl<'c> Decrypt<'c> for u32 {
         struct U32Visitor;
         impl<'c> DecipherVisitor<'c> for U32Visitor {
             type Value = u32;
-            fn visit_bytes_vec(self, data: Vec<u8>) -> Result<Self::Value, Unspecified> {
-                let bytes: [u8; 4] = data.try_into().map_err(|_| Unspecified)?;
+            fn visit_bytes_vec(self, data: Protected<Vec<u8>>) -> Result<Self::Value, Unspecified> {
+                let bytes: [u8; 4] = data.risky_unwrap().try_into().map_err(|_| Unspecified)?;
                 Ok(u32::from_le_bytes(bytes))
```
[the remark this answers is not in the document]
coderdan: Deferred — once #181 (`Protected<T>` missing `Drop` impl) lands, the `Vec<u8>` extracted by `risky_unwrap` and dropped on the failing conversion is the same allocation that `Protected<Vec<u8>>` previously owned, so its `Drop` will wipe it without an explicit zeroize here.

packages/encrypt/src/cipher.rs
```diff
@@ -153,7 +198,27 @@ impl<'c> MapCipher for AesMapCipher<'c> {
...
+    {
+        // A key already pending means `encrypt_key` ran without a matching
+        // `encrypt_value` — adopting it here would silently drop the pending
+        // key, which is the same trait-contract violation `encrypt_key`
+        // rejects.
+        if self.current_key.is_some() {
+            return Err(Unspecified);
+        }
+        self.entries
+            .push((key.to_string(), AesCipherText::Passthrough(Box::new(value))));
```
auxesis, line 225: **LOW: No duplicate-key check across completed entries.**

The `is_some()` guard on `current_key` (added in 765ffac, nice catch) closes the in-progress-key contract. But `entries: Vec<(String, AesCipherText)>` still allows the same `&'static str` key to be pushed multiple times across *completed* `encrypt_value` / `passthrough_entry` calls. On decrypt, `Decrypt for HashMap<String, T>::visit_map` inserts into a `HashMap` — silently last-wins on duplicates, no error.

Not exploitable today: Rust's `HashMap` dedups on the encrypt side (via `Encrypt for HashMap<&'static str, T>`), and `AesCipherText` isn't `Serialize` yet. Becomes live the moment either a custom encoder emits entries directly (e.g. a derive macro) or wire-format support lands.

Two cheap options:
1. Document `AesMapCipher` as a positional list — "downstream decoders own dedup" — consistent with the trait's existing "trust the caller" stance.
2. Add a `seen: HashSet<&'static str>` to `AesMapCipher` and `Err(Unspecified)` on duplicates — more defensive.

(1) is cheaper and matches existing posture; (2) makes the contract structurally enforced. Either is fine; flagging so it's a deliberate choice.

packages/encrypt/src/cipher.rs
```diff
@@ -264,6 +328,41 @@ impl<'c> Decipher<'c> for AesDecipher<'c> {
...
+            AesCipherText::Passthrough(_) => Err(Unspecified),
+            other => {
+                let inner = AesDecipher {
+                    cipher: self.cipher,
+                    ciphertext: other,
+                    aad: self.aad,
+                };
+                T::decrypt(inner).map(Some)
+            }
+        }
```
auxesis, line 388: **LOW: Caller-chosen-type-shape ambiguity for nested `Option`s.**

The `other` arm is exactly what lets `Option<Vec<...>>`, `Option<HashMap<...>>`, etc. compose naturally — recurse into `T::decrypt` for any non-`None`, non-`Passthrough` variant. Beautiful and correct.

The side-effect: **`Some(Some("x"))` and `Some("x")` produce identical `AesCipherText::Single(...)` bytes**, and on decrypt:

- `cipher.decrypt::<Option<String>>(ct)` returns `Some("x")`
- `cipher.decrypt::<Option<Option<String>>>(ct)` returns `Some(Some("x"))`

Both succeed without error against the same ciphertext. This is the same property serde has for `Option` — no depth tag in the bytes; the caller's type at the call site decides. Almost certainly fine for everyone who fixes a concrete type at the call site (which is everyone today).

Worth pinning with a doc note + a negative-space test so the property is intentional rather than accidental, and an accidental future "fix" can't break the recursion door:

```rust
#[test]
fn nested_option_shape_indistinguishable() {
    let key = Key::from([13u8; 32]);
    let cipher = Aes256Cipher::new(&key).unwrap();
    let outer: Option<Option<String>> = Some(Some("x".into()));
    let ct = outer.encrypt(&cipher).unwrap();
    // Decrypting as Option<String> succeeds with Some("x"), not Some(Some(...)).
    let decoded: Option<String> = cipher.decrypt(ct).expect("decode should succeed");
    assert_eq!(decoded, Some("x".into()));
}
```

packages/encrypt/src/cipher.rs
```diff
@@ -264,6 +328,41 @@ impl<'c> Decipher<'c> for AesDecipher<'c> {
...
+                Self::decrypt_local_ciphertext(self.cipher, ct, &self.aad)?;
+                Ok(None)
+            }
+            // Passthrough must never be decoded as an Option payload.
+            AesCipherText::Passthrough(_) => Err(Unspecified),
+            other => {
+                let inner = AesDecipher {
+                    cipher: self.cipher,
+                    ciphertext: other,
+                    aad: self.aad,
```
auxesis, line 384: **LOW: AAD does not bind per-element position in `Sequence` ciphertexts.**

This recursion (moved AAD here; cloned per element in `AesSeqAccess::next_element` / `AesMapAccess::next_entry`) is the right shape for the current scope. The invariant it establishes: **every element of an `AesCipherText::Sequence` is sealed under the same AAD with no position binding**. An attacker who can reorder elements of an in-memory `Sequence(Vec<_>)` won't trip any tag-verification failure on decrypt.

Unreachable from the wire today (since `AesCipherText` isn't `Serialize`), so informational. **But it becomes a MEDIUM concern the moment wire-format support lands** unless per-position AAD framing is introduced concurrently. Same concern applies to inter-`Map`-entry value substitution across same-AAD ciphertexts.

Recording so the wire-format follow-up can deliberately choose between:
- (a) Accept positional-invariance as a documented design property (matches serde-style framing), or
- (b) Extend AAD with per-element index/position framing during the seal/open.

Nothing to do in this PR — just want this property to be a deliberate choice rather than an emergent one.

auxesis (commented):
Ran an analysis with @trailofbits's `audit-context-building` and `vulnerability-hunting` skills.

Results:

| Severity | Count |
|---|---|
| Critical / High / Medium | 0 |
| Low | 4 |
| Informational | 4 |
| Security improvements (this PR adds) | 7 |

3 of 4 LOW findings are inline below. 

The 4th doesn't fit GitHub's diff-line constraint so it's here:

### LOW-1: `Key::cipher_key` `risky_ref` lacks inline SAFETY comment

`packages/encrypt/src/key.rs:15` — `CipherKey::new(self.0.risky_ref())`. The line wasn't touched by this PR, so I can't comment inline. The PR's chain-of-custody doctrine (the `Encrypt for Protected<T>` SAFETY block at `packages/aead/src/encrypt/impls.rs:99-106`) now applies consistently to every `risky_*` call site in `aead`/`encrypt` — except this one. Of the 8 production `risky_*` sites at PR head, this is the only one without an inline justification.

Non-blocking suggestion:

```rust
impl Key {
    pub(crate) fn cipher_key(&self) -> Result<CipherKey, Unspecified> {
        // SAFETY: borrowed `&[u8; 32]` is consumed by `CipherKey::new`
        // for the duration of the constructor; the backend cipher key
        // handle holds an opaque copy thereafter.
        CipherKey::new(self.0.risky_ref())
    }
}
```

packages/encrypt/src/cipher.rs
```diff
@@ -113,6 +150,14 @@ impl<'c> SeqCipher for AesSeqCipher<'c> {
         Ok(self)
     }
 
+    fn passthrough_next<T>(mut self, value: T) -> Result<Self, Self::Error>
+    where
+        T: Any + Send + 'static,
+    {
+        self.items.push(AesCipherText::Passthrough(Box::new(value)));
```
auxesis, line 157: **LOW: `passthrough_next` has zero direct test coverage.**

The Cipher-level `passthrough` is well-tested (`roundtrip_passthrough_*`, `passthrough_type_mismatch_returns_err`, `decrypt_option_rejects_passthrough_ciphertext`), but the SeqCipher-level `passthrough_next` is reached by **no test**. If a future refactor inadvertently breaks it (e.g. pushes to the wrong field or wraps differently), no test catches it.

One test pins the happy path **and** the mixed-with-`encrypt_next` shape:

```rust
#[test]
fn seq_cipher_accepts_mixed_encrypt_next_and_passthrough_next() {
    let cipher = Aes256Cipher::new(&Key::from([42u8; 32])).unwrap();
    let ct = (&cipher)
        .encrypt_seq(Some(3))
        .encrypt_next("first", ()).unwrap()
        .passthrough_next(99u32).unwrap()
        .encrypt_next("third", ()).unwrap()
        .end()
        .unwrap();
    match ct {
        AesCipherText::Sequence(items) => {
            assert_eq!(items.len(), 3);
            assert!(matches!(items[0], AesCipherText::Single(_)));
            assert!(matches!(items[1], AesCipherText::Passthrough(_)));
            assert!(matches!(items[2], AesCipherText::Single(_)));
        }
        _ => panic!("expected Sequence"),
    }
}
```

packages/encrypt/src/cipher.rs
auxesis, line 225: **LOW: Mixed encrypted + passthrough entries are untested.**

(Separate concern from the dedupe note above.) `passthrough_entry_succeeds_with_no_pending_key` has *one* passthrough entry. No test mixes `encrypt_value` and `passthrough_entry` in the same map — the state machine (FC.4) allows any ordering as long as the `current_key` guard holds, but the round-trip semantics aren't pinned.

The interesting question: **what does `cipher.decrypt::<HashMap<String, String>>(mixed_map_ct)` do?** Inferring from the design, FC.1's catch-all in `decrypt_bytes` rejects the Passthrough value when `String`'s visitor runs against it, so the whole decode fails. That's the expected behaviour, but it's not pinned anywhere:

```rust
#[test]
fn map_with_mixed_encrypted_and_passthrough_entries_cannot_decode_as_uniform_hashmap() {
    let cipher = Aes256Cipher::new(&Key::from([12u8; 32])).unwrap();
    let ct = (&cipher)
        .encrypt_map()
        .encrypt_key("name").unwrap()
        .encrypt_value("Alice".to_string(), ()).unwrap()
        .passthrough_entry("schema_version", 1u32).unwrap()
        .end()
        .unwrap();
    // Decoding as HashMap<String, String> hits FC.1 in the visitor for the
    // passthrough entry's value; the whole decode fails.
    assert!(cipher.decrypt::<std::collections::HashMap<String, String>>(ct).is_err());
}
```

The **companion happy case** — `HashMap<&'static str, Option<T>>` where mixed shapes (Some/None per value) are expected and decode correctly — is also missing. Together these two tests pin the design boundary: **mixed-shape values work iff the decode type explicitly accepts the variation** (via `Option<T>`); a uniform `T` decoder rejects.

Same structural gap applies to `AesSeqCipher::passthrough_next` mixed with `encrypt_next` — see the inline on `passthrough_next` above.

packages/encrypt/src/cipher.rs
```diff
@@ -264,6 +328,41 @@ impl<'c> Decipher<'c> for AesDecipher<'c> {
...
+    where
+        T: Any + Send + 'static,
+    {
+        match self.ciphertext {
+            AesCipherText::Passthrough(boxed) => {
+                boxed.downcast::<T>().map(|b| *b).map_err(|_| Unspecified)
+            }
+            _ => Err(Unspecified),
+        }
+    }
```
auxesis, line 354: **LOW: Variant-rejection catch-alls untested across the FC.1 cluster.**

The `_ => Err(Unspecified)` at line 340 rejects 4 distinct variant cases (Single, Sequence, Map, None). **None of them are pinned by a test.** `passthrough_type_mismatch_returns_err` exercises the Arm 1 downcast-fail path (Passthrough-with-wrong-type), not the catch-all.

The same gap shape extends across the four `decrypt_*` methods that use this pattern:

| Method | Catch-all rejects | Tests |
|---|---|---|
| `decrypt_bytes` (cipher.rs:300) | Passthrough, Sequence, Map, None | 1 (None — `decrypt_string_rejects_none_ciphertext`) |
| `decrypt_seq` (cipher.rs:314) | Single, Map, None, Passthrough | 0 |
| `decrypt_map` (cipher.rs:328) | Single, Sequence, None, Passthrough | 0 |
| `decrypt_passthrough` (here, line 340) | Single, Sequence, Map, None | 0 |

**2 of 16 catch-all sub-paths tested across the cluster.** This is the doctrine the PR's design most relies on (cluster FC.1 in the deep-context build) — leaving 14 paths invisible to regression elevates the cost of an accidental future change to the match shape.

14 mechanical tests close the gap:

```rust
// decrypt_passthrough rejections (4 tests, same shape)
#[test]
fn decrypt_passthrough_rejects_single_ciphertext() {
    let cipher = Aes256Cipher::new(&Key::from([1u8; 32])).unwrap();
    let ct = "hello".to_string().encrypt(&cipher).unwrap();
    let r: Result<u32, _> = decrypt_passthrough_via(&cipher, ct);
    assert!(r.is_err());
}
// (+3 for Sequence, Map, None)

// decrypt_string rejections (3 more — None is already covered)
#[test]
fn decrypt_string_rejects_passthrough_ciphertext() {
    let cipher = Aes256Cipher::new(&Key::from([1u8; 32])).unwrap();
    let ct = (&cipher).passthrough(42u32).unwrap();
    assert!(cipher.decrypt::<String>(ct).is_err());
}
// (+2 for Sequence, Map)

// decrypt_vec rejections (4)
#[test]
fn decrypt_vec_rejects_single_ciphertext() {
    let cipher = Aes256Cipher::new(&Key::from([1u8; 32])).unwrap();
    let ct = "hello".to_string().encrypt(&cipher).unwrap();
    assert!(cipher.decrypt::<Vec<String>>(ct).is_err());
}
// (+3 for Map, None, Passthrough)

// decrypt_hashmap rejections (3 — Sequence already covered by decrypt_vec's shape)
// Actually 4 to be symmetric: (+4 for Single, Sequence, None, Passthrough)
```

Mechanical, defensive, highest coverage-per-line in the gap analysis. **P1 priority.**

packages/encrypt/src/cipher.rs
auxesis, line 388: **LOW: `Option<T>` type variation is one-deep; the `other` arm's Sequence/Map sub-paths are untested.**

The `other` arm (lines 356-362) is the recursion door — what makes `Option<Vec<T>>`, `Option<HashMap<K,V>>`, `Option<Protected<T>>` compose. Today only `T = String` is exercised (which lands the *Single* sub-path of the `other` arm via `roundtrip_option_some_string`). **The Sequence and Map sub-paths have zero coverage.**

The four canonical T variations close it:

```rust
#[test]
fn roundtrip_option_some_u32() {
    let cipher = Aes256Cipher::new(&Key::from([1u8; 32])).unwrap();
    let ct = Some(42u32).encrypt(&cipher).unwrap();
    assert_eq!(cipher.decrypt::<Option<u32>>(ct).unwrap(), Some(42));
}

#[quickcheck]
fn roundtrip_option_some_vec_of_strings(items: Vec<String>) -> bool {
    let cipher = Aes256Cipher::new(&Key::from([2u8; 32])).unwrap();
    let ct = Some(items.clone()).encrypt(&cipher).unwrap();
    cipher.decrypt::<Option<Vec<String>>>(ct).unwrap() == Some(items)
}

#[test]
fn roundtrip_option_some_hashmap() {
    let cipher = Aes256Cipher::new(&Key::from([3u8; 32])).unwrap();
    let mut m = std::collections::HashMap::new();
    m.insert("name", "Alice");
    let ct = Some(m).encrypt(&cipher).unwrap();
    let decoded: Option<HashMap<String, String>> = cipher.decrypt(ct).unwrap();
    assert_eq!(decoded.unwrap().get("name").map(String::as_str), Some("Alice"));
}

#[quickcheck]
fn roundtrip_option_some_protected_string(plaintext: String) -> bool {
    use vitaminc_protected::{Controlled, Protected};
    let cipher = Aes256Cipher::new(&Key::from([4u8; 32])).unwrap();
    let ct = Some(Protected::new(plaintext.clone())).encrypt(&cipher).unwrap();
    let decoded: Option<Protected<String>> = cipher.decrypt(ct).unwrap();
    decoded.map(Controlled::risky_unwrap) == Some(plaintext)
}
```

Four tests cover: Single sub-path with non-String leaf (u32), Sequence sub-path (Vec), Map sub-path (HashMap), and the Protected-wrap re-establishment (the `map_ok` path in `Decrypt for Protected<T>`). **P2 priority** — pairs with the FC.1 pinning above to give full path-shape coverage on the new decipher methods.

auxesis (commented):
## Test coverage gap analysis

Follow-up to the differential review — same severity scale (LOW / INFO), same recommendation (APPROVE; all gaps are non-blocking but high-value to close).

### Summary

Coverage on **headline cases** is strong: `Option<String>` Some/None, passthrough for `u32` / `String`, and the `AesMapCipher` state machine has full transition coverage (5 tests pin 4 transitions + 1 happy path).

Two structural gaps stand out:

1. **FC.1 catch-all variant-rejection paths — 2 of 16 sub-paths tested (12.5%).** See the `decrypt_passthrough` inline. The under-pinning extends to `decrypt_bytes` (`cipher.rs:300`) and `decrypt_seq` (`cipher.rs:314`) — same `_ => Err(Unspecified)` catch-all design, none of their variant-rejection sub-paths exercised by a test. Those two lines aren't in the diff, so they can't take inline comments, but the gap shape is identical to the `decrypt_passthrough` and `decrypt_map` ones.

2. **Type variation on `Option<T>` is one-deep.** Only `Option<String>` round-trips. `Option<Vec<T>>`, `Option<HashMap<K,V>>`, `Option<Protected<T>>`, `Option<u32>` have zero coverage, and as a consequence the `other`-arm Sequence and Map sub-paths of `decrypt_option` are entirely untested. See the `decrypt_option` inline.

### Smaller gaps (highlights)

- `Decrypt for Vec<u8>` (`packages/aead/src/decipher/impls.rs:7-20`) exists in production with **zero test exercise**. There's no `cipher.decrypt::<Vec<u8>>(ct)` call anywhere; `roundtrip_byte_array` exercises the `[u8; 16]` chain, not the `Vec<u8>` one. Either pin it with `decrypt_byte_array_ciphertext_as_vec_u8` or document why the impl exists.
- `Protected<T>` only covers `String` (`roundtrip_protected_string`). The other three primitive `Decrypt` visitors (`u32`, `Vec<u8>`, `[u8; N]`) aren't exercised end-to-end through `Protected`.
- `passthrough<Option<U>>` would pin the Any-`TypeId` property — `Option<u32>` is a distinct type from `u32` even when the value is `Some(42)`. Cheap test, high semantic value.
- Nonce uniqueness across consecutive encrypt calls on the same `Aes256Cipher` — not explicitly pinned. The ref-cipher is designed for reuse, but no test asserts two consecutive ciphertexts of the same plaintext differ. Fundamental AEAD property; one assertion would catch a future RNG regression.
- `Option<T>` Some + wrong-AAD not pinned through the `Option` type (transitively covered via `decrypt_fails_with_wrong_aad` on the inner String, but not via `decrypt_with_aad::<Option<String>, _>(ct, "wrong")`).

### Priority-ranked closure plan

If only one batch lands:
- **P1 — FC.1 catch-all pinning (~14 tests, ~50 lines).** Mechanical, defensive, highest coverage-per-line. See `decrypt_passthrough` inline for the skeleton.

If two:
- **+ P2 — Option<T> type variation (~4 tests, ~60 lines).** Closes the `decrypt_option` `other`-arm gap. See `decrypt_option` inline for the four tests.

If three:
- **+ P3 — Mixed seq/map entries + passthrough<Option<U>> Any-TypeId pin (~3 tests, ~30 lines).** See `passthrough_entry` and `passthrough_next` inlines.

**P1+P2+P3 ≈ 19 tests, ~140 lines.** Takes the PR from "headline cases tested, edges thin" to "full path × type × combination matrix".

### Not flagged (recorded for context)

- `encrypt_none` error sub-paths (RNG / backend failures) — require failure injection; skip unless test seams added for other reasons.
- `Option<T>` None arm with T variation — the None encrypt path is T-independent (always `Vec::<u8>::new()`).
- `Encrypt for u32` routing decision (765ffac) — `roundtrip_u32` confirms functional equivalence; pinning the routing choice itself would test impl-not-behaviour.

Full breakdown with coverage matrices and code skeletons for every recommendation is in a local companion document. Happy to share the P1 batch as a follow-up commit/PR if useful.

packages/encrypt/src/cipher.rs
coderdan: This has been [raised before](https://github.com/cipherstash/vitaminc/pull/148#issuecomment-4515410685) and the design is intentional.

I'll add a comment so that agents don't keep flagging it.

packages/encrypt/src/cipher.rs
coderdan: Documented in 41af342 — `AesMapCipher`'s rustdoc now frames it as a positional list intended for sources that already enforce key uniqueness (`HashMap`s, structs), performs no duplicate-key checks, and is last-wins on decode. Went with option (1) since it matches the trait's trust-the-caller stance. The wire-format follow-up (#183) records that this needs revisiting if duplicates ever become reachable from a wire format.

packages/encrypt/src/cipher.rs
coderdan: Documented and pinned. 41af342 adds the serde-style note on `decrypt_option`; 145b729 adds `nested_option_shape_is_caller_decided` to lock the property so an accidental future change to the match can't silently close the recursion door. Thanks for the skeleton — used it.

packages/encrypt/src/cipher.rs
coderdan: Covered in 145b729 — `seq_cipher_accepts_mixed_encrypt_next_and_passthrough_next` pins both the happy path and the mixed `encrypt_next`/`passthrough_next` shape (asserts the Single/Passthrough/Single layout). Used your skeleton.

packages/encrypt/src/cipher.rs
coderdan: Covered in 145b729: `map_with_mixed_entries_cannot_decode_as_uniform_hashmap` (a passthrough value fails a uniform `HashMap<_, String>` decode via the FC.1 catch-all) plus the companion happy case `roundtrip_hashmap_of_option_values` (mixed Some/None values decode correctly as `HashMap<_, Option<String>>`). Both design-boundary tests landed.

packages/encrypt/src/cipher.rs
coderdan: Closed in 145b729. Four grouped tests — `decrypt_bytes_rejects_non_single_variants`, `decrypt_seq_rejects_non_sequence_variants`, `decrypt_map_rejects_non_map_variants`, `decrypt_passthrough_rejects_non_passthrough_variants` — assert every non-matching variant is rejected, covering the FC.1 catch-all sub-paths across the cluster. Grouped per-method with a label on each assert (rather than one test per sub-path); same coverage, tighter.

packages/encrypt/src/cipher.rs
coderdan: Closed in 145b729: `roundtrip_option_some_u32` (quickcheck), `roundtrip_option_some_vec_of_strings` (quickcheck), `roundtrip_option_some_hashmap`, `roundtrip_option_some_protected_string` (quickcheck), plus `decrypt_option_some_fails_with_wrong_aad`. Exercises the Single/Sequence/Map sub-paths of the `other` arm and the `Protected` rewrap.

coderdan:
Thanks @auxesis — really thorough review. Addressed the non-blocking findings across two commits (41af342 docs, 145b729 tests), plus a few items from the review body that didn't have an inline thread:

- **LOW-1 (`Key::cipher_key` SAFETY comment):** added in 41af342 — it was the only `risky_*` site in `encrypt` without the chain-of-custody justification the rest of the PR established.
- **`Decrypt for Vec<u8>` unexercised:** pinned in 145b729 via `decrypt_byte_array_ciphertext_as_vec_u8`. Note there is no `Encrypt for Vec<u8>` (no `Encrypt for u8`), so a `[u8; N]` `Single` ciphertext is the canonical producer decodable as `Vec<u8>` — a `Vec<u8>` roundtrip wouldn't compile.
- **Nonce uniqueness:** added `nonce_is_unique_across_encryptions` as a quickcheck property so each case samples a fresh nonce pair.

The `Protected<T>` zeroize-on-error-path items from Copilot stay deferred to #181 (the error-path `Vec<u8>` is the same allocation `Protected` owns, so it wipes for free once the `Drop` lands).

Forward-looking concerns (positional AAD, map dedup-on-decode) are tracked in #183 alongside the `Passthrough` serialization work, so they get decided deliberately when a wire format lands.

auxesis (approved):
So good, nice work @coderdan!

PR #175, opened 2026-05-16
feat(aead, encrypt): HList-based static cipher shape behind `hlist` feature

coderdan:
Stacked on #174. Prototype of the **statically-shaped** ciphertext container, parallel to the existing dynamic `Cipher` / `AesCipherText` path, behind an opt-in `hlist` feature.

## Why a second design?

The two designs target different use cases — the feature flag lets users pick:

|                          | dynamic (default)                   | static (`hlist` feature)            |
|--------------------------|-------------------------------------|-------------------------------------|
| Ciphertext container     | recursive enum + `Vec` nodes        | typed HList — shape is in the type  |
| Passthrough storage      | `Box<dyn Any + Send + 'static>`     | `Passthrough<T>(T)` — direct        |
| Decrypt access           | runtime downcast / shape check      | destructuring, statically infallible |
| Best for                 | wasm32, JS bindings, dynamic schemas | native Rust, derive-macro output    |

## API surface (under `vitaminc_aead::hlist`)

- `HNil` / `HCons<H, T>` HList primitives + `hlist!` / `hlist_pat!` macros
- Leaf types: `Encrypted(LocalCipherText)`, `Absent(LocalCipherText)`, `Passthrough<T>(T)`
- Composite types: `Map<L: HList>`, `Seq<L: HList>`, `Entry<V>`
- `StaticCipher` trait — counterpart to `Cipher` (no `type Ok` — each call produces a different type)
- `StaticMapBuilder<C, L>` — the growing-HList builder
- `Aes256Cipher::open` / `verify_absent` — decrypt-side helpers for the leaf types

## The worked example

The `User`-shaped struct test produces this fully-typed ciphertext:

```rust
type UserCiphertext = Map<HCons<
    Entry<Map<HCons<Entry<Encrypted>, HNil>>>, // preferences (nested)
    HCons<Entry<Absent>,                       // nickname: None
    HCons<Entry<Passthrough<u8>>,              // version (typed!)
    HCons<Entry<Encrypted>,                    // password_hash
    HNil>>>>>;
```

In production this alias would be generated by a derive macro — humans should rarely write it. A `Send` proof in the test file confirms no `Box`/`dyn Any` is introduced.

Decryption is destructuring:

```rust
let Map(HCons(prefs, HCons(nickname, HCons(version, HCons(pw, HNil))))) = ct;
let version: u8 = version.value.0;        // typed, infallible
let pw = cipher.open(pw.value, AAD)?;     // real crypto, fallible only on AAD/tag
```

## Known limitations (documented in `hlist/mod.rs`)

- **`Cipher::Ok` cannot stay a single associated type** — replaced by a separate `StaticCipher` trait. The two trait hierarchies coexist behind the feature flag.
- **Homogeneous `Vec<_>` of varying length is not expressible** in HList form. `Vec<Encrypted>` of uniform primitives still works, but the dynamic path's polymorphic `Vec<AesCipherText>` has no direct equivalent.
- **Type spellings are large** — derive macro should generate the aliases.

## Test plan

- [x] `static_user_roundtrip` — full mixed-field struct round-trip
- [x] `static_open_fails_with_wrong_aad` — AAD verification
- [x] `static_verify_absent_fails_with_wrong_aad` — `None` AAD verification
- [x] `static_passthrough_keeps_type` — passthrough is identity
- [x] `cargo test -p vitaminc-encrypt --features hlist` — 33 pass (29 base + 4 new)
- [x] `cargo test -p vitaminc-aead -p vitaminc-encrypt` (no feature) — 29 + 29 pass (no regression)
- [x] `cargo clippy --workspace --all-targets [--features vitaminc-encrypt/hlist] -- -D warnings` — clean both ways
- [x] `Send` proof on `UserCiphertext` (compile-time)

## Follow-ups

- Derive macro generating the `Encrypt` impl + `type ...Ciphertext = Map<HCons<...>>` alias from a struct definition
- `StaticSeqBuilder` for fixed-length heterogeneous sequences (currently only `Map` is implemented)
- Decide whether the dynamic and static paths should share leaf types (`Encrypted(LocalCipherText)`) or stay isolated

## Internal change

`Aes256Cipher.nonce_generator` and `.key` are now `pub(crate)` so the new `cipher_static` module can reuse them without going through the existing `Cipher` trait surface (which doesn't fit the HList shape).

### DIFF
--- a/packages/aead/Cargo.toml
+++ b/packages/aead/Cargo.toml
@@ -19,6 +19,19 @@ bytes = { version = "^1", features = ["serde"] }
 serde = { workspace = true }
 zeroize = { workspace = true }
 
+[features]
+test-utils = []
+# Opt-in HList-based ciphertext shape. Trades the dynamic `Cipher::Ok` enum
+# (heap-allocated `Vec`s, `Box<dyn Any>` passthrough) for a per-call typed
+# builder where the full structural shape is encoded in the type system —
+# no heap nodes, no type erasure, statically-typed passthrough access.
+#
+# Best suited to native Rust performance-sensitive paths; the dynamic
+# default remains preferable for wasm32 / JS bindings where bundle size
+# and runtime-determined shapes matter more.
+hlist = []
+
+
 [dev-dependencies]
 quickcheck = "1.1.0"
 quickcheck_macros = "1.2.0"
--- a/packages/aead/Cargo.toml
+++ b/packages/aead/Cargo.toml
@@ -20,7 +20,6 @@ serde = { workspace = true }
 zeroize = { workspace = true }
 
 [features]
-test-utils = []
 # Opt-in HList-based ciphertext shape. Trades the dynamic `Cipher::Ok` enum
 # (heap-allocated `Vec`s, `Box<dyn Any>` passthrough) for a per-call typed
 # builder where the full structural shape is encoded in the type system —
--- a/packages/aead/src/hlist/types.rs
+++ b/packages/aead/src/hlist/types.rs
@@ -3,12 +3,52 @@ use crate::LocalCipherText;
 use super::HList;
 
 /// A single sealed value (nonce + ciphertext + tag).
+///
+/// The inner `LocalCipherText` is private: leaves are produced by a cipher
+/// backend's `StaticCipher` impl (or, in future, a derive macro) via
+/// [`Encrypted::from_local`] rather than constructed field-wise by arbitrary
+/// callers. This is encapsulation, not unforgeability — `LocalCipherText:
+/// From<Vec<u8>>` is public and the constructor must be reachable cross-crate,
+/// so the cipher's tag verification on `open` remains the real authenticity
+/// guard.
 #[derive(Debug)]
-pub struct Encrypted(pub LocalCipherText);
+pub struct Encrypted(LocalCipherText);
+
+impl Encrypted {
+    /// Wrap a sealed `LocalCipherText`. Plumbing for cipher backends and derive
+    /// macros — not part of the stable public surface.
+    #[doc(hidden)]
+    pub fn from_local(ct: LocalCipherText) -> Self {
+        Self(ct)
+    }
+
+    /// Consume the leaf, yielding the inner `LocalCipherText` for decryption.
+    #[doc(hidden)]
+    pub fn into_local(self) -> LocalCipherText {
+        self.0
+    }
+}
 
 /// An authenticated "no value" marker — empty plaintext sealed under AAD.
+///
+/// Inner field is private for the same reason as [`Encrypted`].
 #[derive(Debug)]
-pub struct Absent(pub LocalCipherText);
+pub struct Absent(LocalCipherText);
+
+impl Absent {
+    /// Wrap a sealed `LocalCipherText`. Plumbing for cipher backends and derive
+    /// macros — not part of the stable public surface.
+    #[doc(hidden)]
+    pub fn from_local(ct: LocalCipherText) -> Self {
+        Self(ct)
+    }
+
+    /// Consume the marker, yielding the inner `LocalCipherText` for verification.
+    #[doc(hidden)]
+    pub fn into_local(self) -> LocalCipherText {
+        self.0
+    }
+}
 
 /// A typed value carried through the ciphertext container without
 /// encryption. Holds `T` directly — no `Box`, no `Any`.
--- a/packages/aead/src/hlist/types.rs
+++ b/packages/aead/src/hlist/types.rs
@@ -1,0 +1,35 @@
+use crate::LocalCipherText;
+
+use super::HList;
+
+/// A single sealed value (nonce + ciphertext + tag).
+#[derive(Debug)]
+pub struct Encrypted(pub LocalCipherText);
+
+/// An authenticated "no value" marker — empty plaintext sealed under AAD.
+#[derive(Debug)]
+pub struct Absent(pub LocalCipherText);
+
+/// A typed value carried through the ciphertext container without
+/// encryption. Holds `T` directly — no `Box`, no `Any`.
+#[derive(Debug)]
+pub struct Passthrough<T>(pub T);
+
+/// A map of typed `(key, value)` entries. The HList `L` encodes the
+/// structural shape of the map at the type level.
+#[derive(Debug)]
+pub struct Map<L: HList>(pub L);
+
+/// A heterogeneous sequence — length and per-element type known at compile
+/// time. For homogeneous, runtime-length sequences use a plain `Vec` of
+/// leaf types (e.g. `Vec<Encrypted>`).
+#[derive(Debug)]
+pub struct Seq<L: HList>(pub L);
+
+/// One map entry. `key` is a `&'static str`; `value` is one of the leaf
+/// or composite types above.
+#[derive(Debug)]
+pub struct Entry<V> {
+    pub key: &'static str,
+    pub value: V,
+}
--- a/packages/encrypt/src/cipher_static.rs
+++ b/packages/encrypt/src/cipher_static.rs
@@ -20,30 +20,44 @@ impl StaticCipher for &Aes256Cipher {
         A: IntoAad<'a>,
     {
         let local = seal_into_local(self, data, aad)?;
-        Ok(Encrypted(local))
+        Ok(Encrypted::from_local(local))
     }
 
     fn encrypt_none<'a, A>(self, aad: A) -> Result<Absent, Self::Error>
     where
         A: IntoAad<'a>,
     {
-        let local = seal_into_local(self, Protected::new(Vec::new()), aad)?;
-        Ok(Absent(local))
+        let local = seal_into_local(self, Protected::new(Vec::new()), absent_aad(aad))?;
+        Ok(Absent::from_local(local))
     }
+}
+
+/// Domain-separate the absence marker from an empty [`Encrypted`].
+///
+/// Both seal an empty plaintext, so without this their `LocalCipherText` bytes
+/// would be identical under the same `(key, AAD)` and an `Absent` slot could be
+/// swapped for an empty `Encrypted` one (and vice versa) without tripping tag
+/// verification. Prefixing a fixed label binds the "absent" role into the tag
+/// (PAE-encoded via the tuple `IntoAad` impl), so the two no longer
+/// cross-validate. `verify_absent` re-derives the same wrapped AAD.
+fn absent_aad<'a, A: IntoAad<'a>>(aad: A) -> (&'a [u8], A) {
+    (b"vitaminc:absent", aad)
 }
 
 impl Aes256Cipher {
     /// Open an [`Encrypted`] leaf under the supplied AAD. Counterpart to
     /// [`StaticCipher::encrypt_bytes`].
     ///
-    /// The plaintext is returned inside `Protected` so the zeroize-on-drop
-    /// guarantee survives the call boundary, matching
-    /// [`Aes256Cipher::decrypt_with_aad`].
+    /// The plaintext is returned inside `Protected` to thread the chain of
+    /// custody across the call boundary, matching
+    /// [`Aes256Cipher::decrypt_with_aad`]. Note the zeroize-on-drop guarantee
+    /// is not yet in effect — `Protected<T>` has no `Drop` impl today; that is
+    /// tracked in #181.
     pub fn open<'a, A>(&self, ct: Encrypted, aad: A) -> Result<Protected<Vec<u8>>, Unspecified>
     where
         A: IntoAad<'a>,
     {
-        open_local(self, ct.0, aad)
+        open_local(self, ct.into_local(), aad)
     }
 
     /// Verify an [`Absent`] marker under the supplied AAD. Returns
@@ -54,8 +68,8 @@ impl Aes256Cipher {
     {
         // Sealed plaintext is empty by construction; we only care about the
         // tag verification. The returned `Protected<Vec<u8>>` drops at the
-        // end of this expression.
-        open_local(self, ct.0, aad).map(|_| ())
+        // end of this expression. The AAD is wrapped to match `encrypt_none`.
+        open_local(self, ct.into_local(), absent_aad(aad)).map(|_| ())
     }
 }
 
@@ -92,7 +106,7 @@ where
     A: IntoAad<'a>,
 {
     let aad = aad.into_aad();
-    let (nonce, reader) = ct.into_reader().read_nonce::<NONCE_LEN>();
+    let (nonce, reader) = ct.into_reader().read_nonce::<NONCE_LEN>()?;
     let nonce_bytes = nonce.into_inner();
 
     reader
@@ -104,8 +118,11 @@ where
 #[allow(clippy::unwrap_used)]
 mod test {
     use super::*;
+    use crate::key::tests::DifferingKeyPair;
     use crate::Key;
+    use quickcheck_macros::quickcheck;
     use vitaminc_aead::hlist::{Entry, HCons, HNil, Map, Passthrough, StaticCipher};
+    use vitaminc_aead::LocalCipherText;
     use vitaminc_protected::Controlled;
 
     // A worked example. In production this type alias would be generated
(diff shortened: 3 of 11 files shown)


packages/encrypt/src/cipher_static.rs
```diff
@@ -0,0 +1,225 @@
...
+fn open_local<'a, A>(
+    cipher: &Aes256Cipher,
+    ct: vitaminc_aead::LocalCipherText,
+    aad: A,
+) -> Result<Protected<Vec<u8>>, Unspecified>
+where
+    A: IntoAad<'a>,
+{
+    let aad = aad.into_aad();
+    let (nonce, reader) = ct.into_reader().read_nonce::<NONCE_LEN>();
```
auxesis: **M-01 (Medium) — Panic on under-length input**

`read_nonce::<NONCE_LEN>` calls `array::from_fn(|_| buf.get_u8())` (`packages/aead/src/ciphertext/read_monads.rs:13-21`). `bytes::Buf::get_u8` panics on an empty buffer, so any `Encrypted` / `Absent` carrying fewer than 12 bytes triggers a panic here.

Reachability: `LocalCipherText: From<Vec<u8>>` is public and `Encrypted(pub LocalCipherText)` has a `pub` field, so any caller can construct a too-short `Encrypted` and pass it to `cipher.open(...)`:

```rust
let bad = Encrypted(LocalCipherText::from(vec![1u8, 2, 3]));
let _ = cipher.open(bad, b"any-aad".as_slice()); // PANIC
```

Today the static path has no in-tree serde, so this requires same-process construction — but the whole point of this PR is to be the target of a future derive macro that deserializes from untrusted bytes.

Suggested fix — a length guard at the entry of `open_local`:

```rust
fn open_local<'a, A: IntoAad<'a>>(...) -> Result<Protected<Vec<u8>>, Unspecified> {
    if ct.as_ref().len() < NONCE_LEN + 16 {  // nonce + GCM tag minimum
        return Err(Unspecified);
    }
    // ... rest unchanged
}
```

Same `Unspecified` opaque-error contract as every other failure mode. Alternatively, push the guard into `read_nonce` itself to also fix the latent equivalent in `AesDecipher::decrypt_local_ciphertext`.

packages/encrypt/src/cipher_static.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+}
+
+impl Aes256Cipher {
+    /// Open an [`Encrypted`] leaf under the supplied AAD. Counterpart to
+    /// [`StaticCipher::encrypt_bytes`].
+    ///
+    /// The plaintext is returned inside `Protected` so the zeroize-on-drop
+    /// guarantee survives the call boundary, matching
+    /// [`Aes256Cipher::decrypt_with_aad`].
```
auxesis: **M-02 (Medium) — Doc claims a zeroize-on-drop guarantee that doesn't fire**

This (and the parallel claim on `StaticCipher::encrypt_bytes` at `packages/aead/src/hlist/cipher.rs:23-27`) currently overstates the behaviour. In `packages/protected/src/protected/mod.rs`:

```rust
#[derive(Zeroize, OpaqueDebug)]                       // no ZeroizeOnDrop derive
pub struct Protected<T>(pub(crate) T);
// ...
impl<T: Zeroize> ZeroizeOnDrop for Protected<T> {}    // marker only — empty body
```

`ZeroizeOnDrop` is a *marker trait* in the `zeroize` crate; implementing it doesn't generate a `Drop` impl. No `impl<T> Drop for Protected<T>` exists. Dropping a `Protected<Vec<u8>>` therefore does *not* zeroize the inner buffer — plaintext bytes persist in the freed allocation until reuse.

This is the #181 issue acknowledged in the PR description ("the leaf zeroize won't actually fire until #181 is fixed"). Flagging because the doc comment here makes a stronger claim than the dynamic-side equivalents and will mislead future readers (including future audit passes).

Suggested fix — either:

- **Land #181 first.** Smallest change is to derive the trait instead of marker-impl'ing it:
  ```rust
  #[derive(Zeroize, ZeroizeOnDrop, OpaqueDebug)]
  pub struct Protected<T>(pub(crate) T);
  // remove the manual `impl ZeroizeOnDrop` below
  ```
  Once that lands, this doc becomes true.

- **Or temper the doc** to reflect reality:
  > The plaintext is returned inside `Protected` to thread the chain of custody across the call boundary. The zeroize-on-drop guarantee is tracked separately in #181 and is not yet in effect.

packages/aead/src/hlist/types.rs
```diff
@@ -0,0 +1,35 @@
+use crate::LocalCipherText;
+
+use super::HList;
+
+/// A single sealed value (nonce + ciphertext + tag).
+#[derive(Debug)]
+pub struct Encrypted(pub LocalCipherText);
```
auxesis: **L-01 (Low) — `pub` field defeats newtype provenance**

`pub LocalCipherText` here (and on `Absent` at line 11) means any caller can construct an `Encrypted` from arbitrary bytes:

```rust
let forged = Encrypted(LocalCipherText::from(vec![]));      // arbitrary content
```

Combined with `LocalCipherText: From<Vec<u8>>` being public, the newtype doesn't carry a "this came from a real seal" provenance guarantee — only a nominal Rust-type tag. Direct interaction with M-01: this is why the panic is trivially reachable.

Suggested fix — private field with crate-local construction:

```rust
pub struct Encrypted(LocalCipherText);
impl Encrypted {
    pub(crate) fn new(ct: LocalCipherText) -> Self { Self(ct) }
    pub(crate) fn consume(self) -> LocalCipherText { self.0 }
}
// same for Absent
```

Update `cipher_static.rs:21` to `Ok(Encrypted::new(local))`, `cipher_static.rs:31` to `Ok(Absent::new(local))`, and `cipher_static.rs:46, 58` to `open_local(self, ct.consume(), aad)`. External callers keep `Encrypted` / `Absent` values; they lose the ability to manufacture them.

packages/encrypt/src/cipher_static.rs
```diff
@@ -0,0 +1,225 @@
...
+    {
+        let local = seal_into_local(self, data, aad)?;
+        Ok(Encrypted(local))
+    }
+
+    fn encrypt_none<'a, A>(self, aad: A) -> Result<Absent, Self::Error>
+    where
+        A: IntoAad<'a>,
+    {
+        let local = seal_into_local(self, Protected::new(Vec::new()), aad)?;
```
auxesis: **L-02 (Low) — `Encrypted{empty plaintext}` and `Absent` are byte-identical**

`encrypt_bytes(Protected::new(Vec::new()), aad)` and this `encrypt_none(aad)` both route through `seal_into_local` with an empty plaintext under the *same* AAD. Their `LocalCipherText` output shapes are byte-identical: `nonce(12) ‖ ∅ ‖ tag(16)` = 28 bytes.

The distinction between the two is purely the Rust newtype wrapper. The dynamic `AesCipherText` enum gets a runtime variant tag from serde; the static side discards that discriminator.

No serializer exists yet for the static types, so today the attack requires same-process bytes-mutation — but that's exactly the threat model that derive-macro `Decrypt` impls will eventually expose. An attacker who can swap an `Encrypted` slot with an `Absent` slot (within the same `(key, AAD)`) would have `cipher.open` on the swapped bytes succeed and return an empty `Vec<u8>`, where the caller expected a real plaintext.

Suggested fix — domain-separate the AAD inside `encrypt_none`:

```rust
fn encrypt_none<'a, A: IntoAad<'a>>(self, aad: A) -> Result<Absent, Self::Error> {
    // Bind the absence marker into the AAD so an Absent's tag does not
    // validate as an Encrypted's tag under the same (key, AAD).
    let none_aad = (b"vitaminc:absent" as &[u8], aad);  // tuple → PAE-encoded
    let local = seal_into_local(self, Protected::new(Vec::new()), none_aad)?;
    Ok(Absent(local))
}
```

`verify_absent` re-derives the same `none_aad`. An attempt to `open` an `Absent` (or `verify_absent` an `Encrypted` of empty plaintext) will then fail tag verification cryptographically, not just at the Rust type level.

Alternative: document that the static path is in-memory-only and any future serialization PR must add a variant tag at the bytes level.

packages/aead/Cargo.toml
```diff
@@ -19,6 +19,19 @@ bytes = { version = "^1", features = ["serde"] }
 serde = { workspace = true }
 zeroize = { workspace = true }
 
+[features]
+test-utils = []
```
auxesis: **L-03 (Low) — Empty `test-utils` feature**

```bash
$ grep -rn 'feature = "test-utils"' packages/
# (no matches)
```

No `#[cfg(feature = "test-utils")]` use anywhere in the PR or in the wider worktree. Either the feature is a placeholder for a follow-up — in which case I'd suggest removing it until the follow-up actually adds gated code, since feature surface is API surface — or it slipped in by accident.

If it's intentional, a one-line comment above describing the intent would help future readers.

auxesis (commented):
Ran an analysis with @trailofbits's `audit-context-building` and `vulnerability-hunting` skills. 

Findings:

- 0 critical
- 0 high
- **2 medium**
- **3 low**
- 5 informational
- No `unsafe`, no new constant-time hazards. 

I have not posted the informational ones as they're not important enough to fix.

Findings are commented inline.

packages/encrypt/src/cipher_static.rs
auxesis: **CG-01 (Important) — Panic-on-under-length input is untested**

Tied to M-01 (same anchor — these two travel together). No test in the suite feeds a malformed `Encrypted` / `Absent` into `open` / `verify_absent`, so the `read_nonce` panic is silent in CI today. Once M-01's length guard lands, these tests double as regression guards:

```rust
#[test]
fn open_returns_err_when_too_short_for_nonce() {
    let cipher = Aes256Cipher::new(&Key::from([1u8; 32])).unwrap();
    for n in [0usize, 5, 11] {
        let bad = Encrypted(LocalCipherText::from(vec![0u8; n]));
        assert!(cipher.open(bad, b"aad".as_slice()).is_err());
    }
}

#[test]
fn open_returns_err_when_too_short_for_tag() {
    let cipher = Aes256Cipher::new(&Key::from([1u8; 32])).unwrap();
    for n in [12usize, 15, 20, 27] {
        let bad = Encrypted(LocalCipherText::from(vec![0u8; n]));
        assert!(cipher.open(bad, b"aad".as_slice()).is_err());
    }
}
```

While M-01 is unfixed, the first test will panic instead of returning `Err`. Mark it `#[ignore = "tracks M-01"]` or wrap it in `#[should_panic]` for the interim — then flip to `is_err()` once the guard ships.

packages/encrypt/src/cipher_static.rs
auxesis: **CG-02 (Important) — `Encrypted{empty}` ↔ `Absent` confusion is untested**

Tied to L-02 (same anchor). The current tests confirm `verify_absent` rejects a wrong AAD, but no test confirms that `verify_absent` rejects an `Encrypted` of empty plaintext under the *matching* AAD (or that `open` rejects an `Absent` symmetrically). Both succeed today because the bytes are byte-identical; after the L-02 domain-separation fix, both should fail — the test guards both directions.

```rust
#[test]
fn absent_does_not_validate_as_encrypted_empty() {
    let cipher = Aes256Cipher::new(&Key::from([2u8; 32])).unwrap();
    let aad = b"shared-aad".as_slice();
    let absent = (&cipher).encrypt_none(aad).unwrap();

    // Currently PASSES the open (the bug). After L-02 it should fail.
    let as_encrypted = Encrypted(absent.0);
    assert!(cipher.open(as_encrypted, aad).is_err());
}

#[test]
fn encrypted_empty_does_not_validate_as_absent() {
    let cipher = Aes256Cipher::new(&Key::from([2u8; 32])).unwrap();
    let aad = b"shared-aad".as_slice();
    let encrypted_empty = (&cipher)
        .encrypt_bytes(Protected::new(Vec::new()), aad)
        .unwrap();

    let as_absent = Absent(encrypted_empty.0);
    assert!(cipher.verify_absent(as_absent, aad).is_err());
}
```

Like CG-01 these depend on `Encrypted` / `Absent` still being constructible from the outside. Once L-01 lands and the field goes private, swap the direct construction for a `#[cfg(test)] pub(crate) fn from_local_for_test(...)` helper.

packages/encrypt/src/cipher_static.rs
```diff
@@ -0,0 +1,225 @@
...
+    /// [`StaticCipher::encrypt_bytes`].
+    ///
+    /// The plaintext is returned inside `Protected` so the zeroize-on-drop
+    /// guarantee survives the call boundary, matching
+    /// [`Aes256Cipher::decrypt_with_aad`].
+    pub fn open<'a, A>(&self, ct: Encrypted, aad: A) -> Result<Protected<Vec<u8>>, Unspecified>
+    where
+        A: IntoAad<'a>,
+    {
+        open_local(self, ct.0, aad)
```
auxesis: **CG-03 (Important) — Wrong-key case is untested**

All current negative tests (`static_open_fails_with_wrong_aad`, `static_verify_absent_fails_with_wrong_aad`) vary the AAD against the same cipher. None vary the key. A regression in key handling — say, a refactor that accidentally zeroes the key in some constructor path, or a feature flag that silently swaps backends — wouldn't be caught.

```rust
#[test]
fn open_with_wrong_key_fails() {
    let cipher_a = Aes256Cipher::new(&Key::from([1u8; 32])).unwrap();
    let cipher_b = Aes256Cipher::new(&Key::from([2u8; 32])).unwrap();
    let aad = b"aad".as_slice();

    let ct = (&cipher_a)
        .encrypt_bytes(Protected::new(b"hello".to_vec()), aad)
        .unwrap();

    assert!(cipher_b.open(ct, aad).is_err());
}

#[test]
fn verify_absent_with_wrong_key_fails() {
    let cipher_a = Aes256Cipher::new(&Key::from([1u8; 32])).unwrap();
    let cipher_b = Aes256Cipher::new(&Key::from([2u8; 32])).unwrap();
    let aad = b"aad".as_slice();

    let absent = (&cipher_a).encrypt_none(aad).unwrap();
    assert!(cipher_b.verify_absent(absent, aad).is_err());
}
```

packages/encrypt/src/cipher_static.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    #[test]
+    fn static_open_fails_with_wrong_aad() {
+        let key = Key::from([3u8; 32]);
+        let cipher = Aes256Cipher::new(&key).expect("cipher init");
+        let leaf = (&cipher)
+            .encrypt_bytes(Protected::new(b"hello".to_vec()), b"right".as_slice())
+            .expect("encrypt");
+        assert!(cipher.open(leaf, b"wrong".as_slice()).is_err());
```
auxesis, line 214: **CG-04 (Important) — Only AAD mismatch is tested as a negative**

This test covers the AAD axis of GCM's authentication contract. The other three axes — nonce, ciphertext body, tag — are untested. Any byte flip in any of those should fail tag verification, and the canonical AEAD test matrix should exercise all four.

```rust
fn flip_byte_at(bytes: &[u8], idx: usize) -> Vec<u8> {
    let mut v = bytes.to_vec();
    v[idx] ^= 1;
    v
}

#[test]
fn open_fails_on_tampered_nonce() {
    let cipher = Aes256Cipher::new(&Key::from([3u8; 32])).unwrap();
    let ct = (&cipher)
        .encrypt_bytes(Protected::new(b"hello".to_vec()), b"aad".as_slice())
        .unwrap();
    let bytes = ct.0.as_ref();
    for idx in 0..NONCE_LEN {
        let tampered = Encrypted(LocalCipherText::from(flip_byte_at(bytes, idx)));
        assert!(cipher.open(tampered, b"aad".as_slice()).is_err());
    }
}

#[test]
fn open_fails_on_tampered_ciphertext_body() {
    // same shape, idx in NONCE_LEN..bytes.len()-16
}

#[test]
fn open_fails_on_tampered_tag() {
    // same shape, idx in bytes.len()-16..bytes.len()
}
```

Requires `crate::backend::NONCE_LEN` to be reachable from this test module (already is, via `use super::*` + the `pub(crate)` constant). After L-01 (private `pub` field on `Encrypted`), the construction step swaps for the test-only helper noted on CG-02.

packages/encrypt/src/cipher_static.rs
```diff
@@ -0,0 +1,225 @@
...
+
+fn seal_into_local<'a, A>(
+    cipher: &Aes256Cipher,
+    data: Protected<Vec<u8>>,
+    aad: A,
+) -> Result<vitaminc_aead::LocalCipherText, Unspecified>
+where
+    A: IntoAad<'a>,
+{
+    let nonce = cipher.nonce_generator.generate()?;
```
auxesis, line 84: **CG-05 (Important) — No determinism / nonce-uniqueness assertion**

AES-GCM is non-deterministic: two encryptions of the same plaintext under the same `(key, AAD)` must produce distinct ciphertexts because the nonce is freshly drawn here on each call. Nothing in the current test suite catches a regression in `RandomNonceGenerator` — e.g. a constant seed accidentally shipped into production, a refactor that breaks `from_entropy`, or a feature flag that swaps the RNG for a test stub.

```rust
#[test]
fn encrypt_bytes_is_nondeterministic() {
    let cipher = Aes256Cipher::new(&Key::from([4u8; 32])).unwrap();
    let aad = b"aad".as_slice();
    let pt = || Protected::new(b"hello".to_vec());

    let ct_1 = (&cipher).encrypt_bytes(pt(), aad).unwrap();
    let ct_2 = (&cipher).encrypt_bytes(pt(), aad).unwrap();

    assert_ne!(ct_1.0.as_ref(), ct_2.0.as_ref(),
               "same plaintext + AAD should yield distinct ciphertexts");
    // Specifically, the nonce prefixes must differ:
    assert_ne!(&ct_1.0.as_ref()[..NONCE_LEN], &ct_2.0.as_ref()[..NONCE_LEN]);
}

#[test]
fn encrypt_none_is_nondeterministic() {
    let cipher = Aes256Cipher::new(&Key::from([5u8; 32])).unwrap();
    let aad = b"aad".as_slice();

    let a_1 = (&cipher).encrypt_none(aad).unwrap();
    let a_2 = (&cipher).encrypt_none(aad).unwrap();
    assert_ne!(a_1.0.as_ref(), a_2.0.as_ref());
}
```

Low risk of being flaky: collisions in 96-bit random output are at ~2⁻⁹⁶ per pair, so the `assert_ne!` is reliable to many sigma.

auxesis (commented):
Follow-up coverage-gap review on the same PR head. Posted as a separate review so the test-coverage items triage independently of the M/L findings in https://github.com/cipherstash/vitaminc/pull/175#pullrequestreview-4381628970.

**Summary:** the new code (~511 LOC across 4 files) has 4 `#[test]` functions + 1 compile-time `Send` proof. `quickcheck` is already a dev-dep on this crate but the static path adds no property tests. 5 coverage gaps marked **Important** (CG-01 through CG-05) are inlined below — together they roughly triple the static-cipher test surface (4 → ~12 tests) and close every finding-adjacent gap. CG-01/CG-02 are paired with M-01/L-02 respectively; CG-03/CG-04/CG-05 stand on their own.

10 more coverage gaps (5 **Useful** — empty `encrypt_bytes`, varied AAD shapes, quickcheck roundtrip, `hlist!`/`hlist_pat!` macros, empty `Map<HNil>` — and 5 **Nice-to-have** — `Seq<L>` has no producer in tree, sealed trait compile-fail, 3+ level nesting, wasm32 backend, large plaintext) are excluded from this batch. Happy to follow up with those on request.

packages/encrypt/src/cipher_static.rs
coderdan: Fixed in 164a7f3 — guarded at the source: `read_nonce` now returns `Result<_, Unspecified>` and rejects `len < N` rather than panicking in `get_u8`. That covers both the static `open`/`verify_absent` and the latent dynamic-path equivalent (`AesDecipher::decrypt_local_ciphertext`) — the alternative you suggested. Regression-guarded by the `open_rejects_arbitrary_bytes` property, which shrinks to the empty buffer if it ever regresses.

packages/encrypt/src/cipher_static.rs
coderdan: Fixed in 164a7f3 — tempered both docs (here and `hlist/cipher.rs`) to say the plaintext is returned in `Protected` to thread the chain of custody, but the zeroize-on-drop guarantee is not yet in effect and is tracked in #181.

packages/aead/src/hlist/types.rs
coderdan: Done in 164a7f3 — `Encrypted`/`Absent` inner fields are now private with `#[doc(hidden)] pub from_local`/`into_local`.

Caveat worth recording: the leaves live in `aead` but are constructed in `encrypt` (and, later, by a derive macro), so the constructor must be reachable cross-crate. That makes this encapsulation rather than true unforgeability — `LocalCipherText: From<Vec<u8>>` stays public, the same posture as the dynamic `AesCipherText::Single`. The panic it made reachable is closed by the M-01 guard regardless, so the residual is API-hygiene only.

packages/encrypt/src/cipher_static.rs
coderdan: Fixed in 164a7f3 — `encrypt_none` now binds `b"vitaminc:absent"` into the AAD via the PAE tuple `IntoAad` impl, and `verify_absent` re-derives the same wrapped AAD. An `Absent` and an empty `Encrypted` no longer cross-validate under the same (key, AAD). Guarded both directions by `absent_does_not_validate_as_encrypted_empty` / `encrypted_empty_does_not_validate_as_absent`.

packages/aead/Cargo.toml
coderdan: Removed in 164a7f3 — it was an accidental leftover (no `cfg(feature = "test-utils")` anywhere).

packages/encrypt/src/cipher_static.rs
coderdan: Added in 164a7f3 as `open_rejects_arbitrary_bytes` / `verify_absent_rejects_arbitrary_bytes` — quickcheck properties over an arbitrary `Vec<u8>`. They cover the under-length range and, because quickcheck shrinks, will report the empty buffer as the minimal failing case if the M-01 guard regresses. Went with a property rather than the fixed-length list.

packages/encrypt/src/cipher_static.rs
coderdan: Added in 164a7f3 — `absent_does_not_validate_as_encrypted_empty` and `encrypted_empty_does_not_validate_as_absent`, guarding both directions of the empty-Encrypted/Absent confusion after the domain-separation fix.

packages/encrypt/src/cipher_static.rs
coderdan: Added in 164a7f3 as quickcheck properties `open_with_wrong_key_fails` / `verify_absent_with_wrong_key_fails` over `DifferingKeyPair`, mirroring the dynamic side's wrong-key test.

packages/encrypt/src/cipher_static.rs
coderdan: Added in 164a7f3 — `open_fails_on_tampering_any_byte` flips every byte across the nonce, ciphertext body, and tag and asserts each fails verification. Kept as an exhaustive deterministic sweep rather than a sampled property, since the full sweep is strictly more thorough.

packages/encrypt/src/cipher_static.rs
coderdan: Added in 164a7f3 as quickcheck properties `encrypt_bytes_is_nondeterministic` / `encrypt_none_is_nondeterministic` — distinct ciphertexts and distinct nonce prefixes. Running them as properties samples many nonce pairs per run.

coderdan:
Thanks @auxesis — addressed the full review in 164a7f3, and **rebased onto `main`** now that #174 has merged (PR base retargeted from `feat/encrypt-some-none-passthrough` → `main`; the diff is now just the hlist work plus the shared `read_nonce` fix).

**Mediums**
- **M-01** (panic on under-length input): guarded at the source — `read_nonce` is now fallible and rejects `len < N`. Fixes both the static path and the latent dynamic-path equivalent already on `main`.
- **M-02** (zeroize doc): tempered to reference #181.

**Lows**
- **L-01**: `Encrypted`/`Absent` fields private with `#[doc(hidden)]` constructors. Encapsulation, not unforgeability (cross-crate constraint — see thread); the panic it enabled is closed by M-01 anyway.
- **L-02**: `encrypt_none` AAD domain-separated with `b"vitaminc:absent"`.
- **L-03**: unused `test-utils` feature removed.

**Coverage (CG-01…05)**: added — arbitrary-byte rejection, Encrypted/Absent cross-validation (both directions), wrong-key, exhaustive byte-tamper sweep, and non-determinism. The wrong-key, non-determinism, and arbitrary-byte cases are quickcheck properties; the cross-validation and tamper cases are deterministic. Static-cipher test surface goes 4 → 13.

Verified: base tests 29 aead / 52 encrypt, `--features hlist` 65, clippy clean both ways, fmt clean.

The 5 informational items weren't posted, so nothing pending there — shout if you'd like any of them followed up.

auxesis (approved):
Nice work, thanks @coderdan.

PR #190, opened 2026-06-05
feat(aead): re-introduce ContextTag on the revised Cipher/Decipher traits

coderdan:
PR #148 redesigned the Cipher/Encrypt/Decipher/Decrypt traits and dropped the
old ContextTag, which referenced the pre-redesign API. Re-introduce it as an
encrypt-side wrapper so the type system, not the call site, guarantees a value's
context tag is always bound as AAD — the one piece of behaviour raw tuple AAD
cannot enforce.

In the revised API, AAD is threaded through Encrypt at seal time but injected
into the Decipher by the concrete cipher at open time, so ContextTag implements
Encrypt and folds (extra_aad, tag) into the AAD, while ContextTag::aad /
aad_with rebuild the matching AAD for the cipher's decrypt entry point. refine
nests tags into a PAE-encoded tuple for hierarchical context.

The Encrypt impl bounds the tag on IntoAad<'static> (owned tags and &'static
str) and re-borrows the encoded Aad for the call lifetime via covariance,
sidestepping the HRTB that an Encrypt-level lifetime used to carry.

Covered by mock-cipher unit tests in vitaminc-aead and an end-to-end AES-256-GCM
roundtrip suite in vitaminc-encrypt (wrong/omitted/refined context all fail).

### DIFF
--- a/packages/aead/src/context.rs
+++ b/packages/aead/src/context.rs
@@ -1,0 +1,449 @@
+//! Context-tagged encryption.
+//!
+//! This module provides [`ContextTag`], a wrapper that binds additional authenticated data (AAD)
+//! to a plaintext value **at the type level**. When a `ContextTag` is encrypted, the embedded tag
+//! is automatically folded into the AAD alongside any extra AAD passed to
+//! [`encrypt_with_aad`](crate::Encrypt::encrypt_with_aad).
+//!
+//! The point of the wrapper is the type-level guarantee: a value wrapped in `ContextTag` *cannot*
+//! be encrypted without its context. The revised [`Cipher`]/[`Encrypt`] API already accepts tuple
+//! AAD (e.g. `value.encrypt_with_aad(cipher, (tag, extra))`), so the byte-level behaviour is
+//! nothing more than [`IntoAad`] composition — but tuple AAD is opt-in at every call site and easy
+//! to forget. `ContextTag` moves the obligation into the type system so the compiler enforces it.
+//!
+//! # Where AAD lives in the revised API
+//!
+//! Encryption threads AAD through [`Encrypt::encrypt_with_aad`](crate::Encrypt::encrypt_with_aad),
+//! so `ContextTag` participates by implementing [`Encrypt`]. Decryption is different: the revised
+//! [`Decrypt`]/[`Decipher`] traits do **not** thread AAD through the type being decrypted — the
+//! concrete cipher injects the AAD when it constructs the decipher (e.g.
+//! `cipher.decrypt_with_aad::<T, _>(ciphertext, aad)`). There is therefore nothing for a
+//! `ContextTag` *type* to enforce on the decrypt path; the recovered value is simply `T`.
+//!
+//! To keep the two sides symmetric without guessing the tuple layout, [`ContextTag::aad`] and
+//! [`ContextTag::aad_with`] build the exact AAD the wrapper bound at encrypt time, ready to hand to
+//! a concrete cipher's decrypt entry point.
+//!
+//! # AAD encoding
+//!
+//! When encrypting a `ContextTag`, the final AAD is formed by combining the extra AAD with the
+//! embedded tag as the tuple `(extra_aad, tag)`, which [`IntoAad`] PAE-encodes:
+//!
+//! ```text
+//! final_aad = PAE(extra_aad, tag)
+//! ```
+//!
+//! PAE (Pre-Authentication Encoding) prefixes each piece with its length, preventing
+//! canonicalisation attacks where different inputs could otherwise produce identical byte strings.
+//!
+//! With [`refine`](ContextTag::refine), the tag itself is a nested tuple that is recursively
+//! PAE-encoded:
+//!
+//! ```text
+//! ContextTag::new(data, "a").refine("b")
+//!   ──► tag = ("a", "b")
+//!   ──► final_aad = PAE(extra_aad, PAE("a", "b"))
+//! ```
+
+use crate::{Aad, Cipher, Encrypt, IntoAad};
+
+/// A wrapper that pairs a plaintext value with a context tag used as additional authenticated
+/// data (AAD).
+///
+/// `ContextTag` guarantees, at the type level, that `tag` is folded into the AAD whenever `inner`
+/// is encrypted. This is the recommended way to enforce that a value is always sealed against a
+/// specific context (a user id, table name, column name, …): because the tag is part of the type,
+/// the compiler will not let you forget it.
+///
+/// The tag can be any type that implements [`IntoAad<'static>`](IntoAad) — `&'static str`,
+/// `String`, `u64`, `Vec<u8>`, and tuples of these. (Non-`'static` borrowed tags such as a
+/// short-lived `&str` are intentionally excluded; promote them to an owned `String` or a
+/// `&'static str`.)
+///
+/// # Encrypting
+///
+/// `ContextTag` implements [`Encrypt`], so it is encrypted like any other value. The embedded tag
+/// is bound automatically:
+///
+/// ```rust,ignore
+/// use vitaminc_aead::{ContextTag, Encrypt};
+///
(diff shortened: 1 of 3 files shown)


packages/aead/src/context.rs
```diff
@@ -0,0 +1,684 @@
...
+///   *inside another* `ContextTag` (`ContextTag::new(ContextTag::new(v, "b"), "a")`) folds the AAD
+///   twice into `((extra, "a"), "b")`, which the symmetric [`context`](ContextTag::context) +
+///   `refine` decrypt path *cannot* reconstruct (it produces `(extra, ("a", "b"))`). Always layer
+///   hierarchy with `ContextTag::new(v, "a").refine("b")`.
+/// - **`ContextTag` has no [`Decrypt`] impl**, only an [`Encrypt`] impl. It therefore composes on
+///   the encrypt side (you *can* build `Vec<ContextTag<…>>` and encrypt it) but a `ContextTag`
+///   nested inside another structure has no symmetric decrypt path — decrypt only at the top level
+///   via [`context`](ContextTag::context) / [`aad`](ContextTag::aad). Don't bury a `ContextTag`
+///   inside a collection or struct you intend to read back.
+/// - **A unit tag `()` still binds a (non-empty) tag.** `ContextTag::new(v, ())` seals against
```
auxesis, line 137: **Gap:** The documented unit-tag boundary — `ContextTag::new(v, ())` seals against `PAE(empty, empty)`, *not* empty AAD — has no test. A regression that short-circuited a `()` tag to empty AAD (so `cipher.decrypt(ct)` with no AAD would succeed) would silently contradict this doc. Pin the byte-level claim alongside the sibling `encrypt_with_no_extra_aad_still_binds_tag`:

```rust
#[test]
fn unit_tag_still_binds_nonempty_aad() {
    let cipher = MockCipher::new();
    ContextTag::new("secret", ())
        .encrypt(&cipher)
        .expect("encryption should succeed");
    // A `()` tag binds PAE(empty, empty) — distinct from empty AAD.
    assert_eq!(cipher.captured_aad(), ((), ()).into_aad().as_bytes());
    assert_ne!(cipher.captured_aad(), Aad::empty().as_bytes());
}
```

Expected: passes today; fails if a `()` tag is ever folded to empty AAD.

auxesis (commented):
## Test coverage review — `ContextTag`

**Verdict:** The new `ContextTag` code is exceptionally well tested. Byte-level AAD binding (mock-cipher unit tests in `aead/src/context.rs`) and end-to-end roundtrip/tamper behaviour against real AES-256-GCM (`encrypt/tests/context_tag.rs`) are both covered across `new`, `refine`, chained `refine`, `into_parts`, `aad`/`aad_with`, `context`, the `decrypt`/`decrypt_with_aad` helpers, and the `Encrypt` impl — including a strong negative matrix (wrong tag, wrong extra AAD, omitted context, partial/unrefined tag, swapped `aad_with` args, composite values, integer tags). The folding logic is branch-free, so there are no untested branches. Only two small, defensible gaps remain (inline), plus a few minor notes below.

No security issue noticed.

## Additional coverage gaps not posted inline

- **Nested-`ContextTag` footgun asymmetry** (`packages/aead/src/context.rs:128`): the docs claim wrapping a `ContextTag` inside another folds the AAD twice into `((extra,"a"),"b")`, which the symmetric `context`+`refine` path (`(extra,("a","b"))`) cannot reconstruct. That documented asymmetry is never pinned — a mock-cipher unit test capturing the AAD of `ContextTag::new(ContextTag::new(v,"b"),"a").encrypt(&cipher)` would guard the claim against a future change to the layout. Lower value (it tests a documented "don't").
- **Fresh-nonce nondeterminism**: the new real-cipher e2e harness never asserts that two seals of the same `(value, tag)` differ. This is primarily the cipher's responsibility (`ContextTag` only adds AAD), and `AesCipherText` derives only `Debug` (no `PartialEq`), so a drop-in `assert_ne!` isn't possible without a serialization helper — hence not posted inline.
- **`Vec<u8>` tag type** (`packages/aead/src/context.rs`): the docs list `Vec<u8>` as a supported tag type, but only `&str`/`String`/`u64`/tuples are exercised.
- **Wrong-*inner*-refine via the helper** (`packages/encrypt/tests/context_tag.rs`): `context("a").refine("WRONG")` (correct length, wrong inner value) isn't distinctly tested; the wrong-tag and partial-refine axes that exist largely subsume it.

packages/encrypt/tests/context_tag.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    let plaintext: String = cipher
+        .decrypt_with_aad(ciphertext, ContextTag::aad_with("row:99", "table:users"))
+        .expect("decryption failed");
+
+    assert_eq!(plaintext, "secret");
+}
+
+#[test]
+fn decrypt_fails_with_wrong_context() {
```
auxesis, line 47: **Gap:** This negative-test cluster exercises every *AAD* tamper axis (wrong tag, wrong extra AAD, omitted context, unrefined tag, swapped `aad_with` args) but never the *key* axis — the classic "wrong-AAD covered, wrong-key missing" lopsided pattern. `encrypt/src/cipher.rs` already establishes a `decrypt_fails_with_wrong_key` convention; mirror it here so the e2e tamper matrix is complete for a `ContextTag`-sealed value:

```rust
#[test]
fn decrypt_fails_with_wrong_key() {
    let ciphertext = ContextTag::new("secret", "user:42")
        .encrypt(&cipher())
        .expect("encryption failed");

    // Same context, different key: GCM authentication must reject it.
    let other = Aes256Cipher::new(&Key::from([7u8; 32])).expect("failed to create cipher");
    let result: Result<String, _> =
        other.decrypt_with_aad(ciphertext, ContextTag::aad("user:42"));
    assert!(result.is_err(), "wrong key must not decrypt");
}
```

Expected: passes today; fails if a `ContextTag` ciphertext were ever decryptable under the wrong key.

auxesis (approved):
Thanks @coderdan, approved with comments about test gaps.

packages/aead/src/context.rs
coderdan: Added in ff29275 as `unit_tag_still_binds_nonempty_aad` (your exact suggestion), right after `encrypt_with_no_extra_aad_still_binds_tag`. Asserts the bound AAD is `((), ()).into_aad()` and `!= Aad::empty()`. Verified green locally.

packages/encrypt/tests/context_tag.rs
coderdan: Added in ff29275 as `decrypt_fails_with_wrong_key` in the e2e negative cluster — seals under `cipher()` (key 42) and decrypts under a fresh key (7) with the *correct* context; GCM rejects it. Verified green locally.

coderdan:
Thanks for the thorough coverage review @auxesis. Both inline gaps are addressed in ff29275, and I also picked up three of the four "additional notes":

- **`Vec<u8>` tag type** → `vec_u8_tag_is_accepted` (mock-cipher, mirrors the `&str`/`String` cases).
- **Nested-`ContextTag` asymmetry** → `nested_context_tag_folds_aad_twice` — captures the AAD of `ContextTag::new(ContextTag::new(v, "b"), "a")` and pins it to `(((), "a"), "b")`, asserting it is `!=` the symmetric refine layout `((), ("a", "b"))`. The documented footgun is now guarded.
- **Wrong-*inner*-refine** → `refined_context_fails_with_wrong_inner_aad` — correct outer tag and correct tuple *shape* but a wrong inner value, distinct from the inner-omitted `refined_context_fails_with_unrefined_aad`.

I **skipped the fresh-nonce nondeterminism** note, agreeing with your inline reasoning: nonce freshness is the cipher’s responsibility (`ContextTag` only contributes AAD), and `AesCipherText` derives only `Debug` (no `PartialEq`), so an `assert_ne!` on two seals would need a serialization helper that does not belong in this PR. Happy to file a follow-up against the cipher crate if you think the seal-determinism property is worth a dedicated test there.

All five additions are tests only — no production code changed. fmt + clippy clean.

PR #191, opened 2026-06-06
refactor(aead)!: thread AAD through the decrypt path to mirror encrypt

coderdan:
## What & why

Makes the **decrypt path symmetric with the encrypt path**. Today AAD is baked into the concrete `AesDecipher` at construction, while encryption threads it per-call through `Cipher`/`Encrypt`. That asymmetry meant there was no generic decrypt entry point and blocked a clean `ContextTag::decrypt_with_aad` helper.

AAD is now threaded per-call:

- `Decipher::decrypt_bytes`/`decrypt_seq`/`decrypt_map`/`decrypt_option` take an `aad: A` parameter.
- New **`Decrypt::decrypt_with_aad`** — the dual of `Encrypt::encrypt_with_aad`; `Decrypt::decrypt` now defaults to it with `Aad::empty()`.
- `Aes256Cipher::decrypt_with_aad` delegates to `T::decrypt_with_aad` (the `AesDecipher` no longer carries an `aad` field); `AesSeqAccess`/`AesMapAccess` re-supply their stored AAD per element.

## Deliberately kept

The **visitor pattern** (`DecipherVisitor` / `SeqAccess` / `MapAccess`), the **`Ok<T>` GAT**, and the **infallible `map_ok`** are retained — they are the correct pull-side duals of the encrypt-side builder, not incidental indirection. Per-element AAD rides on the `Decipher` methods so `SeqAccess`/`MapAccess` signatures are untouched.

## Breaking change

Breaking to the `Decipher`/`Decrypt` traits. These are unreleased (v0.2.0-pre), so there are no external consumers. The public `cipher.decrypt(ct)` / `cipher.decrypt_with_aad(ct, aad)` entry points are unchanged.

## Verification

- `cargo test -p vitaminc-aead -p vitaminc-encrypt` — 29 + 52 unit, all doctests, all ~40 roundtrip/negative cipher tests (incl. wrong-AAD, wrong-shape, option-tag-auth) pass.
- `cargo build --workspace`, `cargo test -p vitaminc-aead --all-features` (hlist), `cargo fmt --check` all pass.

## Follow-up

Unblocks a `ContextTag::decrypt_with_aad` helper built on `Decrypt::decrypt_with_aad` (separate PR).

### DIFF
--- a/packages/encrypt/src/cipher.rs
+++ b/packages/encrypt/src/cipher.rs
@@ -579,6 +579,27 @@ mod test {
     }
 
     #[quickcheck]
+    fn roundtrip_vec_with_aad(key: Key, plaintext: Vec<String>) -> bool {
+        // Positive counterpart to `decrypt_seq_fails_with_wrong_aad`: every element
+        // must authenticate when the *correct* AAD is re-supplied per element by
+        // `AesSeqAccess::next_element`. Exercises the borrowed-AAD seq path across
+        // arbitrary element counts (the empty vec has no element tags and trivially
+        // roundtrips). A wrong-AAD-fails test alone can't catch a regression where
+        // the per-element AAD is dropped or mis-borrowed so even the correct AAD
+        // fails — only this positive multi-element roundtrip does.
+        let aad = "seq-aad";
+        let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
+        let ciphertext = plaintext
+            .clone()
+            .encrypt_with_aad(&cipher, aad)
+            .expect("Encryption failed");
+        let decrypted: Vec<String> = cipher
+            .decrypt_with_aad(ciphertext, aad)
+            .expect("Decryption failed");
+        decrypted == plaintext
+    }
+
+    #[quickcheck]
     fn decrypt_fails_with_wrong_key(keys: DifferingKeyPair, plaintext: String) -> bool {
         // `DifferingKeyPair` guarantees the two keys are distinct, so this test
         // is deterministic — a key collision cannot make it spuriously fail.
--- a/packages/encrypt/src/cipher.rs
+++ b/packages/encrypt/src/cipher.rs
@@ -261,17 +261,30 @@ impl Aes256Cipher {
         // AAD is threaded through the `Decrypt`/`Decipher` call chain (mirroring how
         // `Encrypt::encrypt_with_aad` threads it on the encrypt side), not baked into the
         // decipher up front.
-        T::decrypt_with_aad(
-            AesDecipher {
-                cipher: self,
-                ciphertext,
-            },
-            aad,
-        )
+        T::decrypt_with_aad(self.decipher(ciphertext), aad)
+    }
+
+    /// Construct a [`Decipher`] over `ciphertext` bound to this cipher.
+    ///
+    /// This is the decrypt-side counterpart to passing `&cipher` (a [`Cipher`])
+    /// on the encrypt side: it hands callers a concrete [`Decipher`] they can
+    /// drive directly via [`Decrypt::decrypt_with_aad`], which is what generic
+    /// decrypt-side helpers (e.g. `ContextTag`) build on. The ergonomic
+    /// [`decrypt`](Aes256Cipher::decrypt) /
+    /// [`decrypt_with_aad`](Aes256Cipher::decrypt_with_aad) methods are thin
+    /// wrappers around it.
+    pub fn decipher(&self, ciphertext: AesCipherText) -> AesDecipher<'_> {
+        AesDecipher {
+            cipher: self,
+            ciphertext,
+        }
     }
 }
 
-struct AesDecipher<'c> {
+/// A [`Decipher`] over a single [`AesCipherText`], produced by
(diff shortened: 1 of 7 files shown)


packages/encrypt/src/cipher.rs
```diff
@@ -381,21 +412,23 @@ impl<'c> Decipher<'c> for AesDecipher<'c> {
...
 }
 
-struct AesSeqAccess<'c> {
+struct AesSeqAccess<'c, 'a> {
     cipher: &'c Aes256Cipher,
     items: std::vec::IntoIter<AesCipherText>,
-    aad: Vec<u8>,
+    // Held as `Aad` (copy-on-write) rather than an owned `Vec<u8>` so a borrowed
+    // AAD stays borrowed; the per-element `clone` below is then cheap for the
+    // common `&str`/`&[u8]` case.
```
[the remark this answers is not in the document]
coderdan: Already addressed in cf5bcf4 — the comment was reworded to drop the stale "per-element clone" wording and now states there's no per-element allocation in either the borrowed (`&str`/`&[u8]`) or owned (`Vec`/PAE) case. Good catch; the same point came up in our own review pass. Resolving.

packages/encrypt/src/cipher.rs
```diff
@@ -524,6 +551,25 @@ mod test {
             .is_err()
     }
 
+    #[quickcheck]
+    fn decrypt_seq_fails_with_wrong_aad(key: Key, plaintext: Vec<String>) -> bool {
```
auxesis, line 563: **Gap:** The sequence path gained a negative wrong-AAD test but no positive AAD roundtrip — the same lopsided shape flagged in prior audits (wrong-AAD covered, correct-AAD roundtrip missing). The Single (`roundtrip_string_with_aad`) and Map (`roundtrip_hashmap_with_aad`) paths both have a positive AAD roundtrip; the sequence path only has the empty-AAD `roundtrip_vec_of_strings`. A wrong-AAD-fails test can't catch a regression where `AesSeqAccess::next_element` re-supplies the borrowed AAD incorrectly so that even the *correct* AAD fails per element — only a positive multi-element roundtrip does.

```rust
#[quickcheck]
fn roundtrip_vec_with_aad(key: Key, plaintext: Vec<String>) -> bool {
    // Positive counterpart to `decrypt_seq_fails_with_wrong_aad`: every element
    // must authenticate when the *correct* AAD is re-supplied per element by
    // `AesSeqAccess::next_element`. Exercises the borrowed-AAD seq path across
    // arbitrary element counts (the empty vec has no element tags and trivially
    // roundtrips).
    let aad = "seq-aad";
    let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
    let ciphertext = plaintext
        .clone()
        .encrypt_with_aad(&cipher, aad)
        .expect("Encryption failed");
    let decrypted: Vec<String> = cipher
        .decrypt_with_aad(ciphertext, aad)
        .expect("Decryption failed");
    decrypted == plaintext
}
```

Expected: passes — confirms the sequence path roundtrips with non-empty AAD. Would fail if the per-element AAD were dropped, mis-borrowed, or applied to only a subset of elements such that correct-AAD decryption errors.

auxesis (commented):
## Test coverage review

This PR threads AAD through the decrypt path — supplied per call and *borrowed* per element in `AesSeqAccess`/`AesMapAccess` instead of cloned into the decipher up front. The new `decrypt_seq_fails_with_wrong_aad` test is a welcome addition, and the Single, Map, and Option paths each already carry **both** a positive AAD roundtrip and a negative wrong-AAD test.

The sequence path is the one asymmetry: it gained the negative case but no positive AAD roundtrip. This is the classic lopsided anti-pattern — a wrong-AAD-fails test alone cannot catch a bug where the per-element borrowed AAD is supplied incorrectly such that the *correct* AAD also fails to authenticate. One inline comment below.

Things I checked that are **not** gaps:
- `Aes256Cipher::decipher()` / the newly-exported `AesDecipher` — exercised transitively (every `decrypt_with_aad` routes through `self.decipher(...)`, and `decrypt_passthrough_via` drives it directly).
- The generic `aead` `Decrypt` impls (`impls.rs`) and the new default `Decrypt::decrypt` method — covered through the `Aes256Cipher` roundtrip/wrong-AAD tests (empty-AAD path via `Aad::empty()` is hit by every `cipher.decrypt(...)`).
- Map per-element borrowing — covered by the existing `roundtrip_hashmap_with_aad` + `decrypt_hashmap_fails_with_wrong_aad` pair.

No crypto/security concerns noted incidentally.

auxesis (approved):
Thanks for this @coderdan. 

Approved, with some comments about missing test coverage to be fixed before merge.

packages/encrypt/src/cipher.rs
coderdan: Added in 29b4472 as `roundtrip_vec_with_aad`, right after `decrypt_seq_fails_with_wrong_aad` — your exact positive multi-element seq roundtrip. Verified green locally (`cargo test -p vitaminc-encrypt --all-features roundtrip_vec_with_aad`). The seq path now has both the wrong-AAD-fails and correct-AAD-roundtrip cases, matching the Single/Map paths. Thanks for the catch.

PR #195, opened 2026-06-07
feat(aead): extract a reusable CipherTree codec for structured ciphers

coderdan:
## What

Factors the recursive ciphertext **structure** out of `Aes256Cipher` into reusable, crypto-agnostic machinery in `vitaminc-aead`, so any cipher producing the same shape (single value / sequence / map / option / passthrough) reuses it and supplies only its per-leaf seal/open.

New in `vitaminc-aead`:
- `CipherTree<L>` — the recursive container, generic over the leaf payload `L`.
- `LeafSealer` / `LeafOpener` — the two small per-cipher operations.
- `TreeCipher<S>` / `TreeDecipher<O>` (+ `TreeSeq`/`TreeMap` and the seq/map access drivers) — generic `Cipher`/`Decipher` impls that perform the whole structural traversal.

`Aes256Cipher` is refactored onto it: `AesCipherText = CipherTree<LocalCipherText>`, `AesDecipher = TreeDecipher<AesOpener>`, and the hand-written `Seq`/`Map` ciphers, `Decipher` impl, and access structs are deleted (**~290 fewer lines**).

## Why

The structural drive/walk was duplicated verbatim between `Aes256Cipher` and the ZeroKMS-backed cipher being built in cipherstash-suite (`stack-encrypt`). This is the shared foundation that cipher can build on — sealing each leaf under its own data key while reusing everything here.

## Design note: why not a blanket impl

The generic impls are on the concrete local types `TreeCipher`/`TreeDecipher`, **not** `impl<C: LeafSealer> Cipher for &C`. A conditional blanket would conflict (E0119) with any direct `Cipher`/`Decipher` impl — and those traits are public extension points meant to be implemented directly (e.g. by `stack-encrypt`, and by the test doubles here). Concrete ciphers wrap themselves in a `LeafSealer`/`LeafOpener` instead.

## Verification

- Behaviour unchanged: the full `vitaminc-encrypt` suite passes on **both** the aws-lc-rs and RustCrypto (`_test-rust-crypto-backend`) backends — incl. the map-contract, option-shape, passthrough, AAD-mismatch, and nonce-uniqueness tests.
- `vitaminc-aead` 42 unit + 8 doctests; `cargo build --workspace --all-features` ✅; clippy `-D warnings` ✅; `fmt --check` ✅.

## Stacking

> Stacked on **#190** (base `claude/vitaminc-aead-issue-178-vCypS`). Merge the stack bottom-up, or retarget to main once #190 lands.

### DIFF
--- a/packages/aead/src/tree.rs
+++ b/packages/aead/src/tree.rs
@@ -1,5 +1,5 @@
-//! A reusable recursive ciphertext tree and the generic [`Cipher`]/[`Decipher`]
-//! machinery that drives it.
+//! A reusable recursive ciphertext tree, a structural encrypt codec, and the
+//! generic [`Decipher`] machinery that drives it.
 //!
 //! Most ciphers that produce a structured ciphertext (a single sealed value, a
 //! sequence, a map, …) share the *same* structural traversal — only the per-leaf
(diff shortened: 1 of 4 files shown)


packages/aead/src/tree.rs
```diff
@@ -0,0 +1,448 @@
...
+    ) -> Result<CipherTree<M>, E> {
+        Ok(match self {
+            CipherTree::Single(leaf) => CipherTree::Single(f(leaf)?),
+            CipherTree::None(leaf) => CipherTree::None(f(leaf)?),
+            CipherTree::Sequence(items) => {
+                let mut out = Vec::with_capacity(items.len());
+                for item in items {
+                    out.push(item.try_map_leaves(f)?);
+                }
+                CipherTree::Sequence(out)
```
coderdan: Self-review: could this be done with a combinator/iterator instead?

packages/aead/src/tree.rs
```diff
@@ -0,0 +1,448 @@
...
+            }
+            CipherTree::Map(entries) => {
+                let mut out = Vec::with_capacity(entries.len());
+                for (k, v) in entries {
+                    out.push((k, v.try_map_leaves(f)?));
+                }
+                CipherTree::Map(out)
+            }
```
coderdan: Same question here

packages/aead/src/tree.rs
coderdan: Done in 652695e — the `Sequence` arm is now:

```rust
CipherTree::Sequence(items) => CipherTree::Sequence(
    items
        .into_iter()
        .map(|item| item.try_map_leaves(f))
        .collect::<Result<Vec<_>, E>>()?,
),
```

`collect::<Result<_, _>>()` short-circuits on the first `Err` exactly like the old `for` + `?`, so behaviour is identical.

packages/aead/src/tree.rs
coderdan: Same — the `Map` arm in 652695e:

```rust
CipherTree::Map(entries) => CipherTree::Map(
    entries
        .into_iter()
        .map(|(k, v)| v.try_map_leaves(f).map(|v| (k, v)))
        .collect::<Result<Vec<_>, E>>()?,
),
```

The `.map(|v| (k, v))` re-pairs each transformed value with its (cleartext) key before collecting into the `Result`.

PR #217, opened 2026-06-22
ci: add cargo-mutants in-diff gate, config, and mise task

coderdan:
## What

Wires **cargo-mutants** (mutation testing) into the workspace as a permanent gate, mirroring how we set up cargo-crap (#205):

- **`.github/workflows/mutants.yml`** — a per-PR gate (`🧬 Mutants gate`) that runs `cargo mutants --in-diff` over **only the lines a PR changes**, posts a sticky comment with the results, and **fails when a changed line has a surviving (or timed-out) mutant**. A full-workspace run (~600 mutants) is far too slow for a per-PR gate, so `--in-diff` keeps it fast and relevant. Reuses crap.yml's LocalStack + cargo-binstall idioms (the unmutated baseline runs the `vitaminc-kms` tests, which need LocalStack).
- **`.cargo/mutants.toml`** — shared config: `--all-features` (so feature-gated code like `hlist` / `_test-rust-crypto-backend` is actually exercised, not falsely "missed"), excludes the proc-macro derive crates (their logic runs at compile time, not in the instrumented test binary), and generous timeouts for the property/round-trip tests.
- **`mise.toml`** — adds the `cargo:cargo-mutants` tool and a `mutants` task for a full local run.
- **`.gitignore`** — ignores the `mutants.out` output dir.

## Why

Mutation testing rewrites small pieces of logic (flip `<` to `<=`, replace a body with `Default::default()`, …) and reruns the tests; a surviving mutant is a line the suite doesn't actually pin down — exactly the comparisons, AAD handling, and length checks a crypto library most wants covered before it ships. We did the one-off mutation **inventory** earlier and fixed the Tier-1 gaps it surfaced (merged via #212); this PR wires the tool in so those gains don't regress and new PRs are held to the same bar.

## How it behaves

- **In-diff only:** mutates just the PR's added/changed lines, so the gate trips only when a PR introduces logic its own tests don't catch.
- **Gate:** fails the check on any surviving/timed-out mutant in the diff; unviable mutants (mutations that don't compile) are not failures.
- **Sticky comment:** posts a caught/missed/unviable/timeout table plus the list of surviving mutants to fix.

## Validation

Ran `cargo mutants --in-diff` locally end-to-end against a sample diff: it generates mutants only for the changed lines, the baseline + mutant runs complete, and the gate/comment logic produces the right pass/fail. Confirmed `--all-features` is threaded into the cargo invocations (so feature-gated code isn't falsely reported as surviving). Workflow YAML and TOML validated.

## To make it blocking

Mark **`🧬 Mutants gate`** as a required status check in the repo ruleset (same as the CRAP gate).

https://claude.ai/code/session_015jJwUcMbyjuttxTLh4vh6q

### DIFF
--- a/.cargo/mutants.toml
+++ b/.cargo/mutants.toml
@@ -1,0 +1,34 @@
+# Configuration for cargo-mutants — mutation testing for the vitaminc workspace.
+#
+# Mutation testing rewrites small pieces of logic (flip a `<` to `<=`, replace a
+# body with `Default::default()`, etc.) and reruns the tests: a mutant that
+# survives is a line the suite does not actually pin down. In a crypto library
+# that is exactly the logic — comparisons, AAD handling, length checks — we most
+# want a test to catch before it ships.
+#
+# CI runs this `--in-diff` as a per-PR gate (.github/workflows/mutants.yml): only
+# the lines a PR changes are mutated, so the gate trips when a PR adds logic its
+# tests don't exercise. `mise run mutants` runs the whole workspace locally.
+# Both read the settings below, so they stay in sync.
+
+# Build and test with every feature on, so feature-gated code — the `hlist`
+# static-ciphertext API and the `_test-rust-crypto-backend` — is actually
+# compiled and exercised. Without this, a mutant in that code would "survive"
+# only because the gating feature was off, not because a test is missing.
+additional_cargo_args = ["--all-features"]
+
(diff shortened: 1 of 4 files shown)


auxesis (approved):
Thanks for this @coderdan. 

@copilot's feedback actually looks accurate, which is genuinely surprising given how terribly wrong its analysis of GitHub Actions workflows usually are. 

Approved, with the issues identified by @copilot requiring fixes.

PR #222, opened 2026-06-23
feat(aead): implement Encrypt/Decrypt for the Equatable controlled type

### LINKED ISSUE #221: Implement Encrypt/Decrypt for the Equatable controlled type
## Summary

`Equatable<T>` (from `vitaminc-protected`) does not implement the `aead` `Encrypt`/`Decrypt` traits, so an `Equatable`-wrapped secret can't be carried through the encryption pipeline the way `Protected<T>` can. You currently have to manually unwrap to the inner value before encrypting and re-wrap after decrypting.

## Background

This supersedes the original issue cipherstash/vitaminc#36 (closed as *not planned*) and PR cipherstash/vitaminc#37 (now closed). That PR could not be revived: it targeted the pre-restructure `aead/src/encrypt.rs` and `decrypt.rs` single files and the old `Encrypt { type Encrypted }` / `Decrypt` trait shapes, all of which were replaced by the current driver-style `Encrypt` and GAT-based `Decipher`/`Decrypt<'c>` design. It was ~270 commits behind `main`, so a fresh implementation against the current structure is the right path.

## Proposed implementation

Mirror the existing `impl ... for Protected<T>` in both `encrypt/impls.rs` and `decipher/impls.rs`, accounting for `Equatable<T>` being a `Controlled` wrapper over an inner `T: Controlled` (`Inner = T::Inner`):

* `Encrypt for Equatable<T>` where `T: Controlled, T::Inner: Encrypt` — `risky_unwrap()` to the inner value and forward; leaf impls re-wrap in `Protected` so the zeroize chain of custody is preserved. `Equatable` adds nothing to the ciphertext.
* `Decrypt<'c> for Equatable<T>` where `T: Controlled, T::Inner: Decrypt<'c>` — decrypt the innermost value, then rebuild both wrapper layers via `Controlled::init_from_inner`.

Add a `roundtrip_equatable_string` quickcheck test alongside `roundtrip_protected_string` in `vitaminc-encrypt`, covering both the aws-lc and rust-crypto backends.

## Notes

`Equatable` is an in-memory constant-time-equality wrapper, so the wrapper layer is reconstructed structurally on decrypt — there is no on-the-wire representation distinguishing it from `Protected`.

coderdan:
## Summary

Implements `aead`'s `Encrypt`/`Decrypt` traits for `Equatable<T>`, so an `Equatable`-wrapped secret can flow through the encryption pipeline the way `Protected<T>` already can — no more manually unwrapping before encrypting and re-wrapping after decrypting.

Closes #221.

## Background

This supersedes the original issue #36 (closed *not planned*) and the stale PR #37. That PR couldn't be revived: it targeted the pre-restructure `aead/src/encrypt.rs` / `decrypt.rs` single files and the old `Encrypt { type Encrypted }` / `Decrypt` trait shapes, all since replaced by the current driver-style `Encrypt` and GAT-based `Decipher`/`Decrypt<'c>` design. This is a fresh implementation against the current structure.

## Changes

Mirrors the existing `Protected<T>` impls, accounting for `Equatable<T>` being a `Controlled` wrapper over an inner `T: Controlled` (`Inner = T::Inner`):

- **`Encrypt for Equatable<T>`** (`encrypt/impls.rs`) — `where T: Controlled, T::Inner: Encrypt`. Unwraps to the inner value and forwards; leaf impls re-wrap in `Protected` before crossing the cipher boundary, preserving the zeroize chain of custody.
- **`Decrypt<'c> for Equatable<T>`** (`decipher/impls.rs`) — `where T: Controlled + Send + 'c, T::Inner: Decrypt<'c> + 'c`. Decrypts the innermost value, then rebuilds both wrapper layers via `Controlled::init_from_inner`.

`Equatable` is an in-memory constant-time-equality wrapper and adds nothing to the ciphertext, so its layer is reconstructed structurally on decrypt — there's no on-the-wire representation distinguishing it from `Protected`.

## Testing

- `roundtrip_equatable_string` — round-trips `Equatable<Protected<String>>` (the byte-vec path).
- `roundtrip_equatable_u32` — round-trips `Equatable<Protected<u32>>`, exercising the fixed-size `encrypt_bytes_array` path under the wrapper.
- `decrypt_equatable_fails_with_wrong_aad` — negative-AAD case for the wrapper.
- All pass on **both** the aws-lc and rust-crypto backends; `clippy -D warnings` and `cargo fmt --check` clean.

## Review fixes (post-open)

Addressed code-review feedback:

- **Tightened the `Decrypt` bound** from `Self: Send` to the more precise `T: Send`. `Self: Send` was the loosest patch that satisfied the `Decrypt: Send` supertrait; constraining the wrapper directly (the way `Protected<T>` derives `Send` from its inner) removes a latent *encryptable-but-not-decryptable* asymmetry for any `Controlled` type with a `Send` inner but non-`Send` outer.
- **Expanded test coverage** beyond the single `Equatable<Protected<String>>` case: added the fixed-size-inner (`u32`) round-trip and the wrong-AAD negative test described above.

A possible follow-up (not done here): the `Encrypt`/`Decrypt` bodies are generic over `Controlled` and will be copy-pasted for the other wrappers (`Exportable`, `Usage`). A declarative macro over the wrapper list would collapse the duplication — the `equatable/mod.rs` "blanket impl for all Paranoid types?" TODO already marks the spot.

https://claude.ai/code/session_018swQ3Fa1GrWdpTQp1Dmj1s

### DIFF
--- a/packages/aead/src/decipher/impls.rs
+++ b/packages/aead/src/decipher/impls.rs
@@ -1,7 +1,7 @@
 use super::{Decipher, DecipherVisitor, Decrypt, MapAccess, SeqAccess};
 use crate::{IntoAad, Unspecified};
 use std::collections::HashMap;
-use vitaminc_protected::{Controlled, Protected};
+use vitaminc_protected::{Controlled, Equatable, Protected};
 use zeroize::Zeroize;
 
 impl<'c> Decrypt<'c> for Vec<u8> {
@@ -142,6 +142,34 @@ where
         D::map_ok(
             T::decrypt_with_aad(decipher, aad),
             Protected::init_from_inner,
+        )
+    }
+}
+
+impl<'c, T> Decrypt<'c> for Equatable<T>
+where
+    // `Send` is required by the `Decrypt: Send` supertrait and is load-bearing,
+    // not over-tightening. Unlike the sibling `Protected` impl — which bounds
+    // `T: Decrypt`, transitively giving `T: Send` — this impl bounds the
+    // flattened `T::Inner: Decrypt`, which only proves `T::Inner: Send`. `T`
+    // itself is otherwise unconstrained (`Controlled` has no `Send` supertrait),
+    // so `T: Send` is not implied and must be stated for `map_ok`'s `U: Send`
+    // requirement (and the impl's own `Send`) to hold.
+    T: Controlled + Send + 'c,
+    T::Inner: Decrypt<'c> + 'c,
+{
+    fn decrypt_with_aad<'a, D, A>(decipher: D, aad: A) -> D::Ok<Self>
+    where
+        D: Decipher<'c>,
+        A: IntoAad<'a>,
+    {
+        // Decrypt the innermost value, then rebuild both wrapper layers
+        // (`Equatable<Protected<_>>`) via `init_from_inner` — the mirror of the
+        // `Encrypt` impl, which unwraps them. The AAD is threaded through to the
+        // inner decrypt so the binding matches what the `Encrypt` impl sealed.
+        D::map_ok(
+            <T::Inner as Decrypt<'c>>::decrypt_with_aad(decipher, aad),
+            Equatable::init_from_inner,
         )
     }
 }
--- a/packages/aead/src/encrypt/impls.rs
+++ b/packages/aead/src/encrypt/impls.rs
@@ -4,7 +4,7 @@ use crate::{
     Aad, IntoAad,
 };
 use std::collections::HashMap;
-use vitaminc_protected::{Controlled, Protected};
+use vitaminc_protected::{Controlled, Equatable, Protected};
 use zeroize::Zeroize;
 
 impl Encrypt for u32 {
@@ -104,6 +104,34 @@ where
         // boundary, so the bare-`T` stack window opened here is bounded by
         // the inner `Encrypt::encrypt_with_aad` call. Custom `Encrypt` impls
         // are responsible for their own discipline.
+        self.risky_unwrap().encrypt_with_aad(cipher, aad)
+    }
+}
+
+impl<T> Encrypt for Equatable<T>
+where
+    // Deliberately no `Zeroize` bound, unlike the sibling `Protected<T>` impl
+    // above (`T: Encrypt + Zeroize`): `Controlled` already governs the zeroize
+    // chain for `T`, and the inner value is re-wrapped in `Protected` by the leaf
+    // impl before it crosses the cipher boundary (see below), so an extra bound
+    // here would over-constrain callers without adding protection. The divergence
+    // is a conscious choice, not drift.
+    T: Controlled,
+    T::Inner: Encrypt,
+{
+    fn encrypt_with_aad<'a, C, A>(self, cipher: C, aad: A) -> Result<C::Ok, C::Error>
+    where
+        C: Cipher,
+        A: IntoAad<'a>,
+    {
+        // `Equatable` is an in-memory constant-time-equality wrapper; it adds
+        // nothing to the ciphertext — an `Equatable<Protected<T>>` ciphertext is
+        // interchangeable with a `Protected<T>` one, as the ciphertext does not
+        // attest the wrapper. Mirror the `Protected<T>` impl above: unwrap to the
+        // inner value and let the leaf `Encrypt` impl re-wrap it in `Protected`
+        // before crossing the cipher boundary, preserving the zeroize chain of
+        // custody. The `Equatable` layer is reconstructed structurally on decrypt
+        // (see the `Decrypt` impl).
         self.risky_unwrap().encrypt_with_aad(cipher, aad)
     }
 }
--- a/packages/encrypt/src/cipher.rs
+++ b/packages/encrypt/src/cipher.rs
@@ -678,6 +678,115 @@ mod test {
     }
 
     #[quickcheck]
+    fn roundtrip_equatable_string(key: Key, plaintext: String) -> bool {
+        use vitaminc_protected::{Controlled, Equatable, Protected};
+
+        let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
+        let value: Equatable<Protected<String>> = Equatable::new(plaintext.clone());
+        let ciphertext = value.encrypt(&cipher).expect("Encryption failed");
+        let decrypted: Equatable<Protected<String>> =
+            cipher.decrypt(ciphertext).expect("Decryption failed");
+        decrypted.risky_unwrap() == plaintext
+    }
+
+    #[quickcheck]
+    fn roundtrip_equatable_u32(key: Key, plaintext: u32) -> bool {
+        use vitaminc_protected::{Controlled, Equatable, Protected};
+
+        // `u32` routes through the fixed-size `encrypt_bytes_array` path, so this
+        // exercises the `Equatable` wrapper over that branch — not just the
+        // byte-vec path covered by `roundtrip_equatable_string`.
+        let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
+        let value: Equatable<Protected<u32>> = Equatable::new(plaintext);
+        let ciphertext = value.encrypt(&cipher).expect("Encryption failed");
+        let decrypted: Equatable<Protected<u32>> =
+            cipher.decrypt(ciphertext).expect("Decryption failed");
+        decrypted.risky_unwrap() == plaintext
+    }
+
+    #[quickcheck]
+    fn roundtrip_equatable_with_aad(key: Key, plaintext: String) -> bool {
+        use vitaminc_protected::{Controlled, Equatable, Protected};
+
+        // Positive correct-AAD roundtrip *through* the `Equatable` layer — the
+        // wrong-AAD test only proves rejection, never that the right AAD recovers
+        // the value. Also asserts the rebuilt wrapper's own constant-time
+        // `PartialEq` (the reason `Decrypt` reconstructs `Equatable` via
+        // `init_from_inner` rather than handing back a bare `Protected`).
+        let aad = "equatable-aad";
+        let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
+        let value: Equatable<Protected<String>> = Equatable::new(plaintext.clone());
+        let ciphertext = value
+            .encrypt_with_aad(&cipher, aad)
+            .expect("Encryption failed");
+        let decrypted: Equatable<Protected<String>> = cipher
+            .decrypt_with_aad(ciphertext, aad)
+            .expect("Decryption failed");
+        let expected: Equatable<Protected<String>> = Equatable::new(plaintext.clone());
+        decrypted == expected && decrypted.risky_unwrap() == plaintext
+    }
+
+    #[quickcheck]
+    fn decrypt_equatable_fails_with_wrong_aad(key: Key, plaintext: String) -> bool {
+        use vitaminc_protected::{Equatable, Protected};
+
+        let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
+        let value: Equatable<Protected<String>> = Equatable::new(plaintext);
+        let ciphertext = value
+            .encrypt_with_aad(&cipher, "correct-aad")
+            .expect("Encryption failed");
+        cipher
+            .decrypt_with_aad::<Equatable<Protected<String>>, _>(ciphertext, "wrong-aad")
+            .is_err()
+    }
+
+    #[quickcheck]
+    fn decrypt_equatable_fails_with_wrong_key(keys: DifferingKeyPair, plaintext: String) -> bool {
+        use vitaminc_protected::{Equatable, Protected};
+
+        // Completes the tamper matrix with the key axis (sibling `String` path has
+        // both wrong-AAD and wrong-key). `DifferingKeyPair` guarantees the two keys
+        // are distinct, so a key collision cannot make this spuriously pass.
+        let DifferingKeyPair(key_a, key_b) = keys;
+        let cipher_a = Aes256Cipher::new(&key_a).expect("Failed to create cipher A");
+        let cipher_b = Aes256Cipher::new(&key_b).expect("Failed to create cipher B");
+        let value: Equatable<Protected<String>> = Equatable::new(plaintext);
+        let ciphertext = value.encrypt(&cipher_a).expect("Encryption failed");
+        cipher_b
+            .decrypt::<Equatable<Protected<String>>>(ciphertext)
+            .is_err()
+    }
+
+    #[quickcheck]
+    fn equatable_ciphertext_is_wrapper_agnostic(key: Key, plaintext: String) -> bool {
+        use vitaminc_protected::{Controlled, Equatable, Protected};
+
+        // `Equatable` adds nothing to the ciphertext, so the two are interchangeable:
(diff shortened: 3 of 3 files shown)


packages/encrypt/src/cipher.rs
```diff
@@ -602,6 +602,47 @@ mod test {
...
+        let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
+        let value: Equatable<Protected<u32>> = Equatable::new(plaintext);
+        let ciphertext = value.encrypt(&cipher).expect("Encryption failed");
+        let decrypted: Equatable<Protected<u32>> =
+            cipher.decrypt(ciphertext).expect("Decryption failed");
+        decrypted.risky_unwrap() == plaintext
+    }
+
+    #[quickcheck]
+    fn decrypt_equatable_fails_with_wrong_aad(key: Key, plaintext: String) -> bool {
```
auxesis, line 730: **Gap:** The `Equatable` AEAD support tests the wrong-AAD tamper axis but not the wrong-key axis — the lopsided-negative-case anti-pattern. The base `String` path covers both (`decrypt_fails_with_wrong_aad` at :517 *and* `decrypt_fails_with_wrong_key` at :528), and the `DifferingKeyPair` helper already exists, so this is a near drop-in.

```rust
#[quickcheck]
fn decrypt_equatable_fails_with_wrong_key(keys: DifferingKeyPair, plaintext: String) -> bool {
    use vitaminc_protected::{Equatable, Protected};

    // `DifferingKeyPair` guarantees the two keys are distinct, so a key
    // collision cannot make this spuriously pass.
    let DifferingKeyPair(key_a, key_b) = keys;
    let cipher_a = Aes256Cipher::new(&key_a).expect("Failed to create cipher A");
    let cipher_b = Aes256Cipher::new(&key_b).expect("Failed to create cipher B");
    let value: Equatable<Protected<String>> = Equatable::new(plaintext);
    let ciphertext = value.encrypt(&cipher_a).expect("Encryption failed");
    cipher_b
        .decrypt::<Equatable<Protected<String>>>(ciphertext)
        .is_err()
}
```

Expected: passes — ciphertext sealed under `key_a` must not decrypt under the distinct `key_b`, so `decrypt` returns `Err` and the test returns `true`.

packages/encrypt/src/cipher.rs
```diff
@@ -602,6 +602,47 @@ mod test {
         decrypted.risky_unwrap() == plaintext
     }
 
+    #[quickcheck]
+    fn roundtrip_equatable_string(key: Key, plaintext: String) -> bool {
```
auxesis, line 681: **Gap:** There's a no-AAD roundtrip and a wrong-AAD *failure* test, but no positive correct-AAD roundtrip for `Equatable`. Every sibling composite has one (`roundtrip_option_some_with_aad` :666, `roundtrip_hashmap_with_aad` :557), and nothing currently confirms that `decrypt_with_aad` returns the original plaintext *through* the `Equatable` layer — the failure test only asserts it errors on mismatch.

```rust
#[quickcheck]
fn roundtrip_equatable_with_aad(key: Key, plaintext: String) -> bool {
    use vitaminc_protected::{Controlled, Equatable, Protected};

    let aad = "equatable-aad";
    let cipher = Aes256Cipher::new(&key).expect("Failed to create cipher");
    let value: Equatable<Protected<String>> = Equatable::new(plaintext.clone());
    let ciphertext = value
        .encrypt_with_aad(&cipher, aad)
        .expect("Encryption failed");
    let decrypted: Equatable<Protected<String>> = cipher
        .decrypt_with_aad(ciphertext, aad)
        .expect("Decryption failed");
    decrypted.risky_unwrap() == plaintext
}
```

Expected: passes — encrypting and decrypting under the same AAD reconstructs the original value, so `risky_unwrap()` equals the input and the test returns `true`.

auxesis (commented):
## Test-coverage review — `Equatable` AEAD support

The new `Encrypt`/`Decrypt` impls for `Equatable<T>` (in `packages/aead`) are landed with three tests in `packages/encrypt/src/cipher.rs`: a String roundtrip, a u32 roundtrip (fixed-size path), and a wrong-AAD rejection. Good coverage of both encrypt paths and the AAD tamper axis.

Two gaps stand out, both grounded in patterns this file already establishes for sibling types:

1. **Wrong-key rejection is missing.** The base `String` path tests *both* `decrypt_fails_with_wrong_aad` and `decrypt_fails_with_wrong_key` (cipher.rs:517 / :528). The `Equatable` tests cover only wrong-AAD — the classic lopsided-tamper-axis gap. The `DifferingKeyPair` quickcheck helper already exists, so the test is a near drop-in.
2. **No positive correct-AAD roundtrip.** `Option` and `HashMap` each have a `roundtrip_*_with_aad`; `Equatable` only ever encrypts-with-AAD inside the *failure* test, so nothing confirms `decrypt_with_aad` returns the original plaintext through the `Equatable` layer.

Both are flagged inline with paste-ready sketches.

## Additional coverage gaps not posted inline

- **Equality-semantics preservation (low priority).** The whole reason the `Decrypt` impl rebuilds the wrapper via `init_from_inner` (rather than the test just using `Protected`) is to preserve `Equatable`'s constant-time `PartialEq`. The roundtrip tests only assert `decrypted.risky_unwrap() == plaintext`, which checks the inner value, not the rebuilt wrapper's own equality. A one-liner asserting `decrypted == Equatable::new(plaintext)` (uses `Equatable`'s `PartialEq`, valid since `String: ConstantTimeEq`) would directly exercise the reconstructed wrapper. Marginal, since `==` reduces to the same inner comparison — noting for completeness, not blocking.
- **u32 / fixed-size negative cases** are not separately tested for `Equatable`, but the base-path wrong-AAD/wrong-key tests cover the AEAD failure mechanism generically, so this is out of scope.

No crypto/security concerns noted in the impls themselves — the `Encrypt` impl correctly defers re-wrapping to the leaf impl to preserve the zeroize chain of custody, and `Decrypt` mirrors it.

packages/aead/src/decipher/impls.rs
```diff
@@ -115,6 +115,22 @@ where
     }
 }
 
+impl<'c, T> Decrypt<'c> for Equatable<T>
+where
+    T: Controlled + Send + 'c,
```
auxesis, line 158: **Bound inconsistency vs. the sibling `Protected<T>` impl.** This adds `Send + 'c` to `T`, but `Decrypt for Protected<T>` directly above (line 109) carries no `Send`. The commit message frames this as a deliberate tightening — can you confirm `Send` is genuinely load-bearing here (i.e. required to satisfy `Equatable<T>: Send` through the `D::map_ok` GAT / `init_from_inner` fn item) rather than over-constraining callers? If `Protected`'s decrypt compiles without it, it's worth understanding why `Equatable`'s needs it. Over-tightening would needlessly block `Equatable<Protected<!Send>>` from being decrypted, even though `Protected<!Send>` decrypts fine.

packages/aead/src/encrypt/impls.rs
```diff
@@ -108,6 +108,26 @@ where
     }
 }
 
+impl<T> Encrypt for Equatable<T>
+where
+    T: Controlled,
+    T::Inner: Encrypt,
```
auxesis, line 120: **Bound inconsistency vs. the sibling `Protected<T>` impl.** `Encrypt for Protected<T>` above (line 92) requires `T: Encrypt + Zeroize`, but this `Equatable` impl uses `T: Controlled, T::Inner: Encrypt` with no `Zeroize` bound. Probably fine — `Controlled` already governs the zeroize chain — but the two sibling impls now carry different zeroize discipline in their bounds. Can you confirm dropping `Zeroize` is intentional, and that the bare innermost value's stack window is still bounded by the leaf re-wrap that the doc comment claims? (The leaf impls do `Protected::new(…)` before crossing the cipher boundary, so I believe it holds — just flagging the divergence so it's a conscious choice, not drift.)

packages/encrypt/src/cipher.rs
auxesis, line 730: **Coverage gap (non-blocking).** These cover round-trip and wrong-AAD rejection, but nothing pins the *intended* wrapper-agnostic ciphertext invariant. Because `Equatable` adds nothing to the ciphertext, a value encrypted as `Equatable<Protected<String>>` should decrypt cleanly as `Protected<String>` (and bare `String` ciphertext should be interchangeable too). A test asserting cross-type decrypt succeeds would lock that in, so a future change that accidentally introduces a wrapper tag — breaking interchangeability — gets caught here rather than silently. A one-line doc note on the type stating the ciphertext does not attest the wrapper would also help.

packages/aead/src/decipher/impls.rs
auxesis: Confirmed: the `Send` bound is **load-bearing and minimal** — it should stay.

I removed `+ Send` and `cargo check -p vitaminc-aead` fails with three errors, all rooting to the same fact: `Equatable<T>` is a newtype (`Equatable<T>(pub(crate) T)`), so `Equatable<T>: Send` ⟺ `T: Send`. Two independent parts of the contract demand it:

1. **The `Decrypt<'c>: Sized + Send` supertrait** (`decipher/mod.rs:132`) — every `Decrypt` impl must itself be `Send`.
2. **The `D::map_ok` call** — `Decipher::Ok<U>` requires `U: Send + 'c` (`mod.rs:32`) and `map_ok<T, U, F>` re-states it (`mod.rs:46-47`); the body builds `U = Equatable<T>`.

The compiler even suggests `help: consider further restricting type parameter T with trait Send`.

**Why the sibling `Protected<T>` impl gets away without it:** its bound is `T: Decrypt<'c>`, and since `Decrypt: Send`, that *transitively* yields `T: Send` for free. This impl instead bounds the flattened `T::Inner: Decrypt` (not `T: Decrypt`) plus `T: Controlled` — and `Controlled` has no `Send` supertrait, while `T::Inner: Decrypt` only proves `T::Inner: Send`, never `T: Send`. So the inner `T` is otherwise unconstrained and `Send` must be stated explicitly. It's a direct consequence of the flattened-`Inner` impl shape, not over-tightening.

**No asymmetric cost to callers:** the whole decrypt pipeline is `Send`-only by design (the async `BoxFuture<'c, …>` `Ok` backend), so this rejects nothing the rest of the framework would accept — you couldn't decrypt a non-`Send` `Protected<Rc<…>>` either.

(For context, this also refutes the earlier note on this file suggesting the `Send` bound is redundant — it isn't.)

Only suggestion: a one-line comment so the next reader doesn't repeat this investigation, e.g.:
```rust
// `Send` is required by the `Decrypt: Send` supertrait. Unlike the `Protected`
// impl (which bounds `T: Decrypt`, transitively giving `T: Send`), we bound
// `T::Inner: Decrypt`, so `T: Send` is not implied and must be stated.
```

auxesis (approved):
Looks solid, thanks @coderdan. 

Have added comments about coverage gaps, and comments that will be helpful to the next reader.

coderdan:
Rebased onto the latest `main` (linear history, two commits — the earlier merge commit dropped) and addressed the review.

**Heads-up: the rebase required a real API port, not just a replay.** This PR's `Decrypt for Equatable` impl was authored against the pre-#191 `Decrypt` trait (`fn decrypt(decipher)`). #191 ("thread AAD through the decrypt path") has since merged to `main` and reworked the trait to `fn decrypt_with_aad(decipher, aad)`. The rebase applied cleanly *textually* but didn't compile, so I ported the impl to the new signature — threading AAD through to the inner decrypt, mirroring the sibling `Protected` impl exactly. (`Encrypt` was unaffected; AAD was already threaded there.)

Review items:
- ✅ **wrong-key test** — `decrypt_equatable_fails_with_wrong_key`
- ✅ **positive correct-AAD roundtrip** — `roundtrip_equatable_with_aad` (also asserts the rebuilt wrapper's own `PartialEq`, folding in your equality-semantics note)
- ✅ **wrapper-agnostic invariant** — `equatable_ciphertext_is_wrapper_agnostic`, plus a doc note on the `Encrypt` impl
- ✅ **`Send` bound** — kept (load-bearing, per your analysis), with the clarifying comment you suggested
- ✅ **`Zeroize` bound divergence** — confirmed intentional, documented inline
- ⏭️ **u32 negative cases** — skipped (you flagged as out of scope; the base-path failure tests cover the AEAD mechanism generically)

fmt + clippy clean; all 6 `Equatable` tests pass alongside the full aead/encrypt suites.

packages/aead/src/decipher/impls.rs
```diff
@@ -115,6 +115,23 @@ where
     }
 }
 
+impl<'c, T> Decrypt<'c> for Equatable<T>
+where
+    T: Controlled + 'c,
+    T::Inner: Decrypt<'c> + 'c,
+    Self: Send,
```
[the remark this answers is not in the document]
coderdan: Investigated in depth on auxesis's sibling thread: the `Send` bound is **load-bearing, not redundant**. Unlike `Protected<T>` (which bounds `T: Decrypt`, transitively giving `T: Send`), this impl bounds the flattened `T::Inner: Decrypt`, so `T: Send` is not implied — it is required by the `Decrypt: Send` supertrait and by `map_ok`'s `U: Send`. Removing `+ Send` fails to compile (three errors). I've added a comment explaining this so it isn't re-litigated.

packages/encrypt/src/cipher.rs
coderdan: Added in 41d5234 as `decrypt_equatable_fails_with_wrong_key` — your drop-in, using the existing `DifferingKeyPair` helper. Verified green locally.

packages/encrypt/src/cipher.rs
coderdan: Added in 41d5234 as `roundtrip_equatable_with_aad`. I also folded in your equality-semantics note from the review body: it asserts `decrypted == Equatable::new(plaintext)` (exercising the rebuilt wrapper's own constant-time `PartialEq`) in addition to `risky_unwrap() == plaintext`. Verified green.

packages/aead/src/decipher/impls.rs
coderdan: Confirmed and kept — added your suggested comment (lightly expanded) in 01a1655.

One note: the rebase onto current `main` ported this impl to the reworked `decrypt_with_aad` API (it was written against the old `decrypt(decipher)` signature, pre-#191). The `Send` reasoning is unchanged — the `Decrypt: Send` supertrait and `map_ok`'s `U: Send` bound still require it, and `T::Inner: Decrypt` still only proves `T::Inner: Send`.

packages/aead/src/encrypt/impls.rs
coderdan: Confirmed intentional. `Controlled` already governs `T`'s zeroize chain, and the inner value is re-wrapped in `Protected` by the leaf `Encrypt` impl before it crosses the cipher boundary — so a `Zeroize` bound here would over-constrain callers without adding protection. Documented the divergence inline on the `where` clause in 01a1655 so it reads as a conscious choice rather than drift.

packages/encrypt/src/cipher.rs
coderdan: Added in 41d5234 as `equatable_ciphertext_is_wrapper_agnostic`: it encrypts as `Equatable<Protected<String>>` and decrypts as bare `Protected<String>`, and encrypts bare and decrypts back through the `Equatable` layer — pinning interchangeability both directions. Also added the doc note you suggested to the `Encrypt` impl: the ciphertext does not attest the wrapper. Verified green.

Issue #252, opened 2026-07-24
Sequence elements are not bound to their position or length (reorder/truncation undetectable)


coderdan:
## Problem

`AesCipherText::Sequence` is a plain `Vec<AesCipherText>` and every element is sealed against the same caller-supplied AAD. An attacker holding a stored ciphertext can therefore, without detection:

- **reorder** elements,
- **truncate** the sequence (drop trailing or interior elements),
- **duplicate** an element.

This is the same vulnerability class as the map key-swap issue fixed by binding keys into per-entry AAD via `Aad::for_map_entry` (see the map key authentication work), but sequences need a different treatment.

## Why one mechanism doesn't fit

There are two distinct sequence semantics with conflicting integrity requirements:

1. **Order-significant sequences** (tuples, positional records): the position *is* part of the meaning. The natural fix is binding each element's index — and the total length — into its AAD (e.g. `PAE(domain, aad, index, len)`). Reorder, truncation, and duplication all become detectable.

2. **Order-independent collections** (a vector of values/records encrypted for insertion into a database): decryption order cannot be guaranteed to match encryption order — query access is effectively random, and elements may be decrypted individually or in subsets. Index binding would *break* legitimate use. Integrity here likely means something else (e.g. per-element identity binding chosen by the caller, or set-level authentication when the whole collection is read back), and truncation detection may not even be meaningful when partial reads are the norm.

That implies **two distinct mechanisms** — likely surfaced as distinct types or distinct `SeqCipher` entry points, so the caller declares which semantics they want and the API can't silently give them the wrong integrity property.

## Status

Deferred for now in favour of the map-key authentication work. Should be resolved before 1.0, ideally batched with other ciphertext-format-breaking changes (map key AAD binding already breaks the map format in the same release window).

PR #253, opened 2026-07-24
feat(aead)!: authenticate map keys and accept runtime-derived keys

### LINKED ISSUE #269: Authenticate map keys in structured AEAD encryption
Keys in an encrypted map travel in the clear inside the ciphertext container, but nothing binds a key to the value sealed beneath it. An attacker holding a stored map ciphertext can swap or rename keys and decryption still succeeds, silently reassigning an encrypted value to a different field. Values are sealed under the caller's AAD alone, so the tag check has nothing to catch the tampering with.

Each map value should be sealed against an AAD derived from the entry's key, applied symmetrically on the encrypt and decrypt sides, with a domain-separation label that keeps entry AAD disjoint from user-composed tuple AAD (such as the `(aad, tag)` layout used by `ContextTag`). The contract needs to be documented on the `MapCipher`/`MapAccess` traits so third-party cipher implementations perform the same binding. Reordering entries must still succeed — maps are unordered — while any change to the key/value pairing must fail to open.

The same work should relax map keys from `&'static str` to owned/borrowed strings so maps whose keys are only known at runtime can be encrypted. That is a prerequisite for encrypting dynamically shaped values (JS objects, Python dicts, Go maps) across a future FFI boundary, and it needs an `Encrypt` impl for `HashMap<String, T>` to match the existing `Decrypt` impl.

This changes the map ciphertext format, so it should land inside the same format-breaking window as other pre-1.0 integrity work. Sequences have the analogous gap and are deliberately deferred to cipherstash/vitaminc#252.

Addressed by cipherstash/vitaminc#253.

coderdan:
## Summary

First core-trait step toward NAPI/FFI encryption support: makes encrypted maps safe and usable for runtime-keyed data.

* **Map keys are now authenticated.** Keys travel in the clear inside `AesCipherText::Map`, and were previously not bound to their values — swapping or renaming keys in a stored ciphertext went undetected, silently reassigning encrypted values to different fields. Each value is now sealed against a new public helper `Aad::for_map_entry(aad, key)` = `PAE("vitaminc/aead/map-entry/v1", aad, key)`, applied symmetrically in `AesMapCipher::encrypt_value` and `AesMapAccess::next_entry`. The domain-separation label keeps entry AAD disjoint from user-composed tuple AAD (e.g. `ContextTag`'s `(aad, tag)` layout). The contract is documented on `MapCipher` and `MapAccess` for other cipher implementations.
* **Runtime-derived map keys.** `MapCipher::encrypt_key` / `passthrough_entry` / `encrypt_entry` now accept `impl Into<Cow<'static, str>>` instead of `&'static str` (source-compatible for existing callers), and `HashMap<String, T>` gains an `Encrypt` impl so runtime-keyed maps round-trip with the existing `Decrypt` impl. This is a prerequisite for encrypting JS objects across the upcoming NAPI boundary.

## Breaking changes

* **Ciphertext format:** map ciphertexts sealed before this change no longer decrypt (per-entry AAD derivation changed).
* **API:** third-party `MapCipher` implementations must update key-parameter signatures and perform the `Aad::for_map_entry` binding on both sides.

## Deliberately out of scope

Sequences have the analogous gap — elements are not bound to their position or the sequence length, so reorder/truncation/duplication of `Vec` ciphertexts is undetectable. Ordered sequences (tuples) and order-independent collections (records encrypted for random-access database reads) need two distinct mechanisms, so this is deferred to cipherstash/vitaminc#252 — ideally resolved before 1.0 within the same format-breaking window as this change.

## Testing

* New tamper tests: swapped keys fail decryption; renamed key fails; reordering entries *without* touching key↔value pairing still succeeds (maps are unordered).
* New round-trip test for runtime `String` keys with AAD.
* Byte-level pinning tests for the `for_map_entry` encoding (exact PAE bytes, disjointness from tuple AAD, key sensitivity/injectivity).
* Full workspace suite, clippy, and `cargo fmt --check` clean. (`vitaminc-kms` integration test requires LocalStack on :4566 and is unaffected by this change.)

Closes #269

### DIFF
--- a/packages/aead/src/aad/mod.rs
+++ b/packages/aead/src/aad/mod.rs
@@ -72,6 +72,51 @@ impl<'a> Aad<'a> {
         const MAP_ENTRY_DOMAIN: &[u8] = b"vitaminc/aead/map-entry/v1";
         pae::encode(&[MAP_ENTRY_DOMAIN, self.as_bytes(), key.as_bytes()])
     }
+
+    /// Derives the AAD a structural marker must be sealed against.
+    ///
+    /// Markers are sealed empty plaintexts whose tag is the only thing
+    /// authenticating a structural fact — "this sequence is empty", "this
+    /// map is empty", "this value is absent". Like [`for_map_entry`], the
+    /// encoding is a labelled three-piece `PAE(domain, aad, kind)`: the
+    /// leading domain label keeps marker AAD disjoint from user-supplied
+    /// composite AAD (a two-piece `PAE(label, aad)` would collide with a
+    /// caller binding the tuple `(label, aad)` at the top level), and the
+    /// `kind` piece keeps the marker kinds disjoint from each other, so a
+    /// stored marker can never be replayed as a different structural claim.
+    ///
+    /// [`Cipher::encrypt_none`](crate::Cipher::encrypt_none),
+    /// [`SeqCipher::end`](crate::SeqCipher::end) and
+    /// [`MapCipher::end`](crate::MapCipher::end) implementations are
+    /// required to derive marker AAD through these methods — both sides
+    /// must use them, or nothing decrypts.
+    ///
+    /// [`for_map_entry`]: Aad::for_map_entry
+    fn for_marker(&self, kind: &[u8]) -> Aad<'static> {
+        const MARKER_DOMAIN: &[u8] = b"vitaminc/aead/marker/v1";
+        pae::encode(&[MARKER_DOMAIN, self.as_bytes(), kind])
+    }
+
+    /// Marker AAD for an empty sequence — see [`for_marker`](Aad::for_marker)
+    /// (private; this method and its siblings are the public surface).
+    pub fn for_empty_sequence(&self) -> Aad<'static> {
+        self.for_marker(b"empty-sequence")
+    }
+
+    /// Marker AAD for an empty map.
+    pub fn for_empty_map(&self) -> Aad<'static> {
+        self.for_marker(b"empty-map")
+    }
+
+    /// Marker AAD for an authenticated absent value (`Option::None`).
+    ///
+    /// Domain separation here is what stops a `Single` leaf sealed under the
+    /// bare AAD from being re-tagged as a `None` marker (silent authenticated
+    /// data deletion) — and, symmetrically, an absence marker from validating
+    /// as an encrypted empty byte string.
+    pub fn for_none(&self) -> Aad<'static> {
+        self.for_marker(b"none")
+    }
 }
 
 /// Types that can be canonically converted into an [`Aad`].
@@ -282,6 +327,45 @@ mod tests {
         let bound = Aad::from_slice(b"ctx").for_map_entry("name");
         let tuple = (Aad::from_slice(b"ctx"), "name").into_aad();
         assert_ne!(bound.as_bytes(), tuple.as_bytes());
+    }
+
+    #[test]
+    fn marker_aads_pin_encoding() {
+        // The exact bytes are a wire-format commitment: PAE(domain, aad, kind).
+        // Changing them breaks decryption of existing markers.
+        let aad = Aad::from_slice(b"ctx");
+        for (derived, kind) in [
+            (aad.for_empty_sequence(), b"empty-sequence".as_slice()),
+            (aad.for_empty_map(), b"empty-map"),
+            (aad.for_none(), b"none"),
+        ] {
+            let expected = Aad::pae(&[b"vitaminc/aead/marker/v1", b"ctx", kind]);
+            assert_eq!(derived.as_bytes(), expected.as_bytes());
+        }
+    }
+
+    #[test]
+    fn marker_aads_differ_from_tuple_aad_and_each_other() {
+        // The domain label keeps marker AAD disjoint from a user binding a
+        // matching (label, aad) tuple at the top level, and the kind piece
+        // keeps the three markers disjoint from one another.
+        let aad = Aad::from_slice(b"ctx");
+        let markers = [
+            aad.for_empty_sequence(),
+            aad.for_empty_map(),
+            aad.for_none(),
+        ];
+        for (i, m) in markers.iter().enumerate() {
+            let tuple = (
+                b"vitaminc/aead/marker/v1".as_slice(),
+                Aad::from_slice(b"ctx"),
+            )
+                .into_aad();
+            assert_ne!(m.as_bytes(), tuple.as_bytes());
+            for other in &markers[i + 1..] {
+                assert_ne!(m.as_bytes(), other.as_bytes());
+            }
+        }
     }
 
     #[test]
--- a/packages/aead/src/aad/mod.rs
+++ b/packages/aead/src/aad/mod.rs
@@ -40,11 +40,14 @@ impl<'a> Aad<'a> {
         self.0.is_empty()
     }
 
-    /// Converts a borrowed `Aad` into an owned one, copying the bytes if necessary.
-    pub fn into_owned(self) -> Aad<'a> {
+    /// Converts a borrowed `Aad` into an owned one, copying the bytes if
+    /// necessary. The result borrows nothing, so it can outlive the input —
+    /// which is what lets the sequence and map sub-ciphers hold the AAD they
+    /// were constructed with for the lifetime of the builder chain.
+    pub fn into_owned(self) -> Aad<'static> {
         match self.0 {
-            x @ Cow::Borrowed(_) => Self(x.into_owned().into()),
-            Cow::Owned(_) => self,
+            Cow::Borrowed(slice) => Aad(Cow::Owned(slice.to_vec())),
+            Cow::Owned(owned) => Aad(Cow::Owned(owned)),
         }
     }
 
--- a/packages/aead/src/aad/mod.rs
+++ b/packages/aead/src/aad/mod.rs
@@ -51,6 +51,27 @@ impl<'a> Aad<'a> {
     pub(crate) fn pae(pieces: &[&[u8]]) -> Self {
         pae::encode(pieces)
     }
+
+    /// Derives the AAD a map entry's value must be sealed against, binding the
+    /// entry `key` to this (caller-supplied) AAD.
+    ///
+    /// Map keys travel in the clear inside a ciphertext container, so without
+    /// this binding an attacker holding a stored ciphertext could swap or
+    /// rename keys undetected, silently reassigning values to different
+    /// fields. [`MapCipher::encrypt_value`](crate::MapCipher::encrypt_value)
+    /// and [`MapAccess::next_entry`](crate::MapAccess::next_entry)
+    /// implementations are required to derive each entry's effective AAD
+    /// through this method — both sides must use it, or nothing decrypts.
+    ///
+    /// The encoding is `PAE(domain, aad, key)`. The leading domain-separation
+    /// label keeps the result disjoint from user-supplied composite AAD: a
+    /// caller binding the tuple `(aad, key)` at the top level (e.g. via
+    /// [`ContextTag`](crate::ContextTag)) encodes `PAE(aad, key)`, which can
+    /// never collide with a map entry's three-piece, labelled encoding.
+    pub fn for_map_entry(&self, key: &str) -> Aad<'static> {
+        const MAP_ENTRY_DOMAIN: &[u8] = b"vitaminc/aead/map-entry/v1";
+        pae::encode(&[MAP_ENTRY_DOMAIN, self.as_bytes(), key.as_bytes()])
+    }
 }
 
 /// Types that can be canonically converted into an [`Aad`].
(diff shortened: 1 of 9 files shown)


coderdan:
Review performed by me, Codex, running with GPT-5.6 Sol.

**P1: Authenticate keys for empty composite values**

`packages/encrypt/src/cipher.rs:218-219`

Map-key authentication remains bypassable for empty composite values because no AEAD operation authenticates their derived entry AAD.

When `U` is an empty `Vec` or `HashMap`, `encrypt_with_aad` emits an empty container without performing any AEAD seal, and decryption iterates zero children without verifying `entry_aad`. An attacker can therefore rename an entry such as `roles: []` to `is_admin: []` and decryption still succeeds, bypassing the new key-authentication guarantee.

Please add an authenticated marker or otherwise verify the derived AAD for values with no encrypted leaves.

coderdan:
## Code review findings (automated review of the stack at #254's tip; all findings trace to this PR's changes or interact with them)

The core map-key AAD binding verified sound — encrypt side (`Aad::for_map_entry` in `packages/encrypt/src/cipher.rs`) exactly mirrors decrypt, with pinned-encoding and adversarial tests. Workspace check and full test suites (incl. README doctests) pass. Findings below, ranked.

### Blockers

**1. All-passthrough composites bypass empty-marker authentication** — `packages/encrypt/src/cipher.rs` (`AesSeqCipher::end` / `AesMapCipher::end`)

`end()` seals the authenticated marker only when `self.items`/`self.entries` `is_empty()`, but passthrough items count as content while authenticating nothing. `encrypt_seq().passthrough_next(Box::new(v))?.end(aad)` returns `Sequence([Passthrough])` — non-empty, so `aad` is silently dropped. Decryption succeeds under ANY wrong AAD, and if the composite is a map value, its cleartext key can be renamed or swapped undetected — the exact key-renaming attack the `EmptySequence`/`EmptyMap` markers and `Aad::for_map_entry` were added to prevent, and a direct violation of the `SeqCipher::end` trait contract ("must authenticate aad even when no encrypted elements were appended").

*Fix direction: authenticate whenever zero encrypted items were appended, not zero total items (or reject all-passthrough composites outright — a composite with no sealed content has nothing authenticating it on the decrypt side either).*

**2. Some→None laundering via `decrypt_option`** — `packages/encrypt/src/cipher.rs` (`decrypt_option` `None` arm)

The `None` arm verifies the leaf's tag but discards the plaintext without checking it is empty (unlike the new `verify_empty_marker`), and the `None` marker is sealed against the bare AAD with no domain label. An attacker holding a stored ciphertext takes `Some("secret")` encrypted under AAD `x` — `Single(leaf)` — and re-tags the variant as `None(leaf)`: the tag verifies over plaintext `"secret"` with AAD `x`, the result is discarded, and `Ok(None)` is returned — silent authenticated data deletion. Same gap in `decrypt_any`'s `None` arm.

*Fix direction: check plaintext emptiness in the `None` arms and domain-separate the `None` marker AAD, same treatment as the new empty-composite markers.*

### Should-fix

**3. Empty-marker AAD uses unlabelled 2-piece PAE and lives in the wrong crate** — `packages/encrypt/src/cipher.rs` (`empty_sequence_aad` / `empty_map_aad`)

The helpers use `PAE(domain, aad)` — the exact collision shape `Aad::for_map_entry`'s own docs say requires a labelled 3-piece encoding. A caller using tuple AAD `(b"vitaminc/aead/empty-sequence/v1", x)` produces bytes identical to the empty-sequence marker AAD for `x`: an empty plaintext sealed under that AAD can be rewrapped as `EmptySequence` and decrypts as a valid empty `Vec` under AAD `x` (shape confusion). The constants are also private to `vitaminc-encrypt` with no pinned-encoding tests, while satisfying a trait contract defined in `vitaminc-aead` — every future `Cipher` implementor must byte-copy them or produce incompatible markers.

*Fix direction: `Aad::for_empty_sequence`/`for_empty_map` in `vitaminc-aead` using the labelled 3-piece PAE, with pinned-encoding tests (contrast `for_map_entry_pins_encoding`).*

**4. Sequences bind neither element count nor position** — `packages/encrypt/src/cipher.rs` (`decrypt_seq`)

Each element is sealed against the same bare AAD and `end()` ignores its `aad` argument when non-empty, so truncation, duplication, reordering, and cross-ciphertext splicing of sequence elements (or extracting a lone element as a `Single`) go undetected. Mostly pre-existing, but this PR's empty-marker machinery shows the fix shape: an always-sealed end marker binding `(aad, count)` would close truncation, the all-passthrough hole (finding 1), and the empty case with one mechanism.

**5. `StaticMapBuilder::encrypt_entry` seals against bare AAD — no key binding** — `packages/aead/src/hlist/cipher.rs`

The static/hlist counterpart of `MapCipher` in the same crate still seals values with no `Aad::for_map_entry` binding, so its cleartext `Entry.key` remains unauthenticated while the `MapCipher` trait now documents key binding as a hard contract. `cipher_static.rs`'s comments explicitly anticipate a future deserializer feeding untrusted input to this path; once static-map ciphertexts are serialized, the exact attack this PR closes on the dynamic path reopens there. Two map APIs in one workspace enforcing opposite security properties under the same documented contract.

**6. Passthrough map entries form an unauthenticated mutable channel** — `packages/encrypt/src/cipher.rs` (`passthrough_entry`)

Nothing prevents a passthrough entry duplicating/shadowing an encrypted entry's key at encrypt time, and an attacker can freely add, remove, or rewrite passthrough entries (key and value) in a stored ciphertext. Example: mixed map `{ "schema_version": passthrough(2), "data": encrypted }` — attacker rewrites `schema_version` to 1 to steer a legacy parsing path, or injects `("role", Passthrough(...))` beside the sealed `"role"` entry to shadow it for consumers that fetch passthrough fields by key. Documented as a design exception; an end marker over the key set (finding 4's mechanism) would close it.

### Minor

**7. `end(aad)` is consulted only when the composite turns out empty** — `packages/aead/src/cipher.rs`

An `Encrypt` impl that threads `aad` to `encrypt_next` but `()` to `end` compiles and passes every non-empty-data test, then fails to decrypt in production the first time the collection is empty (marker sealed under `PAE(domain, "")`, opaque `Unspecified`). Storing the AAD once at `encrypt_seq`/`encrypt_map` construction would remove the duplicated parameter, this shape-dependent mismatch class, and the per-element `aad.clone()` in the `Encrypt` impls.

**8. Wire-format break with no version tag** — map values now open against `for_map_entry`-derived AAD and `Sequence([])`/`Map([])` shapes are rejected, so ciphertexts from the previous release fail to decrypt with the same opaque `Unspecified` as tampering. Intentional (`feat!`, pre-1.0), but worth an explicit release-note callout and/or a container version marker before serde lands.

**9. `Encrypt for HashMap<String, T>` is a verbatim copy of the `HashMap<&'static str, T>` impl** — `packages/aead/src/encrypt/impls.rs`. Since `encrypt_key` accepts `K: Into<Cow<'static, str>>`, one generic impl (`K: Into<Cow<'static, str>> + Eq + Hash`) replaces both with no coherence conflict — next protocol change otherwise lands in one and not the other.

**10. Empty-composite decrypt arms store the domain-wrapped marker AAD in the access structs** — `packages/encrypt/src/cipher.rs`. Dead value today (the iterator is empty), but any future change that reads the stored AAD would derive `PAE(domain, PAE(domain, aad))` and diverge from the encrypt side. Store the raw caller AAD.

coderdan:
## Fixes for the review findings above

Two commits pushed to this branch; the rest of the stack (#254, #255, #256, #257) is rebased on top.

### `fix(aead)!: close marker and passthrough authentication bypasses`

Covers findings 1, 2, 3, 5 and 10 — all four bypasses shared one root cause: markers and composites that authenticated less than their shape claimed.

**Finding 1 — all-passthrough composites (blocker).** `SeqCipher::end`/`MapCipher::end` now track whether any item arrived through the *authenticated* path (`encrypt_next`/`encrypt_value`) rather than testing total item count. A composite whose items are all passthrough is rejected at `end()`, and `decrypt_seq`/`decrypt_map` refuse to open one arriving from storage. The trait docs now state the requirement for other implementors.

**Finding 2 — Some→None laundering (blocker).** The `None` marker is sealed against `Aad::for_none` and both decrypt arms (`decrypt_option` and `decrypt_any`) now go through `verify_empty_marker`, so the sealed plaintext must actually be empty *and* the tag must bind the domain-separated AAD. A `Single` leaf can no longer be re-tagged as an authenticated absence, and an empty `Single` can't either.

**Finding 3 — marker AAD collisions.** All marker AADs moved into `vitaminc-aead` as `Aad::for_empty_sequence` / `for_empty_map` / `for_none`, sharing a private `for_marker(kind)` that emits the labelled three-piece `PAE(domain, aad, kind)` — the same shape `for_map_entry` uses. The `kind` piece also keeps the three marker types disjoint from one another, so a stored marker can't be replayed as a different structural claim. Pinned-encoding tests added, matching `for_map_entry_pins_encoding`. The static path's absent marker (which had its own two-piece `(b"vitaminc:absent", aad)` label) now derives through the same `for_none`, so both absence markers share one wire commitment.

**Finding 5 — static map key binding.** `StaticMapBuilder::encrypt_entry` and `none_entry` now bind the entry key via `Aad::for_map_entry`, matching the dynamic `MapCipher` contract. New `Aes256Cipher::open_entry` / `verify_absent_entry` counterparts derive the same binding on the way out, and adversarial tests cover the renamed-key case for both.

**Finding 10 — empty-composite access structs.** They now stash the caller's raw AAD instead of the marker derivation, so a future visitor reading it can't silently double-wrap.

### `refactor(aead): merge duplicate HashMap Encrypt impls into one generic impl`

Finding 9. One impl generic over `K: Into<Cow<'static, str>> + Eq + Hash` replaces the two verbatim copies, so the next protocol change can't land in one and miss the other.

### Deliberately not fixed here

- **Finding 4 (sequence count/position binding)** and **finding 6 (passthrough entries as an unauthenticated channel)** — both want the same mechanism: an always-sealed end marker binding `(aad, count)` or the entry key set. That is a larger wire-format change than this PR should carry, and it wants its own design pass. Worth an issue.
- **Finding 7 (`end(aad)` only consulted when empty)** — the ergonomic fix is storing the AAD at `encrypt_seq`/`encrypt_map` construction, which changes the `Cipher` trait shape. Also better as a follow-up.
- **Finding 8 (wire-format break)** — intentional and pre-1.0; this commit widens it (marker AAD derivations changed), so it needs a release-note callout. `BREAKING:` trailers on both commits cover the changelog.

Also fixed in passing: an earlier rebase of this stack accidentally swept 472 `bindings/go/vcencrypt/guest/target/` build artifacts into #254's commit. Removed — that branch is back to 8 source files.

Full workspace check and test suites pass on every branch in the stack (the one failure, `vitaminc-kms tests::test_finalize`, needs LocalStack on :4566 and is unrelated). `cargo fmt --check` clean. The Go bindings' guest wasm was rebuilt against the fixed cipher and its tests pass.

coderdan:
## Finding 7 also fixed: composite AAD is now captured once

Pushed as `refactor(aead)!: capture composite AAD once at sub-cipher construction`, on top of the security fixes.

**The problem, reproduced.** `SeqCipher`/`MapCipher` took the AAD twice — per item and again at `end()` — and only the *empty* branch of `end()` read its copy. So an `Encrypt` impl that threaded the real AAD to the elements but something else to `end()` compiled and behaved perfectly until the first empty collection:

```rust
// one wrong argument, in an otherwise ordinary impl
.try_fold(cipher.encrypt_seq(Some(len)), |c, item| c.encrypt_next(item, aad.clone()))?
.end(())          // should be .end(aad)
```

```
Tags(vec!["a", "b"])  → encrypt → decrypt   ✅ Ok(["a", "b"])
Tags(vec![])          → encrypt → decrypt   ❌ Err(Unspecified)
```

Data-dependent, not code-dependent: it type-checks, every test with realistic non-empty fixtures passes, and it ships. It then surfaces as an opaque `Unspecified` indistinguishable from key corruption or tampering — and asymmetrically, since *encryption* succeeds, producing a ciphertext that can never be read back.

**The fix.** `Cipher::encrypt_seq`/`encrypt_map` now take the AAD, and the sub-ciphers hold it:

```rust
cipher.encrypt_seq(Some(len), aad)   // captured once
    .encrypt_next(item)?             // ...reused per element
    .end()                           // nothing left to mismatch
```

`encrypt_next`, `encrypt_value`, `encrypt_entry` and `end` take no AAD at all, so the mismatch is no longer expressible — the class of bug is gone rather than documented against. A regression test drives the sub-cipher directly (the shape a derive macro would generate) and asserts the empty case round-trips *and* stays AAD-sensitive.

**Bonus: one fewer allocation per element.** The old shape cloned the AAD per item. `AesSeqCipher`/`AesMapCipher` now own it once and lend it per element, so an owned AAD — which every nested composite has, since `for_map_entry` allocates — costs one allocation per container instead of one per element. `Aad::into_owned` returns `Aad<'static>` to make that storable; it already copied the bytes, the old signature just pinned the result to the input lifetime for no reason.

**BREAKING:** `Cipher`/`SeqCipher`/`MapCipher` implementations must move the AAD parameter from the per-item methods and `end()` to `encrypt_seq`/`encrypt_map`. Ciphertext bytes are unaffected.

---

Per your calls on the rest: **4** is already tracked in #252, **6** is intentional for the record-shaped use case (query results can't guarantee every field is present; opt-in tightening is a future option), and **8** is moot — nothing released. All three dropped.

Whole stack rebased and green (the single failing suite everywhere is `vitaminc-kms tests::test_finalize`, which needs LocalStack on :4566).

packages/aead/src/aad/mod.rs
```diff
@@ -40,17 +40,86 @@ impl<'a> Aad<'a> {
...
+    /// and [`MapAccess::next_entry`](crate::MapAccess::next_entry)
+    /// implementations are required to derive each entry's effective AAD
+    /// through this method — both sides must use it, or nothing decrypts.
+    ///
+    /// The encoding is `PAE(domain, aad, key)`. The leading domain-separation
+    /// label keeps the result disjoint from user-supplied composite AAD: a
+    /// caller binding the tuple `(aad, key)` at the top level (e.g. via
+    /// [`ContextTag`](crate::ContextTag)) encodes `PAE(aad, key)`, which can
+    /// never collide with a map entry's three-piece, labelled encoding.
+    pub fn for_map_entry(&self, key: &str) -> Aad<'static> {
```
freshtonic, line 74: This labelled derivation is the right primitive for the map contract: it binds the cleartext key without conflating it with caller-composed tuple AAD. Extending the same domain-separated pattern to empty/`None` markers is important because those shapes can themselves be map values.

freshtonic (approved):
## Standards

Pass. The trait contracts, domain separation, breaking-format implications, and negative security cases are documented and tested. The additional marker and static-map changes are cohesive with the core guarantee: map keys would not be meaningfully bound if empty, absent, passthrough-only, or static composite values could bypass equivalent authentication.

## Spec

Pass. Runtime keys and `HashMap<String, T>` support are present, and key/value pairing is authenticated while map reordering remains valid. Sequence position/length authentication remains explicitly out of scope.

Approved. The expanded hardening is justified by the integrity invariant rather than unrelated scope.

PR #254, opened 2026-07-24
feat(aead)!: shared generic CipherText container and typed passthrough currency

### LINKED ISSUE #270: Share one generic ciphertext container and give passthrough a declared type
The recursive ciphertext tree (single / sequence / map / none / passthrough) is currently spelled out per cipher as `AesCipherText`, and the suite has its own near-identical `ZeroKmsCipherText`. Anything that walks a ciphertext container — most immediately an FFI bridge mapping the tree onto a host language's value tree — has to be written once per cipher. The shape is the same in every case; only the sealed leaf type differs, and only the leaf carries a byte-format commitment.

The container should become generic over its leaf type and its passthrough payload, with the existing per-cipher types reduced to aliases so variant construction and matching are unaffected. Downstream consumers, including `stack-encrypt` (cipherstash-suite#2019), then write container-walking code once. The container deliberately gets no canonical byte form.

Separately, passthrough values are currently carried through `Cipher`/`SeqCipher`/`MapCipher`/`Decipher` as `T: Any + Send` generics, forcing type erasure and a downcast even for ciphers where the passthrough is always one known host type. A cipher should declare its passthrough currency once as an associated type: Rust-native ciphers keep a boxed `dyn Any + Send` (with a typed downcasting convenience on decrypt), while an FFI-targeting cipher can declare an owned host-value type and pass values in and out untouched.

This is API-breaking for third-party cipher implementations but changes no ciphertext bytes.

Addressed by #254 (stacked on #253).

coderdan:
## Summary

> ⚠️ **Stacked on** cipherstash/vitaminc#253 (`feat/aead-map-key-auth`) — the diff here is just the container extraction and passthrough refactor. Merge cipherstash/vitaminc#253 first, or retarget this to `main` once it lands.

Second core-trait step toward NAPI/FFI encryption support, and groundwork for `stack-encrypt` (cipherstash-suite#2019):

* **Shared generic container.** `vitaminc-aead` now provides `CipherText<Leaf, P>` — the recursive `Single`/`Sequence`/`Map`/`None`/`Passthrough` tree, generic over the sealed leaf type and the passthrough payload. `AesCipherText` becomes an alias for `CipherText<LocalCipherText, BoxedPassthrough>` (variant construction and matching are unaffected), and `ZeroKmsCipherText` in the suite can become `CipherText<DataKeyCipherText, _>` the same way. Downstream consumers — most immediately the NAPI container↔JS-object walk — get written once against the generic container instead of per cipher. Only `Leaf` carries a byte-format commitment; the container has no canonical byte form by design.
* **Typed passthrough currency.** The `Cipher`/`SeqCipher`/`MapCipher`/`Decipher` traits gain a `Passthrough` associated type, replacing the `T: Any + Send` generics on `passthrough`, `passthrough_next`, `passthrough_entry`, and `decrypt_passthrough`. A cipher declares its currency once: Rust-native ciphers use `Box<dyn Any + Send>` (callers box in; downcast out via the new typed convenience `AesDecipher::decrypt_passthrough_as::<T>()`), while FFI-targeting ciphers can declare an owned host-value type and carry passthrough values value-in/value-out with no type erasure — the NAPI case where a passthrough is "always a Value" and the cipher never inspects it.

## Breaking changes

* **API only, no ciphertext-format change.** Third-party `Cipher`/`SeqCipher`/`MapCipher` implementations add the `Passthrough` associated type and drop the generic parameters from the passthrough methods; `Decipher::decrypt_passthrough` now returns the currency directly rather than downcasting to a caller-chosen `T`.

## Testing

* Existing passthrough round-trip, type-mismatch, and state-machine tests all updated to the boxed currency and passing unchanged in behaviour.
* Full workspace suite, clippy, and `cargo fmt --check` clean (kms integration test still requires LocalStack; unrelated).

Closes cipherstash/vitaminc#270

### DIFF
--- a/packages/aead/src/container.rs
+++ b/packages/aead/src/container.rs
@@ -1,0 +1,56 @@
+/// The recursive ciphertext container shared by tree-shaped [`Cipher`]
+/// implementations.
+///
+/// The shape mirrors the structure of the plaintext that was encrypted: a
+/// single value yields [`Single`](CipherText::Single), while non-empty `Vec`s
+/// and `HashMap`s yield [`Sequence`](CipherText::Sequence) and
+/// [`Map`](CipherText::Map). Empty composites use the authenticated marker
+/// variants [`EmptySequence`](CipherText::EmptySequence) and
+/// [`EmptyMap`](CipherText::EmptyMap), since they contain no element
+/// ciphertexts to authenticate the supplied AAD. Nested structures are
+/// represented recursively.
+///
+/// Concrete ciphers instantiate the two parameters:
+///
+/// - `Leaf` is the sealed leaf ciphertext (nonce + ciphertext + tag) — e.g.
+///   [`LocalCipherText`](crate::LocalCipherText) for a local AES cipher, or an
+///   envelope leaf that additionally carries a wrapped data key.
+/// - `P` is the passthrough payload type, fixed by the cipher's
+///   [`Cipher::Passthrough`](crate::Cipher::Passthrough) associated type.
+///   Rust-native ciphers typically use `Box<dyn Any + Send>`; FFI-targeting
+///   ciphers use an owned host-value type carried value-in/value-out.
+///
+/// Only the `Leaf` type carries a byte-level format commitment (and should
+/// implement `serde` where ciphertexts are persisted). The container itself
+/// has no canonical byte representation: hosts project it onto their native
+/// structures (e.g. a JS object tree across an FFI boundary), with leaves as
+/// opaque byte strings.
+///
+/// [`Cipher`]: crate::Cipher
+#[derive(Debug)]
+pub enum CipherText<Leaf, P> {
+    /// A single sealed value.
+    Single(Leaf),
+    /// A sequence of ciphertexts produced from a `Vec`-shaped plaintext.
+    Sequence(Vec<CipherText<Leaf, P>>),
+    /// An empty sequence, authenticated under domain-separated AAD via the
+    /// marker leaf produced at [`SeqCipher::end`](crate::SeqCipher::end).
+    EmptySequence(Leaf),
+    /// A map of (cleartext key, ciphertext value) pairs produced from a
+    /// map-shaped plaintext. Keys are stored in the clear but are bound into
+    /// each value's AAD via [`Aad::for_map_entry`](crate::Aad::for_map_entry),
+    /// so they cannot be swapped or renamed undetected (passthrough entries
+    /// excepted).
+    Map(Vec<(String, CipherText<Leaf, P>)>),
+    /// An empty map, authenticated under domain-separated AAD via the marker
+    /// leaf produced at [`MapCipher::end`](crate::MapCipher::end).
+    EmptyMap(Leaf),
+    /// The authenticated absent marker produced by
+    /// [`Cipher::encrypt_none`](crate::Cipher::encrypt_none). Stores a sealed
+    /// empty plaintext whose tag binds the supplied AAD.
+    None(Leaf),
+    /// A value passed through the container **unencrypted and
+    /// unauthenticated** via [`Cipher::passthrough`](crate::Cipher::passthrough).
+    /// Non-sensitive data only — see the trait docs.
+    Passthrough(P),
+}
--- a/packages/aead/src/decipher/mod.rs
+++ b/packages/aead/src/decipher/mod.rs
@@ -1,7 +1,5 @@
 pub mod impls;
 
-use std::any::Any;
-
 use vitaminc_protected::Protected;
 
 use crate::{Aad, IntoAad, Unspecified};
@@ -30,6 +28,11 @@ pub trait Decipher<'c>: Sized {
     type Ok<T>
     where
         T: Send + 'c;
+
+    /// The payload type returned by
+    /// [`decrypt_passthrough`](Decipher::decrypt_passthrough) — matches the
+    /// encrypt-side [`Cipher::Passthrough`](crate::Cipher::Passthrough).
+    type Passthrough: Send + 'c;
 
     /// Transform the inner value of an [`Ok`](Decipher::Ok) container.
     ///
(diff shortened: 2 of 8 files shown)


packages/aead/src/decipher/mod.rs
```diff
@@ -78,14 +81,17 @@ pub trait Decipher<'c>: Sized {
         A: IntoAad<'a>;
 
     /// Recover a value stored via [`Cipher::passthrough`](crate::Cipher::passthrough).
-    /// Returns an error if the ciphertext is not a passthrough or the stored
-    /// type does not match `T`.
+    /// Returns an error if the ciphertext is not a passthrough.
+    ///
+    /// The payload comes back as the decipher's
+    /// [`Passthrough`](Decipher::Passthrough) currency, exactly as stored —
+    /// value-in/value-out. For the `Box<dyn Any + Send>` currency used by
```
coderdan: Don't refer to the associated type as "currency". Its confusing.

coderdan:
Automated code review ran against this branch's tip (stack diff vs `main`). This PR's own change — the shared generic `CipherText<Leaf, P>` container and `Passthrough` associated type — **verified sound**: all callers and implementors updated, workspace check and full test suites green (including README doctests), and the old `AesCipherText` enum had only a `Debug` derive, so nothing was lost in the aliasing.

All 10 findings trace to the base PR's changes (map-key authentication + empty-composite markers) or pre-existing design gaps they interact with — see the full write-up on #253: https://github.com/cipherstash/vitaminc/pull/253#issuecomment-5102088151. Fixes will land there and this branch will be rebased.

coderdan:
Rebased onto #253's fixes (https://github.com/cipherstash/vitaminc/pull/253#issuecomment-5102281938). Two changes land in this branch:

- The passthrough-rejection tests moved with the container refactor, so they now box their payloads (`passthrough_next(Box::new(1u32))`) to match this PR's `Passthrough` associated type.
- An earlier rebase of this stack accidentally swept 472 `bindings/go/vcencrypt/guest/target/` build artifacts into this branch's commit. Removed — the commit is back to its 8 intended source files. (The guest crate's own `.gitignore` covers `target/`; the files only got in because `bindings/` doesn't exist yet at this point in the stack, so they were untracked here.)

packages/aead/src/decipher/mod.rs
coderdan: Fixed — "currency" is gone from the Rust docs across the whole stack.

The `Passthrough` associated type is now described plainly as the passthrough **payload type**. The concrete phrasings:

- `Cipher::Passthrough` — "The payload type carried by `passthrough` values" (the metaphor was redundant next to the plain-English gloss that was already there)
- `Decipher::decrypt_passthrough` — "comes back as the decipher's `Passthrough` type… Where that type is `Box<dyn Any + Send>`, as it is for Rust-native ciphers, callers downcast themselves"
- `CipherText::map_passthrough` — "the bridge between ciphers with different passthrough payload types"
- `BoxedPassthrough`, the README implementor skeleton, and (further up the stack) `passthrough_boxed`, `aead-value`'s transport codec, and the WASI guest comments all follow the same wording.

Applied on this branch and propagated through #255, #256 and #257, so the term doesn't reappear higher in the stack.

packages/aead/src/container.rs
```diff
@@ -0,0 +1,56 @@
...
+///
+/// Only the `Leaf` type carries a byte-level format commitment (and should
+/// implement `serde` where ciphertexts are persisted). The container itself
+/// has no canonical byte representation: hosts project it onto their native
+/// structures (e.g. a JS object tree across an FFI boundary), with leaves as
+/// opaque byte strings.
+///
+/// [`Cipher`]: crate::Cipher
+#[derive(Debug)]
+pub enum CipherText<Leaf, P> {
```
freshtonic, line 31: Good seam: the container owns only recursive shape, while `Leaf` owns persisted ciphertext bytes and `P` owns host passthrough representation. That separation is what lets downstream walkers be shared without accidentally freezing a second wire format.

freshtonic (approved):
## Standards

Pass. Extracting the recursive container removes per-cipher duplication without introducing a byte-format abstraction the project does not own. The passthrough associated types are tied together by equality bounds, keeping the public trait family coherent.

## Spec

Pass. `CipherText<Leaf, P>`, the AES alias, declared passthrough payload types, native boxed downcasting convenience, and the deliberate absence of a canonical container encoding all match #270.

Approved. CI is green and I found no blocking correctness, security, or API-design issue.

packages/aead/src/container.rs
```diff
@@ -0,0 +1,56 @@
...
+    EmptySequence(Leaf),
+    /// A map of (cleartext key, ciphertext value) pairs produced from a
+    /// map-shaped plaintext. Keys are stored in the clear but are bound into
+    /// each value's AAD via [`Aad::for_map_entry`](crate::Aad::for_map_entry),
+    /// so they cannot be swapped or renamed undetected (passthrough entries
+    /// excepted).
+    Map(Vec<(String, CipherText<Leaf, P>)>),
+    /// An empty map, authenticated under domain-separated AAD via the marker
+    /// leaf produced at [`MapCipher::end`](crate::MapCipher::end).
+    EmptyMap(Leaf),
```
freshtonic, line 47: `Sequence(vec![])` and `EmptySequence` - what's the difference?

Looks like there are two type-valid ways to construct an empty sequence. Same problem with `Map` & `EmptyMap`.

If you _really_ need different variants then may I introduce you to the [`vec1`](https://crates.io/crates/vec1) crate?

coderdan: Good question — the distinction is cryptographic, not representational.

Every element of a `Sequence` carries its own AEAD tag bound to the caller's AAD. Strip all the elements and there is zero cryptographic material left: an empty composite authenticates *nothing*. Before the marker variants, an empty composite decrypted vacuously under **any** AAD — anyone could truncate a stored sequence to "empty", or fabricate one, without holding the key. `EmptySequence(Leaf)` closes that: when `SeqCipher::end` runs with zero items it seals a real marker leaf under `aad.for_empty_sequence()`, so "this value was empty" becomes a claim only the key holder can make, bound to the same AAD context as any other value (see `fbdd79c` on this branch).

Two variants rather than one shared `Empty(Leaf)` because the markers are domain-separated: `for_empty_sequence` / `for_empty_map` / `for_none` are distinct PAE derivations, so an empty-map marker replayed where an empty sequence is expected fails the tag check.

You're right that `Sequence(vec![])` is type-constructible, but it fails closed at both ends: the encrypt side never emits it (empty goes to the marker), and the decrypt side rejects it — the guard requires at least one authenticating (non-passthrough) element, which an empty vec can't satisfy.

`vec1` is a fair suggestion for making the invalid state unrepresentable rather than merely rejected — it would complement the markers (the sealed marker leaf is still required either way) and doesn't touch wire bytes. Happy to take that as a follow-up; the cost is a dependency plus churn at the construction/match sites, including the FFI bridge code that builds containers.

freshtonic: In that case I suggest `vec1` as a follow-up. Algebraic data types are much more useful when they are correct by construction. Things like `Arbitrary` impls with no post-filtering for invalid values, proofs probably get easier if we ever go down that road.

freshtonic: It's also worth stating that having "empty" represented two ways will trigger some devs (like me!) and won't be able to drop it until they've done some code archeology to find answers.

PR #255, opened 2026-07-24
feat(aead-napi): NAPI value bridge crate for Node.js FFI encryption

### LINKED ISSUE #271: Encrypt dynamically typed JavaScript values from Node.js
Node.js applications need to encrypt values whose shape is only known at runtime — arbitrary JS objects, arrays, and scalars — and get the plaintext back with its original JS type intact. Nothing in the stack supports this today: encryption is driven by Rust types naming their plaintext up front, and there is no bridge between a V8 value and the structured cipher traits.

We want a value-bridge crate that models a JS value as an owned, `Send`, self-describing tree (null/undefined/bool/number/string/bytes/array/object), with conversions confined to the JS thread so the crypto itself can run anywhere and is async-ready by construction. Leaves should seal as a tagged payload *inside* the AEAD envelope so types round-trip exactly (including NaN payloads and negative zero) and type confusion requires forging the auth tag. Arrays and objects should ride the cipher's existing sequence and map modes, with object keys AAD-bound by the map key authentication work.

Two core capabilities are missing to make this work. Decryption needs a `deserialize_any`-style entry point: a dynamically typed value cannot name its plaintext type, so the decipher must inspect the ciphertext shape and dispatch to the matching visitor, including a visitor method for authenticated absent markers whose AAD binding is verified before the visitor runs so absence cannot be forged. The container also needs a fallible way to re-home its passthrough payload between ciphers with different passthrough currencies.

Because object keys travel in cleartext and are attacker-writable in stored ciphertexts, the boundary must reject prototype-pollution keys in both directions, bound recursion depth, reject sparse arrays and non-byte typed arrays explicitly, and UTF-8-validate decrypted strings before they reach JS.

Addressed by #255 (stacked on #254, which stacks on #253).

coderdan:
## Summary

> ⚠️ **Stacked on #254** (which stacks on #253). Diff here = `decrypt_any`/`visit_none`/`map_passthrough` core additions + the new crate. Retarget to `main` as the base PRs land.

The third step of the NAPI plan: the bridge crate itself, plus the one core capability it needed.

### Core additions (`vitaminc-aead` / `vitaminc-encrypt`)

- **`Decipher::decrypt_any`** — the analog of serde's `deserialize_any`. A dynamically typed value can't name its plaintext type up front, so the decipher inspects the ciphertext shape and dispatches to the matching visitor method. The new **`DecipherVisitor::visit_none`** (default: error) receives authenticated absent markers, with the AAD binding verified *before* the visitor is invoked, so absence cannot be forged. Passthrough ciphertexts are deliberately not dispatched (typed `decrypt_passthrough` only).
- **`CipherText::map_passthrough`** — fallibly re-homes a tree's passthrough payload type between ciphers with different passthrough currencies (`Box<dyn Any + Send>` ↔ owned FFI value).

### New crate: `vitaminc-aead-napi`

- **`NapiValue`** — owned, `Send`, self-describing JS value tree (null/undefined/bool/number/string/bytes/array/object). Conversions run on the JS thread; crypto can run anywhere (async-ready by construction). Secret leaves in `Protected`; V8-side originals documented as unwipeable.
- **Tagged-leaf `Encrypt`/`Decrypt`** — leaves seal as `[tag] ++ payload` *inside* the AEAD envelope: JS types round-trip and type confusion requires forging the auth tag. Numbers round-trip by IEEE-754 bit pattern (`NaN` payloads, `-0` survive). Arrays/objects use the cipher's sequence/map modes; object keys are AAD-bound per #253. Rust `Option::None` ciphertexts surface as JS `null`.
- **`JsCipherText<Leaf, P>`** — projects the generic container (#254) onto plain JS `{t: "ct"|"none"|"seq"|"map"|"pt", v}` nodes with `Buffer` leaves, and back. No container byte-format is introduced; only leaves carry one.
- **Boundary hardening** — `__proto__`/`constructor`/`prototype` rejected in both directions (keys travel cleartext and are attacker-writable in stored ciphertexts); recursion depth bounded (128); sparse arrays, non-byte typed arrays, and `Date` (for now) rejected explicitly; decrypted string leaves UTF-8-validated before reaching JS.

### Packaging

- Workspace member + release-plz version group entry. Lib crate only — the cdylib/npm addon comes separately and stays out of the release group.
- Rust unit tests link via napi's `noop` feature (no Node host needed): 12 tests covering leaf/nested/empty round-trips, number edge cases, AAD binding, tag tampering (unknown tag, truncated number, invalid UTF-8, string≠bytes), object key-swap detection end-to-end, and forbidden-key classification. The `Env`-dependent conversion paths compile against napi v3 but are exercised by the addon's Node tests later.

## Breaking changes

- `Decipher` implementations must add `decrypt_any` (API-only; no ciphertext-format change).

Closes #271


coderdan:
Rebased onto the security fixes in #253 (https://github.com/cipherstash/vitaminc/pull/253#issuecomment-5102281938). One change lands in this branch: `decrypt_any`'s `None` arm — introduced by this PR's self-describing decryption commit — verified the marker tag against the bare AAD without checking the sealed plaintext was empty, the same Some→None laundering hole fixed in `decrypt_option`. It now goes through `verify_empty_marker` with `Aad::for_none`, matching the `decrypt_option` path.

coderdan:
Rebased onto the AAD-once trait change in #253 (https://github.com/cipherstash/vitaminc/pull/253#issuecomment-5103318919). `NapiValue`'s `Encrypt` impl moves its AAD to `encrypt_seq`/`encrypt_map` and drops the per-element `aad.clone()`; behaviour and ciphertext bytes unchanged.

coderdan:
## CRAP and mutants gates fixed

Both failed for one root cause — six functions at zero coverage — but they split into two genuinely different kinds, so the fix differs.

**Two were ordinary Rust that just lacked tests. They now have them.**

`CipherText::map_passthrough` (CRAP 132, 0% → 11.0, 100%): tests cover every container variant, payloads nested at three depths, the visit-once property, the never-invoked case for a tree with no passthrough, and error propagation from the top level, from inside a sequence, and from inside a map — the property the fallible signature exists for.

`NapiValue`'s `Debug` impl (CRAP 90, 0% → 9.0, 100%): it scored zero because a *passing* suite never formats a value — the impl only runs when an assertion fails. The test now exercises it directly and pins the property worth having: secret leaves render as `<redacted>`, so a downstream assertion failure can't spill plaintext into CI logs.

**Four cannot be reached by `cargo test` at all.** `js_to_value`, `value_to_js`, `node_to_js` and `node_from_js` each take a live `napi_env`/`Unknown`/`Object`, which exists only inside a running V8 isolate. The `napi/noop` dev-dependency stubs the symbols so the crate *links*; it does not make the calls work. Every mutant there survived because the function is unreachable, not because a test was missing.

They're exempted in `.cargo-crap.toml` and `.cargo/mutants.toml` **by function name rather than by file**, so the testable parts of those modules stay gated — `forbidden_key` and friends are still scored, and 31 other `aead-napi` mutants still run. I verified all 15 mutants CI reported are gone and that the two fixed functions reach 100% through tests rather than through the allow-list.

The exemption is recorded as a deferral in the crate README, not a dismissal: that layer needs JS-level tests driving a built addon, alongside the first `#[napi]` entry points that consume this crate. Until then those two modules are unguarded by CI and want review by hand.

All three checks pass.

coderdan:
Review follow-up

The first P1 (authenticating dynamically inferred sequence shape) is intentional for this framework. The primary use case is encrypting rows/records for a database, where insertion order generally does not match retrieval order. Encoding sequence ordering into AAD would therefore break retrieval. There is a tracked follow-up to support explicit sequence commitment in contexts where that behavior is appropriate, so this is not considered an actionable defect in this PR.

The other P1 is valid:

- A huge sparse array such as `new Array(0xffff_ffff)` reaches `Vec::with_capacity(len as usize)` before any element is inspected, allowing a very cheap input to request capacity for billions of values. Array lengths should be capped before allocation, and element presence should be checked explicitly. The same unchecked reservation exists in `node_from_js`.

The P2 conversion findings are also valid. Concrete examples:

1. Explicit `undefined` object properties fail instead of round-tripping:

   ```js
   const input = { x: undefined };
   Object.hasOwn(input, "x"); // true
   await roundTrip(input);    // throws "property vanished during read"
   ```

   napi-rs `Object::get` maps an `undefined` result to `None`, so the converter cannot distinguish this value from a missing property. Top-level `undefined` and `[undefined]` are unaffected.

2. Unsupported object classes are silently flattened:

   ```js
   await roundTrip(new Map([["role", "admin"]])); // {}
   await roundTrip(new Set(["admin"]));            // {}
   await roundTrip(/secret/gi);                     // {}
   await roundTrip(new ArrayBuffer(8));             // {}
   ```

   These values fall through to the plain-object branch. A value with custom enumerable fields is only partially preserved; for example, a `RegExp` with `pattern.note = "important"` returns as `{ note: "important" }`, losing its pattern, flags, and class identity. Non-plain objects should be rejected explicitly.

3. Inherited enumerable properties are copied as own properties:

   ```js
   const prototype = { role: "admin" };
   const input = Object.assign(Object.create(prototype), { name: "Alice" });

   Object.keys(input); // ["name"]
   const output = await roundTrip(input);
   Object.keys(output);           // ["name", "role"]
   Object.hasOwn(output, "role"); // true
   ```

   napi-rs `Object::keys` uses `napi_get_property_names`, which includes enumerable prototype properties. Inherited enumerable getters are consequently invoked when the values are read as well. Own-only enumeration or an explicit own-property filter is needed both in `js_to_value` and for ciphertext maps in `node_from_js`.

coderdan:
## Code review findings

Review of the branch diff. Two findings from the first pass were withdrawn after discussion and are noted at the bottom rather than listed as defects.

### Integrity

**1. Duplicate map keys are accepted on decrypt, enabling single-field rollback** — `packages/encrypt/src/cipher.rs:619`

Given two ciphertexts of one logical record under the same key + AAD (v1 `balance=100`, v2 `balance=5`), both `balance` values are sealed under the identical `for_map_entry(aad, "balance")`. Appending v1's `("balance", ct)` entry to v2's entries yields a map where both duplicates verify. `HashMapVisitor` is last-wins (`packages/aead/src/decipher/impls.rs:124`), as is the JS path's `obj.set` per entry, so an attacker keeps every other current field while rolling one back. Confirmed with a PoC recovering `balance=100`.

This is *not* covered by the column-projection argument that retires finding A below: a projection returns a subset of **distinct** keys, so no legitimate flow yields the same key twice in one container. Tracking seen keys in `AesMapAccess::next_entry` and erroring on a repeat closes it without constraining membership at all. Not exploitable when callers bind a record id/version into the AAD.

**2. `decrypt_any` dispatches on unauthenticated container shape** — `packages/encrypt/src/cipher.rs:507`

A stored `{t:"ct", v:leaf}` (a `Single` sealed under AAD `a`) can be rewrapped by an attacker as `{t:"seq", v:[{t:"ct", v:leaf}]}`. `decrypt_any` matches on the in-memory variant, dispatches to `decrypt_seq`, and `AesSeqAccess` verifies the element under the same bare AAD that `decrypt_bytes` (line 420) uses for a top-level `Single` — byte-identical, so it succeeds and the app receives `["x"]` where it stored `"x"`. Likewise a legitimate `EmptySequence` marker nested one level decrypts as `[[]]`, since the inner marker re-verifies under the same `for_empty_sequence(aad)`.

The old caller-typed `Decrypt` API foreclosed this (tests at line 1565 pin variant rejection); the self-describing path removes the type pin without adding a shape commitment, and `packages/aead-napi/src/ciphertext.rs` rebuilds that variant tree from an attacker-writable JS object. Also not a projection case — a `Single` rewrapped as a one-element `Sequence` isn't a subset of anything.

### NAPI bridge

**3. `Object::keys` walks the prototype chain into the sealed plaintext** — `packages/aead-napi/src/convert.rs:82`

napi 3.11.0's `Object::keys` calls `napi_get_property_names` (`object.rs:897`), which per the Node-API spec equals `napi_get_all_property_names` with `napi_key_include_prototypes | napi_key_enumerable` — not own-only. If anything in the process sets `Object.prototype.role = "admin"`, or the caller passes a class instance with enumerable prototype members, `encrypt({name:"x"})` seals `{name:"x", role:"admin"}` and decryption faithfully returns the injected key. The `forbidden_key` guard blocks only `__proto__`/`constructor`/`prototype` and cannot block arbitrary inherited names. Same issue at `packages/aead-napi/src/ciphertext.rs:177`, where a hostile node's inherited keys become ciphertext map entries.

**4. JS `Map`, `Set`, `DataView`, `RegExp`, `WeakMap` silently encrypt as `{}` — data loss, no error** — `packages/aead-napi/src/convert.rs:75`

`js_to_value` checks only `is_buffer`/`is_typedarray`/`is_array`/`is_date`; `napi_is_typedarray` is false for `DataView` and there is no `is_map`/`is_set` check. Encrypting `{pii: new Map([["ssn","123-45-6789"]])}` reaches the plain-object branch, where `Object::keys` returns nothing because the contents live in internal slots, producing `NapiValue::Object(vec![])`. That encrypts successfully to an `EmptyMap` marker (`value.rs:133`, `cipher.rs:285`) and decrypts to `{}` — the payload was never encrypted and is gone, silently at both ends. `Date`, two lines above, gets an explicit `InvalidArg` rejection; these deserve the same.

**5. The sparse-array rejection is dead code** — `packages/aead-napi/src/convert.rs:69`

napi 3.11.0's `Array::get<T>` (`array.rs:31-46`) returns `None` only when `index >= len`; for any in-bounds index it calls `napi_get_element` (which yields `undefined` for a hole) and unconditionally wraps in `Some` — unlike `Object::get_inner`, which does map `undefined` to `None`. `Unknown::from_napi_value` accepts `undefined`, so `arr.get(i)?` is always `Some` and the `ok_or_else` error can never fire. `encrypt([, , "x"])` silently returns `[undefined, undefined, "x"]` instead of the promised rejection, and `encrypt(new Array(1e8))` performs ~1e8 AEAD seals of `UNDEFINED` leaves. Both the crate README and the `feat(aead-napi)` commit message claim sparse arrays are rejected, so a documented guarantee is currently false.

**6. Attacker-controlled `Vec::with_capacity` can abort the process** — `packages/aead-napi/src/convert.rs:67`

`Array::from_napi_value` stores only `napi_get_array_length`'s `u32` (`array.rs:130-141`) and `get()` is lazy, so `new Array(2**32-1)` yields `len == 4294967295` with zero elements materialized. `js_to_value` then requests ~137 GB for `>=32`-byte `NapiValue` elements before validating index 0; on macOS/Windows, or Linux under strict overcommit, the allocation fails and `handle_alloc_error` aborts the whole Node process — not a catchable JS throw. `MAX_DEPTH=128` bounds nesting only, never breadth. Identical pattern on the decrypt path at `packages/aead-napi/src/ciphertext.rs:168`. `try_reserve`, or a clamped initial capacity that grows as elements validate, fixes both.

**7. A passthrough projecting to `undefined` can never be read back** — `packages/aead-napi/src/ciphertext.rs:192`

`node_to_js` writes `V_KEY` via `P::to_napi_value`; `NapiValue::Undefined` projects through `<()>::to_napi_value` to `napi_get_undefined` (`convert.rs:115`), and `P = ()` does the same. On the way back, `Object::get` delegates to `get_inner` (`object.rs:840-868`), which returns `Ok(None)` when the property's value type is `Undefined` — so `obj.get(V_KEY)?.ok_or_else(shape_error)?` rejects the crate's own faithful projection with `InvalidArg "malformed ciphertext value"`. A persisted `{t:"pt", v:undefined}` node can never be handed back for decryption, and `JSON.stringify` drops the key entirely with the same result. (`P = Option<T>` at `None` is unaffected — napi emits `null` there, which `get_inner` passes through.)

### Needs a decision, not a fix

**8. No version discriminator across the format break** — `packages/encrypt/src/cipher.rs:257`

The AAD derivation and container-shape changes invalidate every ciphertext written by current `main`: map values sealed under the bare map AAD, `None` markers under the bare AAD (dynamic) or `PAE("vitaminc:absent", aad)` (static, deleted in `cipher_static.rs`), and empty collections stored as unauthenticated `Sequence([])`/`Map([])`. `LocalCipherText` carries no version byte, so an old-format record is indistinguishable from a forgery and recovery requires the previous binary. The breaking commits are marked `feat!`/`fix!` at `0.2.0-pre.1`, so this may be entirely fine — it just needs an explicit confirmation that no persisted data exists anywhere, since there is no migration path.

### Documentation

- Map entry membership is unauthenticated (see A below). Worth stating as a caller obligation in the `decrypt_map` docs — "decrypt does not authenticate membership; verify the key set you projected for" — because the failure mode is asymmetric: deleting *all* entries **is** caught (`Map([])` fails the `.any(|v| !Passthrough)` guard at `cipher.rs:474` and returns `Err`), while deleting *some* returns `Ok`. Nobody should infer from the empty-map rejection that partial deletion is covered.
- `NapiValue::Array`'s comment says element positions are "carried structurally," which reads as an integrity claim. Reword so the constraint is stated explicitly (see A below).
- `CLAUDE.md:16` and `RELEASING.md:15` still say "all 12 crates"; `release-plz.toml` now lists 13 after `vitaminc-aead-napi`.
- The leaf tag bytes and the `{t,v}` node schema are both documented as persisted wire commitments but live inside the Node-specific crate, so planned Python/Go bindings will each have to copy them with no shared source of truth.

### Withdrawn after discussion

**A. Sequence elements have no positional binding; map membership is uncommitted.** Both were raised as tamper findings and both are retired by the same design constraint: this framework encrypts database records, insertion order does not match retrieval order, and column projection (`SELECT foo, bar`) legitimately returns a subset of keys. Binding an index into the AAD would break retrieval, and a key-set commitment sealed at encrypt time would break projection. Explicit sequence commitment for cases where it *is* appropriate is tracked separately. The residual caller obligations are captured under Documentation above.

### Verified clean

The branch's own hardening work checks out: `for_map_entry`/`for_none`/`for_empty_*` derivations are symmetric on both sides, all-passthrough composites are rejected symmetrically, the `None` marker check is strictly stronger than the bare-AAD check it replaced, and the hlist static path composes the same bindings. `cargo fmt --check` and `cargo clippy --workspace --all-targets` are clean. No `CHANGELOG` or version edits were made on the branch, and commit messages follow the conventional-commit convention with crate scopes.

packages/aead-napi/src/convert.rs
```diff
@@ -0,0 +1,337 @@
...
+/// rebuilding one). Bounds recursion so a hostile deeply-nested input cannot
+/// overflow the stack.
+pub(crate) const MAX_DEPTH: usize = 128;
+
+/// Property names that must never be read from or written onto a plain JS
+/// object: assigning them via `napi_set_property` triggers prototype-chain
+/// semantics (prototype pollution), and accepting them on input would let a
+/// value round-trip into that hazard.
+pub(crate) fn forbidden_key(key: &str) -> bool {
+    matches!(key, "__proto__" | "constructor" | "prototype")
```
freshtonic, line 25: Rejecting all three prototype-sensitive names in both directions is the correct closed boundary here. Combined with accepting only plain objects and enumerating own properties, this avoids treating attacker-controlled ciphertext keys as ordinary object assignment.

freshtonic (approved):
## Standards

Pass. The boundary code documents V8 ownership limitations, constrains recursion/allocation, and keeps JS-specific conversion separate from crypto traversal. No actionable standards or maintainability finding.

## Spec

Pass. Self-describing decryption, authenticated tagged leaves, fallible passthrough re-homing, JS-thread conversion, and the required prototype/sparse-array/typed-array/UTF-8 hardening are implemented.

Approved. CI is green and the security-sensitive boundary behavior is both explicit and covered.

tobyhede (commented):
## Review

Good crate — AAD derivations, the `Single`→`Sequence` re-homing defence, and the read-path prototype hardening are all well done. Two real bugs on the conversion path, plus a breaking-changes section that's wrong. Rest is docs and nits.

Verified against vendored `napi 3.11.0`, Node source, and a probe addon on Node 22/24/25.

### Blocking

**`ensure_plain_object` accepts every `Proxy` (`convert.rs:134-152`).** `napi_get_prototype` returns raw `null` for a Proxy and never fires the trap — Node uses V8's legacy `GetPrototype()` (`PrototypeIterator`), not `GetPrototypeV2()`. So we take the null-prototype early return. `is_array`/`is_buffer`/`is_typedarray`/`is_date` are also all false through a Proxy, so everything lands in the object branch:

    new Proxy(new Map([["a",1]]), {})  ->  encrypts as {}
    new Proxy([1,2,3], {})             ->  Object [["0",1],["1",2],["2",3]]
    new Proxy(Buffer.from("hi"), {})   ->  Object [["0",104],["1",105]]

Bare `new Proxy(x, {})`, zero traps, any target — exactly the silent payload destruction this function exists to prevent, and it's what Vue `reactive()`, MobX and ORM wrappers hand you. Fix: cross-check a `null` result against `Reflect.getPrototypeOf`.

**Non-byte typed arrays and `DataView` silently accepted (`convert.rs:170-178`).** `napi_is_buffer` is `IsArrayBufferView()`, not `IsUint8Array()`. So line 171 swallows every typed array and `DataView` — `Float64Array`, `Int32Array`, `BigInt64Array` get reinterpreted as raw LE bytes and decrypt back as a `Buffer`, no error — and the `is_typedarray()` branch at 174 is unreachable. Contradicts the comment at 175-176, README:53-55, and the PR body. Fix: test `is_typedarray()`/`is_dataview()` first, gate on `typed_array_type`.

### Fix before merge (no code change)

**Breaking-changes section is wrong.** It says "API-only; no ciphertext-format change", but three format-visible changes land here: the `WIRE_VERSION` prefix with no legacy read path, `for_leaf` in the AAD at five sites, and `for_sequence_element` on elements. All absent from base and `main`; they arrive in `53d477d` and `1ebf542` (also `!`-marked, also unmentioned). The changes are right — just list them plus a re-encrypt note, since this feeds git-cliff into a joint release. Real-world impact is small: pre-release only, sibling crates the only dependents.

**`decrypt_seq`'s caller note understates the gap (`decipher/mod.rs:73-77`).** It warns only about ordering and prescribes binding position — which doesn't close what's open. `for_sequence_element` takes no index or length:

    duplicate [e0,e0]  -> Ok(["alice","alice"])     truncate [e0] -> Ok(["alice"])
    splice same-AAD    -> Ok(["alice","mallory",…]) truncate to []-> Err

Deliberate and documented at `aad/mod.rs:140-143`, so this is a doc fix. The map note two paragraphs down is the model — name the failure class, warn against inferring coverage from the empty-container rejection. Worth a test alongside `reordered_map_entries_still_decrypt`. Neither README mentions the caveat.

### Worth fixing

- **Writeback uses `[[Set]]`** (`convert.rs:267,280`, `ciphertext.rs:61-103`). Under `Object.freeze(Object.prototype)` or `--frozen-intrinsics`, a plaintext key named `toString`/`valueOf`/`hasOwnProperty`/etc. is silently dropped on decrypt — `Set` runs `kDontThrow`, `napi_ok` propagates. No attacker needed; encrypt accepts those keys happily. Fix: `napi_define_properties` with `writable|enumerable|configurable`. (The pollution-exfiltration angle is weak — needs `defineProperty`, i.e. JS that already owns the addon.)
- **Duplicate keys in `NapiValue::Object`** (`value.rs:137-146`) encrypt fine and never decrypt. Unreachable from JS; a Rust-caller footgun. Dedup or document the precondition.
- **`SeqAccess::next_element` contradicts `decrypt_seq`** (`decipher/mod.rs:199-204`), and `SeqCipher::encrypt_next` is stale the same way — `1ebf542` never touched `aead/src/cipher.rs`. Nothing shipped is at risk; the map methods name their derivations inline and these should too.

### Nits

- PR body's "NaN payloads survive" is false across the JS boundary (V8 canonicalises in `Number::New`); the doc at `value.rs:25-26` is fine, it's scoped to the encoding. `roundtrip_number_edge_cases` never crosses into V8.
- Six rustdoc warnings: `aad/mod.rs:138,154`, `cipher.rs:139`, `decipher/mod.rs:133`, `value.rs:48,75`. Nothing fails.
- `i.to_string()` per index (`convert.rs:193`) — use `itoa`.
- `ensure_plain_object` is 5 napi calls per node; hoisting the baseline saves 2N-2.
- `check_status!` at `convert.rs:73,107,115` for real statuses and fewer lines.
- `EmptyMap`'s "see `decrypt_seq`" pointer (`cipher.rs:519`) went stale on a field rename; make it self-contained.
- `MAX_DEPTH` is fine (~145 KiB worst case vs ~984 KiB), but the encrypt path costs ~1.1 KB/level — 3-8x the converter it was sized against. Worth a comment for whoever raises it.

PR #256, opened 2026-07-24
feat(aead-value)!: extract language-neutral FfiValue crate with INT64 tag

### LINKED ISSUE #272: Make the FFI value model language-neutral before the Node addon ships
The value tree and its sealed leaf encodings currently live inside the Node-specific bridge crate, but nothing in them is JS-specific. Python (PyO3), Go (WASI/wazero, cipherstash-suite#2099) and the EQL layer will all produce and consume the same encodings, and a value encrypted from one language must decrypt from any other. That makes the leaf tag table a cross-language wire commitment — so it has to be frozen and documented in a neutral crate *before* the Node addon ships and starts writing data.

The value tree, its encrypt/decrypt impls, and the tag table should move into their own crate, with the JS crate reduced to a newtype carrying the napi conversions (the orphan rule prevents anything simpler). The neutral crate needs known-answer tests pinning every leaf encoding byte-for-byte — these double as cross-language conformance vectors — plus malformed-leaf hardening: truncated and overlong fixed-width payloads, trailing bytes after payloadless tags, unknown tags, and invalid UTF-8 all rejected.

The model also has a representational hole to close. JS collapses all numbers to f64, but Python `int` and Go `int64` are real types and `42` must not round-trip as `42.0`; unsigned values above `i64::MAX` have no representation at all. Signed and unsigned 64-bit integer tags should be added additively, with a defined decode mapping documented for every supported language (including how an "undefined" value decodes in languages that have no such concept).

Finally the docs should state the contract plainly: what is frozen is the tag table plus the sealed leaf encodings, not any Rust enum — bindings implement the model, not the type. Record the governance rule that a new tag is added only when a value class cannot be represented in the existing model, and only with decode mappings for every supported language. The in-memory container projection should be documented as an app-side convenience, since EQL owns the durable database format and treats sealed leaves opaquely.

Addressed by #256 (stacked on #255 → #254 → #253).

coderdan:
## Summary

> ⚠️ **Stacked on #255** (→ #254 → #253). Diff here = the extraction + INT64 + UINT64. Retarget as the base PRs land.

Fourth step, and the one with deadline pressure: the value tree and its leaf wire format must be language-neutral **before** the Node addon ships, because the same encodings will be produced/consumed by the Python (PyO3) and Go (WASI/wazero, per cipherstash-suite#2099) bindings, and walked by the EQL layer. A value encrypted from one language decrypts from any other — that makes the tag table a frozen cross-language wire commitment.

### New crate: `vitaminc-aead-value`

- **`FfiValue`** (formerly `NapiValue`) — the owned, `Send`, self-describing tree, its `Encrypt`/`Decrypt` impls, and the now-public `tags` module move here unchanged in behaviour. Nothing in them was JS-specific.
- **`INT64 = 0x07` / `FfiValue::Int(i64)`** (two's complement, LE) — new, additive. JS collapses numbers to f64, but Python `int` and Go `int64` are real types; `42` must not round-trip as `42.0`.
- **Known-answer tests** pin every leaf encoding byte-for-byte via a capturing mock cipher — these are the cross-language conformance vectors (e.g. `Null → [0x00]`, `Number(1.5) → [0x04, …, 0xF8, 0x3F]`, `Int(-1) → [0x07, 0xFF×8]`). The README documents the frozen tag table and the per-language type-mapping matrix (including `Undefined` → `None`/`nil` decode for non-JS languages).
- Malformed-leaf hardening tests: truncated/overlong fixed-width payloads, trailing bytes after payloadless tags, unknown tags, invalid UTF-8 — all rejected.

### `UINT64 = 0x08` — closing the unsigned-range hole

- **`UINT64 = 0x08` / `FfiValue::UInt(u64)`** (8 bytes, LE) — unsigned values above `i64::MAX` were the model's only genuine representational hole; this closes it. Same never-reuse/never-renumber treatment and byte-pinned KATs (`UInt(u64::MAX) → [0x08, 0xFF×8]`, truncated/overlong rejection).
- **NAPI numeric rule** now: JS `BigInt` → `Int` if it fits `i64`, else `UInt` if it fits `u64` (via napi's lossless `get_u64`), else rejected. JS `number` still **always** → `Number`. A decoded `UInt` surfaces as a JS `number` within `Number.MAX_SAFE_INTEGER` and a `BigInt` beyond it. The cross-language matrix adds a `UInt` row (Go `uint64`, Python `int`, JS `number`/`BigInt`).

### Model reframing (docs)

The crate/README docs now state plainly that the cross-language **contract is the tag table plus the sealed leaf encodings** (`[tag] ++ payload`, inside the AEAD envelope); `FfiValue` is merely Rust's materialization of that model, and bindings implement the model, not the enum. The governance rule is recorded verbatim: *a new tag is added only when a value class cannot be represented in the existing model, and requires a defined decode mapping for every supported language before it ships.* `UINT64` is the worked example.

### `vitaminc-aead-napi` shrinks to the JS layer

- `NapiValue` becomes an orphan-rule **newtype** over `FfiValue` carrying the `FromNapiValue`/`ToNapiValue` conversions (napi's traits + foreign type can't meet otherwise). `FfiValue` is re-exported for addon convenience.
- `JsCipherText` docs downgraded from "stable convention" to **in-memory/app-side convenience**, reflecting the EQL layering decision: EQL owns the durable cross-language database format and treats sealed leaves opaquely; only the leaf byte format is frozen.

## Breaking changes

- `vitaminc-aead-napi::NapiValue` is now a newtype — construct trees as `FfiValue`, wrap/unwrap at the `#[napi]` boundary. No wire-format change for existing tags; `INT64` and `UINT64` are additive.

## Testing

- Value crate KATs, round-trips (incl. `Int`/`UInt` extremes and Int-vs-Number-vs-UInt variant preservation), tamper/malformed-leaf rejection, key-swap detection; napi crate keeps its boundary-guard tests. Full workspace suite, clippy, `cargo fmt --check` clean.

Closes #272


packages/encrypt/src/cipher.rs
```diff
@@ -76,6 +76,46 @@ impl<'c> Cipher for &'c Aes256Cipher {
...
+        // keeps ownership of the original and is wiped by ZeroizeOnDrop at
+        // end of scope — a `risky_unwrap` would partially-move the bare
+        // `[u8; N]` out and skip that wipe (see issue #170).
+        let src = data.risky_ref();
+        let mut buf = Vec::with_capacity(N + TAG_LEN);
+        buf.extend_from_slice(src);
+        let plaintext = Protected::new(buf);
+
+        let nonce = self.nonce_generator.generate()?;
+        let nonce_bytes: [u8; NONCE_LEN] = nonce.as_ref().try_into().map_err(|_| Unspecified)?;
```
coderdan: This sort of thing is why the CipherTextBuilder exists.

packages/aead-value/src/tagged.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// binding.
+    pub const TAG: u8 = HDR;
+
+    /// Construct a payloadless tagged leaf (`N == 1`): the tag *is* the whole
+    /// plaintext. Used for Null/Undefined/Bool.
+    pub fn tag_only() -> Self {
+        // No heap allocation: the one-byte plaintext lives on the stack until
+        // it is sealed, then is wiped when the `Protected` drops.
+        Self(Protected::new([HDR; N]))
+    }
```
coderdan: This only be implemented for when N actually is 1! Not for any N.

packages/aead-value/src/tagged.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    /// Assemble from the raw little-endian payload bytes, writing the header
+    /// into slot 0 on the stack. The `From` impls on the aliases are the
+    /// ergonomic entry point; this is the shared core.
+    fn from_payload(payload: &[u8]) -> Self {
```
coderdan: Suggest renaming to copy_from_slice because semantically that's exactly what is happening.

packages/aead-value/src/tagged.rs
```diff
@@ -0,0 +1,281 @@
+//! Tagged leaf plaintexts: the `[tag] ++ payload` assembly, typed.
```
coderdan, line 1: We ought to explain why tagged values are even needed (FFI use-case).

packages/aead/src/aad/mod.rs
```diff
@@ -72,6 +72,33 @@ impl<'a> Aad<'a> {
         const MAP_ENTRY_DOMAIN: &[u8] = b"vitaminc/aead/map-entry/v1";
         pae::encode(&[MAP_ENTRY_DOMAIN, self.as_bytes(), key.as_bytes()])
     }
+
+    /// Derives the AAD a leaf must be sealed against when a **schema-aware**
+    /// caller wants to bind the leaf's expected *type* into authentication.
+    ///
+    /// This is the expectation-side type binding. A caller that knows what
+    /// type a field must hold (the future `EqlCipher`) mixes the expected
```
coderdan: Remove the mention of EqlCipher. That might not be a thing - don't guess ahead.

packages/aead/src/aad/mod.rs
```diff
@@ -72,6 +72,33 @@ impl<'a> Aad<'a> {
...
+    /// binding lets a schema enforce a type up front.
+    ///
+    /// The encoding is `PAE(domain, aad, [tag])`. The leading
+    /// domain-separation label keeps the result disjoint from
+    /// [`for_map_entry`](Aad::for_map_entry) and from user-supplied composite
+    /// AAD.
+    pub fn for_leaf_type(&self, tag: u8) -> Aad<'static> {
+        const LEAF_TYPE_DOMAIN: &[u8] = b"vitaminc/aead/leaf-type/v1";
+        pae::encode(&[LEAF_TYPE_DOMAIN, self.as_bytes(), &[tag]])
+    }
```
coderdan: This function is only relevant to the values in the AEAD value crate. It shouldn't be in the generic AEAD crate. It makes the abstraction leaky.

Instead, define a name type in the value crate and implement IntoAad. That was literally what that trait is for (allowing upstream crates to define their own Aad types).

coderdan:
Rebased onto the security fixes in #253 (https://github.com/cipherstash/vitaminc/pull/253#issuecomment-5102281938). No functional changes needed in this branch — the only conflict was textual, where this PR's `Aad::for_leaf_type` and the new marker-AAD helpers (`for_empty_sequence`/`for_empty_map`/`for_none`) landed in the same region of `aad/mod.rs`; both are kept, and `for_leaf_type` still moves out to `aead-value` in the later commit as designed.

coderdan:
Rebased onto the AAD-once trait change in #253 (https://github.com/cipherstash/vitaminc/pull/253#issuecomment-5103318919). Two mechanical updates: `FfiValue`'s `Encrypt` impl moves its AAD to `encrypt_seq`/`encrypt_map` (dropping the per-element clone), and the KAT capturing-cipher test doubles follow the new sub-cipher signatures. The `encrypt_bytes_array` override from `perf(encrypt): seal fixed-width arrays without a reallocating copy` is preserved. Ciphertext bytes unchanged.

coderdan:
Rebased, plus one fix of this branch's own: clippy's `byte_char_slices` fired on `[b'n']` in `aead-value/src/aad.rs` (now `*b"n"`), which was failing the Test gate here and, inherited, on #257.

Also ported the `Debug` redaction test to `FfiValue`. The `NapiValue` version added on #255 didn't survive this branch's extraction of the value tree into `vitaminc-aead-value`, and the impl it covers moved with it — so the same 0%-coverage gap would have reappeared here. The port covers the wider numeric family and the passthrough wrapper too, pinning that secret leaves stay `<redacted>` through every container.

All three checks pass.

coderdan (commented):
Automated review findings (verified, one refuted/scope-excluded candidate omitted).

One finding with no inline anchor in this diff:

**Release-group crate count is stale.** CLAUDE.md:16 and RELEASING.md:15,44 say "All 13 crates release together", but with `vitaminc-aead-value` added, release-plz.toml now has 14 `[[package]]` entries in the version group. Commit ec6ba9a corrected the count for aead-napi only; the aead-value extraction never re-updated it.

packages/encrypt/src/cipher.rs
```diff
@@ -143,6 +174,16 @@ impl<'c> Cipher for &'c Aes256Cipher {
     fn passthrough(self, value: Self::Passthrough) -> Result<Self::Ok, Self::Error> {
         Ok(AesCipherText::Passthrough(value))
     }
+
```
coderdan, line 208: **All-passthrough containers encrypt Ok but can never decrypt.** `AesSeqCipher::encrypt_next` (line ~224) and `AesMapCipher::encrypt_value` (line ~315) set `self.encrypted = true` unconditionally — even when the nested `Encrypt` impl routed through `passthrough_boxed` and produced an unauthenticated `Passthrough` node. That bypasses `end()`'s all-passthrough rejection on the encrypt side, while `decrypt_seq`/`decrypt_map` categorically reject all-passthrough containers.

`FfiValue::Array(vec![FfiValue::Passthrough(...)])` — the exact non-sensitive-metadata shape the aead-napi README advertises — encrypts Ok into a Sequence of only Passthrough nodes carrying no AEAD tag; every subsequent decrypt hits the `any(!Passthrough)` guard and returns `Err(Unspecified)`. Empirically reproduced: encrypt Ok, decrypt permanently Err — silent data loss plus an entirely unauthenticated container.

Fix: set `encrypted` only when the produced node is not `Passthrough`.

packages/aead-value/src/value.rs
```diff
@@ -0,0 +1,1179 @@
...
+                        c.encrypt_next(item)
+                    })?
+                    .end()
+            }
+            FfiValue::Object(entries) => {
+                // Mirror the built-in `HashMap` impls; the cipher binds each
+                // key into its value's AAD via `Aad::for_map_entry`.
+                entries
+                    .into_iter()
+                    .try_fold(cipher.encrypt_map(aad), |c, (key, value)| {
```
coderdan, line 212: **Duplicate object keys: encrypt Ok, decrypt always Err.** `FfiValue::Object` is a public `Vec<(String, FfiValue)>` with no duplicate-key check anywhere on the encrypt path (`AesMapCipher` documents "no duplicate-key checks"), while the new `decrypt_map` hard-rejects duplicate keys.

A Rust caller or future non-JS binding can build `FfiValue::Object(vec![("a", Int64(1)), ("a", Int64(2))])`; encrypt succeeds, the ciphertext is stored, and every decrypt hits the HashSet duplicate rejection — permanent, silent data loss discovered only at read time.

Symmetry fix: reject the duplicate at encrypt time (`AesMapCipher::encrypt_key` already accumulates entries), so the contract holds for every encoder, not just HashMap and the JS binding.

packages/aead-value/src/value.rs
```diff
@@ -0,0 +1,1179 @@
...
+            FfiValue::Undefined => TaggedUndefined::tag_only().encrypt_with_aad(cipher, aad),
+            FfiValue::Bool(false) => TaggedBoolFalse::tag_only().encrypt_with_aad(cipher, aad),
+            FfiValue::Bool(true) => TaggedBoolTrue::tag_only().encrypt_with_aad(cipher, aad),
+            FfiValue::Int32(i) => TaggedInt32::from(i).encrypt_with_aad(cipher, aad),
+            FfiValue::Int64(i) => TaggedInt64::from(i).encrypt_with_aad(cipher, aad),
+            FfiValue::UInt32(u) => TaggedUInt32::from(u).encrypt_with_aad(cipher, aad),
+            FfiValue::UInt64(u) => TaggedUInt64::from(u).encrypt_with_aad(cipher, aad),
+            FfiValue::Float32(f) => TaggedFloat32::from(f).encrypt_with_aad(cipher, aad),
+            FfiValue::Float64(f) => TaggedFloat64::from(f).encrypt_with_aad(cipher, aad),
+            FfiValue::String(s) => TaggedString::new(s).encrypt_with_aad(cipher, aad),
```
coderdan: **Non-UTF-8 `FfiValue::String`: encrypt Ok, decrypt always Err.** The doc here claims "Validated UTF-8 at construction", but the payload is a publicly constructible `Protected<Vec<u8>>` and `TaggedVariable::new` performs no UTF-8 check — while the decrypt visitor's STRING arm rejects non-UTF-8.

`FfiValue::String(Protected::new(vec![0xFF, 0xFE]))` compiles and encrypts successfully; every decrypt reaches `std::str::from_utf8`, fails, and returns `Err(Unspecified)` — the sealed value is permanently unreadable with no error at seal time.

Either validate in the encrypt arm/constructor, or make the variant hold a validated string type.

packages/aead-napi/src/convert.rs
```diff
@@ -234,19 +286,44 @@ fn js_to_value(unknown: Unknown<'_>, depth: usize) -> Result<NapiValue> {
...
+        FfiValue::Float32(f) => unsafe { f64::to_napi_value(env, f64::from(f)) },
+        // 32-bit integers always fit `Number.MAX_SAFE_INTEGER`, so they are
+        // always a plain JS `number`.
+        FfiValue::Int32(i) => unsafe { f64::to_napi_value(env, i as f64) },
+        FfiValue::UInt32(u) => unsafe { f64::to_napi_value(env, u as f64) },
+        FfiValue::Int64(i) => {
+            // Exactly representable → plain number (the common case, and
+            // what JS callers expect for e.g. a Python-written `30`).
+            // Beyond ±MAX_SAFE_INTEGER → BigInt, losslessly.
+            if i.abs() <= MAX_SAFE_INTEGER {
```
coderdan: **`i64::MIN` breaks the number-vs-BigInt split.** `i.abs()` panics on `i64::MIN` under overflow checks ("attempt to negate with overflow" inside Node), and in release wraps to a negative value that wrongly takes the plain-number branch — returning JS number `-9223372036854775808` instead of `BigInt`. Re-encrypting that number then seals it as FLOAT64 instead of INT64, silently flipping the authenticated type across a round trip.

Fix: `i.unsigned_abs() <= MAX_SAFE_INTEGER as u64`.

packages/aead-napi/src/convert.rs
```diff
@@ -151,31 +176,58 @@ fn ensure_plain_object(obj: &Object<'_>) -> Result<()> {
...
+                "BigInt value does not fit in a 64-bit integer",
+            ))
+        }
         ValueType::String => {
             // `String::from_unknown` copies out of the V8 heap; the copy is
             // moved into `Protected` without further duplication. The V8
             // original is owned by the engine and cannot be wiped from here.
             let s = String::from_unknown(unknown)?;
-            Ok(NapiValue::String(Protected::new(s.into_bytes())))
+            Ok(FfiValue::String(Protected::new(s.into_bytes())))
```
coderdan: **Unpaired surrogates are silently sealed as U+FFFD.** `String::from_unknown` (`napi_get_value_string_utf8`) replaces unpaired UTF-16 surrogates with U+FFFD before sealing — mutating the plaintext with no error, contradicting this module's own refuse-loudly principle.

A JS caller encrypting `'a\uD800b'` (a valid JS string, e.g. produced by slicing emoji-containing input) actually seals `'a\uFFFDb'`; decrypt returns a different string than was passed and equality checks against the original fail — silent, unrecoverable corruption.

Either reject ill-formed strings loudly (probe via UTF-16) or document the U+FFFD normalization.

packages/aead-napi/src/convert.rs
```diff
@@ -281,13 +358,19 @@ fn value_to_js(env: sys::napi_env, value: NapiValue) -> Result<sys::napi_value>
             }
             unsafe { Object::to_napi_value(env, obj) }
         }
+        // A passthrough subtree projects onto its plain JS value: the marking
```
coderdan, line 426: **Decrypt output writes can hit polluted prototype setters.** The decrypted-object projection writes properties with `Object::set` (`napi_set_property`, [[Set]] semantics walking the prototype chain) onto an ordinary `Object.prototype`-inheriting object, so a polluted inherited setter on any non-forbidden key receives decrypted plaintext; `forbidden_key` blocks only `__proto__`/`constructor`/`prototype`. Same pattern in `ciphertext.rs` `node_to_js`.

If the host app (or a dependency) has polluted `Object.prototype` with `defineProperty(..., 'email', {set(v){exfil(v)}})`, decrypting `{email: ...}` invokes the setter with the plaintext (leak) and the result object lacks the own property (drop). The intake side already treats prototype pollution as in-threat-model (`own_enumerable_keys`/`ensure_plain_object`); the output side is exposed.

Harden with a null-prototype result object or define-own-property writes.

packages/encrypt/src/cipher.rs
```diff
@@ -101,6 +101,37 @@ impl<'c> Cipher for &'c Aes256Cipher {
...
+    {
+        // Override the trait default (which does a plain `to_vec`, sized
+        // exactly `N`, forcing the in-place seal to reallocate when it appends
+        // the tag). `append_target_plaintext_array` pre-sizes the buffer to
+        // `N + TAG_LEN` so the seal appends the tag into spare capacity and the
+        // ciphertext is produced with a single heap allocation. Output bytes
+        // are identical to the default path.
+        let nonce = self.nonce_generator.generate()?;
+        let nonce_bytes: [u8; NONCE_LEN] = nonce.as_ref().try_into().map_err(|_| Unspecified)?;
+        // Outermost derivation: bind the wire version — see `Aad::for_leaf`.
```
coderdan, line 151: **Wire-format-critical leaf seal/open logic is triplicated.** `decrypt_local_ciphertext` duplicates `open_local` (cipher_static.rs), and `seal_empty_marker`/`encrypt_bytes_vec`/`encrypt_bytes_array` duplicate `seal_into_local`, with `.for_leaf(WIRE_VERSION)` repeated at 6 sites — this branch already had to make the identical version-binding edit in multiple places (the rebase caught `encrypt_bytes_array` missing it).

The copies are line-for-line equivalent and share no type or feature barrier preventing extraction. The next wire-format change applied to one copy but not another produces silently cross-incompatible ciphertexts between the static and dynamic paths, each still round-tripping individually so no test fails.

Fold into one shared private `seal_leaf`/`open_leaf` pair.

packages/aead-value/src/value.rs
```diff
@@ -0,0 +1,1179 @@
...
+            FfiValue::Bytes(b) => TaggedBytes::new(b).encrypt_with_aad(cipher, aad),
+            FfiValue::Array(items) => {
+                // Mirror the built-in `Vec<T>` impl: every element sealed
+                // against `Aad::for_sequence_element` of the caller's AAD.
+                // Element positions are carried structurally only — order is
+                // NOT authenticated, and a tampered ciphertext with permuted
+                // elements still decrypts. See `Aad::for_sequence_element`
+                // for why the index is deliberately unbound.
+                let len = items.len();
```
coderdan: **This arm duplicates the built-in `Vec<T>` `Encrypt` impl token-for-token** (packages/aead/src/encrypt/impls.rs:42) — the comment even says "Mirror the built-in `Vec<T>` impl". Since `FfiValue: Encrypt`, it can simply be `items.encrypt_with_aad(cipher, aad)`; `Passthrough` elements route identically in both forms (via each element's own `Encrypt` impl).

As written, a protocol fix in the built-in `Vec<T>` impl would silently not propagate to `FfiValue` trees, forking the sequence wire protocol.

packages/aead-napi/src/convert.rs
coderdan: [P2] Handle `i64::MIN` without calling `abs`

When an `INT64` containing `i64::MIN` is converted to JavaScript, `i.abs()` panics in debug builds; in release it remains negative, passes this comparison, and is emitted as an imprecise `f64` instead of a `BigInt`. Please use a range comparison or `unsigned_abs()` so this valid wire value remains lossless.

coderdan (commented):
Review performed by GPT-5.6 Sol. I found two edge cases that can produce incorrect or unreadable output; details are inline.

packages/aead-value/src/value.rs
```diff
@@ -0,0 +1,1179 @@
...
+                items
+                    .into_iter()
+                    .try_fold(cipher.encrypt_seq(Some(len), aad), |c, item| {
+                        c.encrypt_next(item)
```
coderdan: [P2] Keep all-passthrough containers from becoming unreadable

When an array or object contains only `Passthrough` children, routing them through `encrypt_next`/`encrypt_entry` makes the AES sub-cipher mark the container as encrypted even though every produced node is passthrough. Encryption therefore succeeds, but `decrypt_seq`/`decrypt_map` rejects the resulting container because it contains no authenticated node. These inputs should either be rejected during encryption or use passthrough-aware sub-cipher methods.

packages/aead-value/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+> A new tag is added only when a value class cannot be represented in the
+> existing model, and requires a defined decode mapping for every supported
+> language before it ships.
+
+The numeric family is the worked example. It divides the number space by
+both **signedness** and **width**: `INT32`/`INT64`/`UINT32`/`UINT64` for
+exact integers and `FLOAT32`/`FLOAT64` for IEEE-754 values. The unsigned
+tags close a genuine representational hole (values above `i64::MAX` had no
+home); the 32-bit widths exist for **schema fidelity with the EQL layer**
```
freshtonic, line 38: This explanation is load-bearing governance, not just background: it makes clear that width is preserved as schema type semantics, so the 32-bit variants are distinct value classes rather than space-saving aliases for the 64-bit tags. Good to pin this before declaring v2 frozen.

freshtonic (approved):
## Standards

Pass. Moving the language-neutral value model out of the NAPI crate removes domain duplication, and the README treats the tag table as a governed cross-language contract rather than a Rust enum detail.

## Spec

Pass. The extraction, signed/unsigned integer coverage, KATs, malformed-leaf rejection, and language mappings satisfy #272. I specifically checked the extra 32-bit tags against the growth rule: the PR defines width as schema-level type fidelity (`int4`/`float4`), records the one-time pre-release v2 renumber, and freezes subsequent evolution.

Approved. This is the right point to make that commitment, before any v2 ciphertext ships.

tobyhede (commented):
## Review: `feat/aead-value` → `feat/aead-napi`

Reviewed at `911d8d1`. Clippy and the touched crates' tests pass. Each finding was re-checked by an independent pass told to refute it; only what survived is listed.

### Should fix before merge

**Decrypted arrays use `[[Set]]`, not own-property defines** — `aead-napi/src/convert.rs:408`
`create_array` → `napi_create_array_with_length` gives a holey array; `Array::set` → `napi_set_element` → `v8::Object::Set` walks the prototype chain. Verified on Node v22.23.1: a setter on `Array.prototype['0']` (or `Object.prototype['0']`) receives the decrypted element, and no own property is created — leak *and* drop. Same via `Object::set` on the fixed `t`/`v` keys in `ciphertext.rs` (68, 73, 78, 83, 85, 90, 111, 116), which corrupts ciphertext nodes rather than leaking plaintext. `aead-napi/README.md:78-81` claims a polluted `Object.prototype` setter "can never observe decrypted plaintext or swallow a property" — false for arrays; reword with the fix.

**`node_from_js` accepts a node the encrypt side can't produce** — `aead-napi/src/ciphertext.rs:220`
Unauthenticated passthrough is deliberate and documented (`container.rs:52-54`, test at `aead-value/src/value.rs:1133`) — not the issue. The issue is intake asymmetry: `js_to_value` never constructs `FfiValue::Passthrough` (`convert.rs:179-186`), so no legitimate JS ciphertext contains one, yet `node_from_js` parses `{t:"pt", v: …}` back out of a store this file calls attacker-writable (`:96-98`). Nothing binds whether a position was *meant* to be passthrough, and the container guards need only one sealed sibling (`any`, not `all` — `encrypt/src/cipher.rs:537`, `:574`). Suggest rejecting `t:"pt"` on intake until the JS passthrough API lands. Not exploitable against serialized Rust ciphertext (no `Serialize`, payload is `Box<dyn Any>`).

### Worth fixing

**Duplicate-key check makes map sealing O(n²)** — `encrypt/src/cipher.rs:336`, `:374`
Per-key linear scan; `decrypt_map:584` already does this in O(n) with a `HashSet`. New in this PR (base was a bare assignment). Measured vs. the same build with only the scans removed: 2.3x at n=500, 25x at 10k, 182x at 50k (4.00 s vs 22 ms). Crypto is flat at ~440 ns/entry, so the crossover is n ≈ 420; nothing bounds map width on encrypt. Mirror the `HashSet`, or a small-map threshold hybrid.

**UTF-16 intake leaves plaintext un-zeroized** — `aead-napi/src/convert.rs:274`
`Utf16String` is a bare `Vec<u16>` with no `Drop`/`Zeroize` (napi 3.11.0 `string.rs:105`), freed un-wiped; `String::from_utf16` sizes in u16 units, so non-ASCII reallocates and leaves further partial copies. A regression — the old `String::from_unknown` produced one buffer that `Protected` wiped. napi-rs exposes no way to reach the buffer, so the fix is `sys::napi_get_value_string_utf16` into a `Protected<Vec<u16>>`. (`zeroize` is a declared dep of this crate, unused throughout `src/`.)

**`seal_leaf` doc asserts a test that doesn't exist** — `encrypt/src/cipher.rs:64-66`
"its output bytes are pinned equal to this path by test" — no such test in the repo, and none can be written: `Aes256Cipher`'s `RandomNonceGenerator` has no injection seam. The duplication is real (`seal_leaf` 75-90 vs `encrypt_bytes_array` 149-163, differing only in `append_target_plaintext_array`), but drift is caught by round-trips through the single shared `open_leaf`. Just correct the sentence.

**`TaggedString::new` seals unvalidated bytes under the STRING tag** — `aead-value/src/tagged.rs:168`
`pub` and externally reachable (`lib.rs:4`, alias at `:189`, crate is published). Verified from external-crate position: `vec![0xFF, 0xFE]` encrypts `Ok`, then fails every decrypt — the silent read-time data loss `Utf8String` exists to prevent (`value.rs:127-135`). Misuse door on a wire-format type, not a live path. Same shape at `:72`, where `tag_only`'s `HDR` is unconstrained.

### Docs and hygiene

- `aead-napi/src/value.rs:15-17` — doc describes a `Number`/`Int` split; those variants no longer exist post-renumber, and the `UInt64` BigInt rule is missing.
- Passthrough rationale contradicts itself: `convert.rs:427` ("an ordinary value to the application") vs `aead-value/src/value.rs:315` ("records which fields travelled in the clear"). Neither README mentions passthrough, so the real mitigation — callers can reject `{t:"pt"}` nodes in the ciphertext they hold — is undocumented.
- `convert.rs:179-186` — the "no path constructs `FfiValue::Passthrough`" note was correct until `911d8d1` inserted `define_own_property` between it and `js_to_value`; it now reads as documenting the wrong function. Move it to line 238.
- `.cargo-crap.toml:45-47` and `.cargo/mutants.toml:29-31` — `define_own_property` was added to both lists, neither closed-list parenthetical updated (README was). These suppress gates on `unsafe` FFI code.
- PR description carries the pre-renumber table (`INT64 = 0x07`, `UINT64 = 0x08`); shipped is `0x05`/`0x07`. Every in-repo artifact is correct — description only.
- `5db5c8b` adds required `Cipher::passthrough_boxed` (`aead/src/cipher.rs:175`) with no `!`/footer. Low impact (unreleased 0.x, trait already rewritten past the last release, zero in-tree fallout) but worth the marker for `cargo-semver-checks`.

### Cleanups

- `convert.rs:200-215` is byte-identical to `:122-137`; `<&str>::to_napi_value` already does this at `:395`.
- `aead-value/src/value.rs:279-283` hand-rolls the check that `TryFrom` at `:167-176` performs. (The re-validation at `convert.rs:393` is unreachable but not deletable — it's the fallible-by-signature `&[u8] → &str` at the FFI boundary; removing it needs `unsafe` or a `Utf8String::as_str()`.)

### Checked and cleared

`value_to_js` lacks the `MAX_DEPTH` guard its siblings have, but this isn't a defect here: the asymmetry predates the PR, `Passthrough` can't nest (terminal in `node_from_js`, unconstructible in `js_to_value`, single-wrap in `visit_passthrough`), and worst case from JS is ~258 frames. No serde on the container, so no unbounded producer exists.

PR #258, opened 2026-07-26
feat(prf): add structured PRF foundation

### LINKED ISSUE #275: Derive deterministic index terms from structured values with a keyed PRF
Searchable encryption needs a keyed PRF evaluated over structured plaintext values to produce deterministic index terms. vitaminc has no PRF primitive today: every consumer would hand-roll byte concatenation over its fields, which invites ambiguous encodings — two different structured values serializing to the same byte stream and therefore colliding on the same term.

The PRF layer should mirror the encryption stack's serde-style shape: a value/visitor API over leaves, sequences, and maps, with typed contexts framed prefix-free so nested structures cannot collide with flat ones, and semantic encoding domains on protected leaves so the same bytes typed differently (e.g. a token vs an email) never produce the same PRF output. Passthrough must be explicit, never implicit.

Backends are pluggable: a protected-key HMAC-SHA256 backend (pad and key normalization delegated to RustCrypto behind ProtectedDigest, keys zeroized) is the reference implementation, and evaluation is deferred via IntoFuture so a remote/KMS-backed implementation can batch an entire nested structure into one round trip. ORE is a sibling primitive with its own crate, not a PRF backend.

Addressed by #258.

coderdan:
## Summary

* add the `vitaminc-prf` crate: a serde-style, backend-agnostic API for deriving structured, domain-separated pseudorandom values (`PrfValue`, `Prf`, sequence and map drivers, visitor machinery, awaitable `ReadyPrf` output)
* support deferred `IntoFuture` execution, resolved heterogeneous child nodes, explicit passthrough, and prefix-free contexts
* enforce semantic `PrfEncoding` domains for protected leaves and typed contexts to prevent ambiguous cross-type encodings
* map builders validate structural invariants: `ValueWithoutKey`, `KeyWithoutValue`, `DanglingKey`, and `DuplicateKey`
* `HashMap` sources derive in deterministic (key-sorted) order, matching `BTreeMap`
* expose the crate from the `vitaminc` facade behind the `prf` feature (which enables `protected`)
* register `vitaminc-prf` in the release-plz version group

The crate contains **no cryptography**. The local HMAC-SHA256 backend lands separately as the `vitaminc-hmac` crate in cipherstash/vitaminc#276 (stacked on this PR) so the implementation can be reviewed and hardened on its own terms. Unit tests here run against a deterministic, non-cryptographic mock backend that implements the full `Prf` contract, proving the abstraction stands alone.

## Review

Findings from the Claude Fable review and an independent GPT-5.6 Sol review are addressed in this PR's history (deterministic `HashMap` ordering, duplicate-key rejection, unified Option context encoding with invariant tests, facade feature fix, changelog revert). The HMAC-specific fixes (zeroized RFC 2104 keying, streamed fixed-size leaves) travel with the backend to cipherstash/vitaminc#276.

## Validation

* `cargo test -p vitaminc-prf` (49 unit tests against the mock backend + doctest)
* `cargo test --workspace --all-features --exclude vitaminc-kms`
* `cargo clippy -p vitaminc-prf --all-targets`
* `cargo fmt --check`

Closes cipherstash/vitaminc#275


packages/prf/src/impls.rs
```diff
@@ -0,0 +1,478 @@
...
+    fn prf_visit_with_context<'a, P, V, C>(self, prf: P, context: C, visitor: V) -> P::Ok<V::Value>
+    where
+        P: Prf,
+        V: PrfVisitor<P::Block, P::Passthrough>,
+        C: IntoPrfContext<'a>,
+    {
+        // Built-in leaves immediately re-wrap bytes in Protected before they
+        // cross the backend boundary.
+        self.risky_unwrap()
+            .prf_visit_with_context(prf, context, visitor)
```
coderdan: [P1] Keep protected inline values wrapped

When `T` is inline, such as `[u8; N]` or an integer, `risky_unwrap` deliberately forgets the wrapper without scrubbing its original slot. The subsequent `PrfValue` implementation wipes only the re-wrapped destination, leaving an abandoned plaintext copy on the stack and defeating `Protected`'s zeroization guarantee.

packages/prf/src/hmac_sha256.rs
```diff
@@ -0,0 +1,304 @@
...
+        Self::new(Protected::new(key.risky_ref().to_vec()))
+    }
+
+    fn derive(
+        &self,
+        data: &Protected<Vec<u8>>,
+        encoding: PrfEncoding,
+        context: &PrfContext<'_>,
+    ) -> [u8; 32] {
+        let framed = PrfContext::pae(&[encoding.as_bytes(), context.as_bytes(), data.risky_ref()]);
```
coderdan: [P1] Zeroize the framed plaintext copy

Whenever `data` is secret, `pae` copies the entire protected value into an ordinary `Vec` owned by `PrfContext`, which has no zeroizing drop. After derivation that allocation is freed unwiped, so heap reuse or a memory dump can recover plaintext despite the original `Protected` value being wiped; stream the framing into HMAC or keep this buffer in a zeroizing container.

packages/prf/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+repository.workspace = true
+homepage.workspace = true
+license.workspace = true
+authors.workspace = true
+keywords.workspace = true
+categories.workspace = true
+
+[dependencies]
+hmac = "0.13.0"
+sha2 = "0.11.0"
```
coderdan: [P1] Enable wiping for HMAC and SHA-256 state

With default dependency features, each derivation drops HMAC's plaintext buffer and SHA-256's key-derived inner/outer states without wiping them. In environments where process memory or core dumps are exposed, this bypasses the protected key lifecycle; enable the dependencies' zeroization features and ensure temporary key-pad state is also scrubbed.

coderdan (commented):
The PRF behavior tests pass, but multiple paths leave protected inputs or key-derived cryptographic state unwiped. These violate the crate's core secret-lifecycle guarantees.

packages/prf/Cargo.toml
coderdan: Fixed in 98b606f. SHA-256 now enables its `zeroize` feature and a compile-time test requires `Sha256: ZeroizeOnDrop`. Production HMAC initialization is local so the normalized key block, ipad/opad state, hashed long key, and inner digest remain in `Protected`; this also avoids the dependency initializer’s ordinary temporary key-pad array. Known-answer and over-block-size-key reference tests pass.

packages/prf/src/impls.rs
coderdan: Fixed in 98b606f. The generic `PrfValue for Protected<T>` implementation is gone. Protected arrays and vectors now cross the backend boundary without unwrapping; protected strings and integers are converted by reference into a new protected byte value while the original wrapper remains live and is wiped on drop.

packages/prf/src/hmac_sha256.rs
coderdan: Fixed in 98b606f. PAE is streamed piece-by-piece directly into the inner SHA-256 state, so derivation no longer allocates an ordinary framed `Vec` containing the protected plaintext.

coderdan:
Review performed by **GPT-5.6 Sol**.

### P1: Zeroize HMAC key-normalization buffers

`packages/prf/src/hmac_sha256.rs:42-43`

Each derivation delegates setup to HMAC 0.13 `new_from_slice`. Its `get_der_key` function creates a plain key-derived `Array` and returns without zeroizing it. After the PRF is dropped, a core dump or later stack disclosure may recover this material; marking only the returned HMAC state `ZeroizeOnDrop` does not wipe constructor temporaries.

Normalize the key in a zeroizing container or use an implementation that wipes these intermediates.

coderdan:
## Code review findings

_This review was performed by **Claude Fable** (Anthropic's Claude Fable 5 model) via Claude Code, at Dan's request._

Six findings survived verification, ranked by severity. The workspace builds clean and all prf/protected tests pass (63 tests + doctests).

---

### 1. 🔴 Correctness — `PrfValue for HashMap` yields nondeterministic entry order
`packages/prf/src/impls.rs:326`

`PrfValue for HashMap` iterates in nondeterministic order, so `MapAccess` entry order (and any position-sensitive or `Vec`-collecting visitor) varies between runs. A user deriving terms from a `HashMap` with a visitor that reads entries positionally (the `RecordVisitor` pattern shown in tests/README) or collects `Vec<[u8;32]>` gets differently-ordered results run to run — intermittent `InvalidValue` errors or unstable serialized index terms. The `BTreeMap` impl is deterministic; nothing warns that `HashMap` is not.

### 2. 🔴 Correctness — `MapPrf` builder silently accepts duplicate keys
`packages/prf/src/hmac_sha256.rs:288`

The builder validates every other structural invariant (`ValueWithoutKey`, `KeyWithoutValue`, `DanglingKey`) but never detects duplicate keys. A handwritten `PrfValue` impl that calls `prf_entry("email", ...)` twice (copy-paste) derives both entries under the identical `for_map_entry` context, `MapAccess` yields duplicate keys, and a visitor collecting into a map silently drops one term — no `PrfBuildError`, no signal that a field's derivation was lost.

### 3. 🟡 Usability — `prf` facade feature does not enable `protected`
`packages/vitaminc/Cargo.toml:32`

A user enabling only `features = ["prf"]` on the vitaminc facade hits a wall: `vitaminc::prf::HmacSha256Prf::new` requires `Protected<Vec<u8>>`, but `vitaminc::protected` is not re-exported, so the API is unusable without discovering the extra feature. The `permutation` feature on line 31 (`permutation = ["dep:vitaminc-permutation", "protected"]`) is the established pattern to follow.

### 4. 🟡 Consistency — two divergent `Option` encodings (context vs value path)
`packages/prf/src/context.rs:158`

`IntoPrfContext for Option` ignores `OPTION_SOME_DOMAIN`, diverging from the `PrfValue` Option path which applies `for_option_some` (context.rs:67). Value-level `Some` is domain-tagged; context-level `Some` is a bare 1-piece PAE. No constructible collision exists today (piece counts disambiguate), but any future unification — or a third Option-handling site picking the "wrong" precedent — silently changes every derived term keyed under Option contexts, an unversionable break for stored index terms.

### 5. 🟡 Process — manual `CHANGELOG.md` edit on a feature branch
`packages/protected/CHANGELOG.md:6`

The repo's CLAUDE.md states: "Do NOT manually edit `CHANGELOG.md` files on feature branches — release-plz manages them". Commit a097483 adds the line "initialize keyed fixed-output functions from protected key material" by hand; release-plz will regenerate this section from commits, risking duplicate or conflicting entries at release time.

### 6. 🟢 Efficiency — `prf_bytes_array` copies every fixed-size secret leaf to a `Vec`
`packages/prf/src/traits.rs:102`

The default `prf_bytes_array` heap-allocates a `Vec` copy of every fixed-size secret leaf, and the `Protected<integer>` impls wrap twice (impls.rs:161–185 adds a third layer for `Protected<int>`). For bulk derivation (e.g. Bloom terms over many fields) that's an avoidable allocation per leaf on the hot path. `HmacSha256Prf::derive` only needs a byte slice, so overriding `prf_bytes_array` to stream the array directly removes the copy.

---

**Refuted during verification** (checked and dismissed): the `ZeroizingHmacSha256` manual-`ZeroizeOnDrop` concern (hmac 0.13 / sha2 0.11 / digest 0.11.2 internals do wipe under the enabled features); Protected-copy hygiene (`Protected` now derives `ZeroizeOnDrop` with a real `Drop`); local-vs-deferred PAE framing mismatch (formats match, quickcheck-covered); test-backend passthrough error-gating (test-only, unobservable).

coderdan:
## Review findings addressed

_Fixes applied by **Claude Fable** (Anthropic's Claude Fable 5 model) via Claude Code, at Dan's request._

All six findings from my earlier review and the GPT-5.6 Sol P1 are now fixed, pushed as seven scoped commits (`80b150e..0df9ca7`). All 70 prf/protected tests pass and `cargo fmt --check` is clean.

### GPT-5.6 Sol P1 — HMAC key-normalization temporaries ✅ confirmed and fixed (`0df9ca7`)
Verified against the hmac 0.13.0 source: `get_der_key` returns the padded key block by value without wiping it, and `new_from_slice` XORs that buffer in place with ipad/opad — both trivially invertible copies of the key left on the stack. (This is distinct from the state-wipe-on-drop claim I refuted earlier, which remains refuted.) `ZeroizingHmacSha256` now performs the RFC 2104 keying itself: every key-derived buffer lives in a `Zeroizing` allocation and the two `sha2` states wipe themselves on drop. Correctness is pinned by the existing known-answer vector plus new quickcheck equivalence against `Hmac<Sha256>` and a deterministic sweep across key-normalization boundaries (0, 1, 63, 64, 65, 131-byte keys).

### 1. HashMap nondeterministic ordering — fixed (`d3619db`)
Entries are now sorted by their derived string key before visiting, so `HashMap` and `BTreeMap` sources produce identical structured results. A new test derives the same entries through both container types and asserts identical ordered output.

### 2. Duplicate map keys — fixed (`ec020d3`)
The builder now surfaces `PrfBuildError::DuplicateKey` from both `prf_value` and `passthrough_entry`, alongside the existing structural invariants. Covered by two new tests, including the mixed derived/passthrough case.

### 3. `prf` facade feature — fixed (`461ec9d`)
`prf = ["dep:vitaminc-prf", "protected"]`, following the `permutation` feature's pattern.

### 4. Divergent Option encodings — unified (`4be7327`)
`IntoPrfContext for Option` now tags `Some` through the same `for_option_some` helper as the value path, so the crate has exactly one `Some` encoding. Two invariant tests were added: a cross-path test documenting that both Option paths share the domain, and a quickcheck collision guard ensuring no bare single-piece PAE can ever equal the Option framing. (This also closes the test gap discussed after the review: the divergence was previously pinned per-path but the relationship between the paths was untested.)

### 5. Manual CHANGELOG edit — reverted (`80b150e`)
`packages/protected/CHANGELOG.md` is restored to the main version; release-plz will regenerate the entry from commits.

### 6. Per-leaf heap copies — fixed (`2b87d4d`)
`HmacSha256Prf` now overrides `prf_bytes_array` and its `derive` is generic over any protected byte container, so integer and fixed-size array leaves stream straight into the digest with no intermediate `Vec`. `Protected<integer>` leaves also use `Controlled::map` instead of re-wrapping. As a side benefit this removes an unwiped-copy surface (the old default's `to_vec()` copy was never zeroized).

packages/prf/src/context.rs
```diff
@@ -0,0 +1,324 @@
...
+    /// The leading domain tag keeps caller-refined contexts disjoint from the
+    /// contexts this crate assigns automatically. Without it, a context built
+    /// as `PrfContext::from_slice(OPTION_SOME_DOMAIN).refine(x)` would encode
+    /// identically to the one derived for `Some(x)`.
+    pub fn refine<'b, C>(&self, component: C) -> PrfContext<'static>
+    where
+        C: IntoPrfContext<'b>,
+    {
+        let component = component.into_prf_context();
+        Self::pae(&[REFINE_DOMAIN, self.as_bytes(), component.as_bytes()])
```
freshtonic, line 65: The dedicated refinement domain is important here: it keeps caller-composed contexts disjoint from the structural contexts assigned internally, while PAE preserves prefix-free composition. This addresses the subtle `Some`/manual-refinement collision class directly.

freshtonic (approved):
## Standards

The new `vitaminc-prf` module follows the repository’s documented conventions, keeps the backend boundary coherent, and introduces no actionable maintainability issues.

## Spec

The structured visitor model, semantic encodings, explicit passthrough, protected-leaf constraints, prefix-free typed contexts, and deferred/batched resolution match the foundation required by #275. The raw `PrfContext` constructors are an intentional escape hatch; automatic and typed contexts remain domain-separated, including caller refinement.

Approved as the foundation layer of the stack. One metadata caveat: because the required HMAC reference backend lands in #276, `Closes #275` should live on the final stack PR (or the two PRs should be merged atomically) so #275 is not closed by this layer alone.

PR #260, opened 2026-07-26
fix(protected): harden digest secret lifecycle

### LINKED ISSUE #259: Harden ProtectedDigest secret lifecycle and public-data channels
## Problem

`ProtectedDigest` does not currently make its lifecycle guarantees enforceable:

- `D: Digest` does not require `ZeroizeOnDrop`, so a digest whose internal state is not wiped can be used behind the protected API.
- `finalize`, `finalize_reset`, and the one-shot `digest` helper first create an ordinary `digest::Array` and only then wrap it, potentially leaving an unwiped inline copy.
- `update` only accepts `Controlled` inputs, forcing public framing/domain-separation bytes to be unnecessarily copied into `Protected`.
- There is no explicit channel for intentionally public digest output.

This was identified while addressing the HMAC-SHA256 lifecycle review on #258.

## Proposed API hardening

- Require the digest implementation to implement `ZeroizeOnDrop`.
- Keep `update` for protected inputs and add an explicitly named `update_public` for public bytes.
- Keep direct protected finalization via `finalize_into` / `finalize_into_reset`.
- Add explicit `finalize_public_into` / `finalize_public_into_reset` methods for intentionally exposed output.
- Remove return-by-value finalization helpers that materialize an ordinary intermediate digest array.
- Replace the one-shot return helper with direct-into protected/public variants if needed.

## Acceptance criteria

- Non-zeroizing digest state cannot instantiate the usable `ProtectedDigest` API.
- Secret inputs and protected intermediate outputs remain behind `Controlled` boundaries.
- Public input and output crossings are explicit in method names.
- No finalization path creates an ordinary temporary digest output before writing its destination.
- SHA-256/SHA-384 protected and public output tests cover incremental and reset finalization.
- Existing workspace tests, Clippy, formatting, MSRV checks, and mutation testing pass.

coderdan:
Closes #259.

## Summary

- require digest state wrapped by ProtectedDigest to implement ZeroizeOnDrop
- keep secret finalization in caller-provided protected storage and remove return-by-value helpers
- add explicit update_public and public-finalization channels
- document the hardened API and enable SHA-2 zeroization in tests

## Validation

- cargo test -p vitaminc-protected --all-features
- cargo test --workspace --all-features --exclude vitaminc-kms
- cargo check --workspace --all-features
- cargo +1.85.1 check -p vitaminc-protected --all-features --tests
- cargo clippy -p vitaminc-protected --all-features --all-targets -- -D warnings
- cargo fmt --all -- --check
- cargo mutants --no-shuffle -vV --in-diff /private/tmp/pr259.diff (3/3 caught)

The commit is GPG-signed with key F7B16D0FFDFCC12F.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -565,6 +565,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cdd35008169921d80bc60d3d0ab416eecb028c4cd653352907921d95084790be"
 dependencies = [
  "hybrid-array",
+ "zeroize",
 ]
 
 [[package]]
@@ -758,6 +759,7 @@ dependencies = [
  "const-oid",
  "crypto-common 0.2.1",
  "ctutils",
+ "zeroize",
 ]
 
 [[package]]
--- a/packages/protected/CHANGELOG.md
+++ b/packages/protected/CHANGELOG.md
@@ -1,4 +1,14 @@
 
+### Security
+
+- require `ProtectedDigest` state to implement `ZeroizeOnDrop`
+- write digest outputs directly into protected or explicitly public destinations
+
+### Breaking changes
+
+- remove return-by-value `ProtectedDigest` finalization helpers
+- add explicit public input and output methods to `ProtectedDigest`
+
 
 
 ### Miscellaneous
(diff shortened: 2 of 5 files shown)


coderdan:
Agent review: Codex/5.6-Sol

Review result: No actionable defects were identified. The protected crate’s tests, doctests, formatting, compilation, and Clippy checks all pass.

PR #268, opened 2026-08-13
test(protected): lock down negative-space guarantees

freshtonic:
## Summary

- add trybuild compile-fail coverage for protected-type negative-space guarantees
- verify controlled wrappers do not implement Deref
- verify bare Protected cannot be compared, serialized, deserialized, copied, or accessed directly
- verify Controlled remains sealed and ProtectedRef cannot be directly constructed
- correct AsProtectedRef documentation to describe its actual construction boundary

## Validation

- cargo test -p vitaminc-protected
- cargo fmt --check

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -957,6 +957,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "glob"
+version = "0.3.4"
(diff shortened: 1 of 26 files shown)


coderdan (approved):
Very cool. Nice work.

PR #276, opened 2026-08-14
feat(hmac): add the vitaminc-hmac PRF backend crate

coderdan:
Stacked on #258, which now ships only the backend-agnostic PRF abstraction. This PR adds the local HMAC-SHA256 backend as its own `vitaminc-hmac` crate so the implementation can be dug into, hardened, and tested on its own terms.

## What's here

- `packages/hmac/src/lib.rs`: `HmacSha256Prf` (keys held in `Protected` behind `ProtectedDigest`), `HmacSeqPrf`/`HmacMapPrf` drivers, and `ZeroizingHmacSha256` — the RFC 2104 keying is performed with every key-derived buffer in a wiped allocation, so construction leaves no unwiped copy of the key (hmac 0.13's own constructor path does).
- `packages/hmac/tests/structured.rs`: backend integration tests — known-answer vectors, oversized-key normalization, local/deferred one-batch equivalence, builder error paths.
- In-crate unit tests: zeroize-on-drop assertions, quickcheck equivalence against RustCrypto `Hmac<Sha256>`, and a deterministic key-normalization boundary sweep (0, 1, 63, 64, 65, 131-byte keys).
- Release plumbing: joins the workspace and the release-plz version group; the facade gains an `hmac` feature (`vitaminc::hmac`) that transitively enables `prf` and `protected`.

## Follow-up candidates (from review discussion)

- Full RFC 4231 test-vector suite
- Assembly-level wipe verification of the keying path
- Possibly genericizing the seq/map resolved-driver logic (currently mirrored between the backend drivers, the core mock, and the deferred test backend)

Tests: `cargo test -p vitaminc-hmac` (21 tests + 3 doctests) and `cargo test --workspace --exclude vitaminc-kms --all-features` green (kms needs LocalStack), clippy and `cargo fmt --check` clean.

---

_PR prepared by Claude Fable (Anthropic's Claude Fable 5 model) via Claude Code, at Dan's request._


packages/hmac/src/lib.rs
```diff
@@ -0,0 +1,517 @@
...
+    type KeySize = <Hmac<Sha256> as KeySizeUser>::KeySize;
+}
+
+impl KeyInit for ZeroizingHmacSha256 {
+    fn new(key: &Key<Self>) -> Self {
+        Self::new_from_slice(key.as_slice()).expect("HMAC-SHA256 accepts keys of any length")
+    }
+
+    fn new_from_slice(key: &[u8]) -> Result<Self, hmac::digest::InvalidLength> {
+        let mut block = Zeroizing::new([0_u8; SHA256_BLOCK_SIZE]);
```
freshtonic, line 92: Good secret-lifecycle choice. Keeping the normalized key block in `Zeroizing` (and likewise the long-key digest below) closes the temporary-buffer gap that a direct generic HMAC constructor would leave, while the equivalence tests retain confidence in the RFC 2104 behavior.

freshtonic (approved):
## Standards

The backend is cohesive, documented, conventionally committed, and well covered. The repeated sequence/map driver machinery is a reasonable future extraction candidate, but it is not a correctness blocker here.

## Spec

This supplies the protected-key HMAC-SHA256 reference backend required by #275, preserves semantic context/input framing, rejects weak imported keys, zeroizes key-derived working state, and demonstrates deferred resolution in one batch. The custom RFC 2104 setup is a deliberate lifecycle-hardening deviation from constructing RustCrypto’s `Hmac` directly: it still uses the RustCrypto digest primitives and is checked for equivalence across boundary lengths, while avoiding unwiped constructor temporaries.

Approved. Minor unrelated clippy/test cleanups could be split for scope hygiene, but do not warrant blocking the stack.

PR #279, opened 2026-08-15
refactor(bindings): vcvalue becomes model-only; the codec moves into vcencrypt

coderdan:
Stacked on #264 (itself on #257). Settles the three design decisions parked on the #257 review threads, per Dan's direction. Implemented by **Claude Fable** (Anthropic's Claude Fable 5 model) via Claude Code, at Dan's request.

## 1. Codec placement — decided: the codec lives with the FFI boundary

`vcvalue` is the durable, dependency-free module a future stack-encrypt Go SDK will import. It now holds **only the value model**: `Plain`, the sealed leaf family (`Sealed`, `SealedNone`, `SealedEmptySeq`, `SealedEmptyMap` with their `driver.Valuer`/`sql.Scanner` impls), and the ordered `Object`/`Field` decode shape.

The transport codec — reader/writer, reflection encoder, natives decoder, ciphertext projection — moves into `vcencrypt`, which owns the wasm boundary the codec exists for. Its entry points (`marshal`/`unmarshal`/`marshalCipherText`/`unmarshalCipherText`) are now **unexported**: the wire format is transport-only, and not exporting it makes the "no storage commitment" claim structural rather than documentary. The only bytes an application can persist are the sealed leaves.

`Encryptable` and the `Encoder` channels move with the codec (their contract *is* the wire recorder), so custom encodings are implemented against `vcencrypt` — the userrecord example is updated accordingly. The cross-language fixture test reaches the unexported decoders via the standard `export_test.go` bridge.

## 2. Unmarshal into caller structs — decided: deferred

Stays documented as future work (now in `vcencrypt`'s package docs, where decode lives). The spike ships dynamic decode shapes.

## 3. `End()` error semantics — decided: first-error-wins is the contract

`SeqEncoder.End`/`MapEncoder.End` return the first error recorded anywhere in the whole encode, not one scoped to their container. Now documented as deliberate and frozen, with `TestEndReturnsFirstEncodeErrorNotContainerScoped` pinning that an inner container's error surfaces from the outer `End`.

## Mechanics

- Pure move plus renames — no codec logic changed; every test moved with its code (fuzz targets, eagerCap pin, malformed-input suites now internal to `vcencrypt`; the driver/scan model tests stay in `vcvalue`).
- Error prefixes in moved code change `vcvalue:` → `vcencrypt:`.
- CI's 386 step now runs both modules, since the load-bearing u32 guards moved with the codec (wazero falls back to its interpreter on 386).
- Both package docs and READMEs rewritten for the new split.

## Verification

Both module suites, `go vet`, `gofmt`, 386 cross-vet, and the userrecord example run green locally; the 386 test run executes in this PR's CI.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -56,8 +56,11 @@ jobs:
         # The codec lives in vcencrypt (wazero falls back to its interpreter
         # on 386); vcvalue is model-only but stays covered too since it is
         # the layer a future stack-encrypt Go SDK imports.
+        # Joined with && because the configured shell (bash -l) does not set
+        # -e: with independent lines, only the last command's exit status
+        # would decide the step, hiding a vcvalue failure.
         run: |
-          (cd bindings/go/vcvalue && CGO_ENABLED=0 GOOS=linux GOARCH=386 go test ./...)
+          (cd bindings/go/vcvalue && CGO_ENABLED=0 GOOS=linux GOARCH=386 go test ./...) &&
           (cd bindings/go/vcencrypt && CGO_ENABLED=0 GOOS=linux GOARCH=386 go test ./...)
 
       - name: run example (userrecord)
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -47,15 +47,18 @@ jobs:
         working-directory: bindings/go/vcencrypt
         run: CGO_ENABLED=0 go test ./...
 
-      - name: test (vcvalue on 32-bit GOARCH)
-        # vcvalue is the durable layer a future stack-encrypt Go SDK imports,
-        # and the u32-bound guards are load-bearing where int is 32 bits: a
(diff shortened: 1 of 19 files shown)


coderdan:
P2 — Make both 32-bit test commands fail the step (.github/workflows/go.yml:60-61)

The configured shell, bash -l {0}, does not enable -e. If the vcvalue 386 test fails but the subsequent vcencrypt test passes, the step exits successfully and hides the failure. Join the commands with && or explicitly enable set -e.

coderdan:
Fixed in 47fc3ae: the two 386 test commands are joined with `&&` (trailing-operator continuation, verified under `bash -l`: a first-command failure short-circuits with exit 1), with a comment noting why — the configured `bash -l` shell has no `-e`, so independent lines would report only the last command's status.

— Claude Fable (Anthropic's Claude Fable 5 model) via Claude Code, at Dan's request.

PR #281, opened 2026-08-16
fix(random)!: replace rejection sampling with fixed-count Lemire draw

### LINKED ISSUE #198: next_bounded_u32: replace rejection sampling with fixed-count Lemire draw (timing channel + biased permutation + password OOB panic)
## Summary

`SafeRand::next_bounded_u32` draws a bounded integer using **rejection sampling**, and `PermutationKey::random` (the Fisher–Yates PRP key derivation) and the password generators all build on it. This has one security issue and two correctness bugs that are best fixed together, because the single-line fix resolves all three.

The timing-channel half of this was just identified and fixed in the sibling `ore.rs` PRP (we replaced rejection sampling with fixed-count wide draws + Lemire reduction).

## 1. Timing side-channel (security)

`packages/random/src/safe_rand.rs:19-32`:

```rust
let cap = max.next_power_of_two();
let mut value = self.0.next_u32() % cap;
while value > max {
    value = self.next_u32() % cap;   // retries until in range
}
```

The number of loop iterations depends on the RNG output. The consumer that matters is `PermutationKey::random` (`packages/permutation/src/key.rs:79-90`), a Fisher–Yates shuffle that calls `next_bounded_u32` once per element with a **secret/plaintext-derived seed** (`SafeRand::from_seed` / `from_controlled_seed`). So PRP-key derivation time is weakly dependent on secret material — the exact timing channel we just removed from `ore.rs`. The password generators (`packages/password/src/lib.rs:70,81,92`) inherit it too (one draw per character).

## 2. Power-of-two branch breaks the inclusive contract → biased permutation (correctness)

`next_bounded_u32` is documented as returning a value "up to and including the given maximum" (`[0, max]`), and the rejection branch honours that (`while value > max` admits `value == max`). But the power-of-two branch (`safe_rand.rs:20-22`) returns `next_u32() % max`, i.e. `[0, max-1]` — **exclusive**.

In `PermutationKey::random`, step `i` needs `j` uniform in `[0, i]`. For every power-of-two `i` (1, 2, 4, 8, …) the draw is silently `[0, i)`, so element `i` can never land in position `i` via that step. At `i = 1` the draw is *always* 0. This biases the generated permutation — a real weakening of the PRP, not just a style nit.

## 3. Reachable out-of-bounds panic in password generation (correctness)

`packages/password/src/lib.rs:70-71`, with `STANDARD_CHARS: [char; 94]` (valid indices `0..=93`):

```rust
let char = rng.next_bounded_u32(STANDARD_CHARS.len() as u32); // [0, 94] inclusive
password[i] = STANDARD_CHARS[char as usize];                  // index 94 -> panic
```

Because `next_bounded_u32(94)` is inclusive, a draw of `94` indexes `STANDARD_CHARS[94]`, which is out of bounds → **panic**, with probability ≈ 1/95 per character. `AlphaNumericPassword` (`next_bounded_u32(62)`, line 81) and `AlphaPassword` (`next_bounded_u32(52)`, line 92) don't panic (62, 52 < 94) but the inclusive bound still includes one extra character beyond the intended sub-alphabet, skewing the distribution.

## Fix: one change resolves all three

Replace rejection sampling with a fixed-count **wide draw + Lemire multiply-high**, returning a **half-open** `[0, range)`:

```rust
/// Uniform in [0, range), no rejection, no modulo bias of practical concern.
fn bounded(rng: &mut impl RngCore, range: u64) -> u64 {
    ((rng.next_u64() as u128 * range as u128) >> 64) as u64
}
```

- **Fixed-count** (exactly one draw) → construction time is seed-independent: closes the timing channel.
- **Branch-free** → no secret-dependent control flow.
- **Uniform to ~2⁻⁵⁵** statistical distance for a 64-bit draw — negligible for these domains (≤ 256, ≤ 94), and a clean statistical term rather than a heuristic.
- **Half-open** semantics fix bugs 2 and 3 by construction: callers pass `range = i + 1` for Fisher–Yates (`[0, i]`) and `range = len` for character indexing (`[0, len)` — never returns `len`, so no OOB and no skew).

Call-site updates:
- `permutation/src/key.rs:82`: `bounded(rng, (i + 1) as u64)`.
- `password/src/lib.rs:70,81,92`: `bounded(rng, STANDARD_CHARS.len() as u64)` / `62` / `52` (or carve the sub-alphabets explicitly).
- Consolidate: there is a second copy of this rejection logic in `packages/random/src/bounded.rs:28-40` (`BoundedRng`); fold to one implementation so the fix can't drift.

## Constant-time note (separate, lower priority)

Beyond the loop, `PermutationKey::random` does `key.swap(i, j)` with secret `j`, and the password generators do `STANDARD_CHARS[secret]` — secret-indexed memory. The `permute`/`depermute` paths are already hardened (`ConditionallySelectable` scans); the *key-generation* path is not. For the small domains here the single-cache-line argument likely suffices (cf. compressed-S-box AES mitigations), but worth a deliberate decision. (In `ore.rs` we landed the wide-draw fix first as the high-value, reviewable change and treated the swap address under the cache-line argument.)

## Acceptance

- [ ] Single `bounded` helper, one draw per call, no rejection loop.
- [ ] Half-open `[0, range)` convention, documented; all call sites updated (`+1` where inclusive was intended).
- [ ] Password OOB panic and sub-alphabet skew fixed.
- [ ] `safe_rand.rs` / `bounded.rs` duplication removed.
- [ ] Tests: boundary ranges (1, powers of two, max domain) + a uniformity test; a Fisher–Yates uniformity test that would have caught the power-of-two bias.

_Found during the ore.rs v2 PRP work (rejection-sampling → fixed-draw Fisher–Yates); filing here since the root cause is shared._

coderdan:
Closes #198. Supersedes #323 (closed) and #321 (closed as duplicate of #198). Stacked on #320 (`test/safe-rand-pins-chacha-stream`); retargets to `main` once that merges.

## What

Replaces the rejection-sampling bounded draw in `vitaminc-random` with a single fixed-count **Lemire multiply-high** reduction: exactly one 64-bit draw per call, no branch on the value drawn, statistical distance from uniform at most `range / 2⁶⁴`.

API shape (revised after the review below):

- **`next_below(n)`**, new, is the primitive: uniform in `0..n`, the bound every index-shaped use wants. It is the `BoundedRng` trait's primary method, implemented for `u32` and `Protected<u32>`, and reachable as `SafeRand::next_below` without importing the trait (the inherent method is generic over the trait, so `rng.next_below(10)` and `rng.next_below(Protected::new(10))` both work).
- **`next_bounded_u32(max)`** and **`BoundedRng::next_bounded(max)`**: kept with their documented inclusive meaning, now honoured for every `max` including `u32::MAX`, and both **deprecated** in favour of `next_below(max + 1)`. Every inclusive entry point routes through one shared helper so the `max + 1` edge is handled exactly once.
- **`BoundedRng<usize>`**, previously `unimplemented!()`, is **removed** rather than implemented: an inclusive draw that takes a `usize` is the call an index-shaped use reaches for by mistake (`next_bounded(items.len())` compiles and indexes one past the end), and a single 64-bit draw is not near-uniform for ranges approaching 2⁶⁴.

Fixes all three problems from #198:

1. **Timing side-channel.** The rejection loop's iteration count depended on RNG output, so `PermutationKey` derivation time was weakly dependent on secret seed material. Now one draw, branch-free.
2. **Power-of-two bias.** The power-of-two branch returned `0..max` while the contract and the other branch were `0..=max`, so Fisher-Yates at power-of-two indices drew from the wrong range. Over 40,000 seeded keys of size 8, element 0 never once landed at position 0. Fisher-Yates now passes `i + 1` to `next_below`, and no longer spends a draw on the `i == 0` step (which can only swap an element with itself).
3. **Password out-of-bounds panic.** `next_bounded_u32(94)` could return 94 and index past `STANDARD_CHARS` (about 1 in 95 per character); alphanumeric and alpha passwords could draw one character outside their set. All three pass their set length to `next_below`.

The `STANDARD_CHARS` table read `c, e, e, f`; the `d` fix is its own `fix(password)` commit so the password changelog records it.

## Commits

1. `fix(password)`: restore `d` to `STANDARD_CHARS`
2. `fix(random)!`: the sampler, API, call sites and tests (the `BREAKING CHANGE:` footer describes the shipped API)
3. `refactor(permutation)`: drop the no-op `i == 0` Fisher–Yates draw (seeded keys unchanged)

## Breaking

- **Seeded derivation changes:** `PermutationKey::from_seed` / `from_controlled_seed` produce different keys than previous releases. The raw `next_u32`/`next_u64`/`fill_bytes` stream is unchanged. Unavoidable: preserving the old sequence means preserving the bias. README examples re-pinned.
- For a power-of-two `max`, `next_bounded_u32(max)` and `BoundedRng::next_bounded(max)` can now return `max`, as documented.
- `BoundedRng<usize>` is no longer implemented for `SafeRand`. Its body was `unimplemented!()`, so no working code breaks, but it is a compile break.
- **`BoundedRng::next_bounded` moved** to a separate, deprecated `BoundedRngInclusive` trait (review follow-up). `BoundedRng` now carries only `next_below`. Callers importing `BoundedRng` to call `next_bounded` must import `BoundedRngInclusive` instead, or migrate to `next_below(max + 1)`.
- The inclusive-form deprecations are warnings, not errors: a caller relying on the old (accidentally exclusive) behaviour at a power-of-two `max` now silently gets one extra reachable value.
- Every bounded draw now consumes a full `u64` instead of a `u32`, so the ChaCha20 stream position shifts for anyone interleaving bounded and unbounded draws off a fixed seed. Downstream fixtures pinned to a seeded `PermutationKey` need re-pinning (the README constants were).

## Tests

- `next_below` and the deprecated shim against a reference multiply-high over the same seed at powers of two, their neighbours and both ends of the `u32` range, which also pins one draw per call.
- Fixed-draw boundary vectors drive the same `below_u32`/`upto_u32` helpers every impl calls, and a seeded test pins the plain, `Protected<u32>` and inherent entry points to those helpers.
- Half-open coverage (`n` never returned), inclusive coverage (`max` reachable), zero-range panic.
- Chi-squared uniformity of the draw; chi-squared Fisher-Yates position matrix; exact replay of the key algorithm on a seed (which also pins that no extra draw is spent).
- Password: never panics from entropy, every character in its own set, every standard character reachable, set has no duplicates. Previously all three password tests were `#[ignore = "wip"]`.

## Review follow-ups (post-approval)

- `next_bounded` split off onto `BoundedRngInclusive` so it can be deleted later without breaking `BoundedRng` again.
- `Protected<u32>` zero bound is checked before `map` unwraps it; panic documented as observable on a secret and pinned.
- "Uniformly distributed" qualified with the bias bound on every public doc.
- `next_bounded_u32(u32::MAX)` driven against the reference stream.
- `j.zeroize()` restored in Fisher–Yates; explicit permutation-validity test across `N`.
- Narrow password sets tested for reachability; `ALPHA_LEN` / `ALPHANUMERIC_LEN` named.

## Not done here

`Password::into_unprotected_string` remains `unimplemented!()`; tests read the field.

https://claude.ai/code/session_01969Nqj2Fk5wzGGQ2xfEBmG
https://claude.ai/code/session_014zPfvCbCsGsFwF547vgoEt


coderdan:
## Automated code review

**Reviewer:** Claude Code, running the `/code-review` skill on model **Claude Fable 5.1** (eight parallel finder agents — line-by-line, removed-behaviour, cross-file trace, reuse, simplification, efficiency, altitude, and repo-conventions — deduplicated and then verified by the main session against head `f81411a` and the branch's git history). Posted on request by @coderdan. Session: https://claude.ai/code/session_01969Nqj2Fk5wzGGQ2xfEBmG

Findings are ordered most-severe first. Line references are to `f81411a`.

---

### 1. `3cbf3f5`'s `BREAKING CHANGE:` footer contradicts the shipped API
[`packages/random/src/bounded.rs#L15`](https://github.com/cipherstash/vitaminc/blob/f81411a666a8931121940f65fd1e734471bfd7ba/packages/random/src/bounded.rs#L15)

The footer says `next_bounded_u32` and `BoundedRng::next_bounded` "now return values in the half-open range [0, range) instead of the inclusive [0, max]", but `f81411a` restored both to inclusive `0..=max`. `main` merges by merge-commit and `cliff.toml` sets `protect_breaking_commits = true`, so this false footer lands verbatim as the input to the Breaking section — while the *true* breaking change (every `PermutationKey::from_seed` key differs) lives only in the unrendered footer body (the template renders subject lines only).

**Failure:** whoever rewrites the release-PR changelog per RELEASING.md step 4 reads the footer and tells `next_bounded(max)` callers to pass `len` instead of `len - 1` — reintroducing exactly the off-by-one this PR fixes — and anyone who persisted seeds rather than keys never learns their derived keys changed.

**Fix:** reword/squash before merge so the surviving `!` commit's subject and footer describe the actual break (seeded keys differ; power-of-two `max` is now reachable). `a15ac0a`'s body ("document the half-open contract") is stale for the same reason.

### 2. `BoundedRng<usize>` is far from uniform for large `max`
[`packages/random/src/bounded.rs#L38`](https://github.com/cipherstash/vitaminc/blob/f81411a666a8931121940f65fd1e734471bfd7ba/packages/random/src/bounded.rs#L38)

The `usize` impl feeds ranges up to 2⁶⁴−1 into the single-draw Lemire reduction, whose bias is only negligible when `range ≪ 2⁶⁴`; the trait doc promises "A uniformly distributed value in `0..=max`" with no caveat.

**Failure:** on 64-bit, `rng.next_bounded((3usize << 62) - 1)` has range 3·2⁶²; 2⁶⁴ = 1·range + 2⁶², so the low 2⁶² outputs each have two preimages and the rest one — half the outputs are exactly twice as likely as the other half, total-variation distance 1/6 (verified numerically). Even at range 2⁴⁸ the bias is ~2⁻¹⁶, nowhere near the "≤ 2⁻³²" the neighbouring `next_below` doc advertises. The only test above `u32` range is `let _ = rng.next_bounded(usize::MAX)`, which asserts nothing.

**Fix:** restrict the `usize` impl to `max <= u32::MAX` (panic / `debug_assert`), draw 128 bits for the `usize` path, or at minimum document that it is near-uniform only for `max` well below 2⁶⁴.

### 3. The undeprecated inclusive trait gains an index-shaped `usize` impl
[`packages/random/src/bounded.rs#L32`](https://github.com/cipherstash/vitaminc/blob/f81411a666a8931121940f65fd1e734471bfd7ba/packages/random/src/bounded.rs#L32)

`f81411a`'s subject says "keep the old names inclusive and deprecated", but only `next_bounded_u32` carries `#[deprecated]`. `BoundedRng::next_bounded` stays live, the README promotes it, and this PR adds a brand-new inclusive `usize` impl — the one impl that naturally receives a *length* rather than a maximum. `a15ac0a` added it explicitly for `rng.next_bounded(items.len())` under the then half-open contract; `f81411a` flipped the contract back without revisiting it.

**Failure:** `let pick = &items[rng.next_bounded(items.len())];` compiles, passes casual testing, and panics out-of-bounds on ~1/(len+1) calls. The old `unimplemented!()` at least failed loudly every time. Meanwhile `next_below` is `u32`-only, so index-shaped `usize` use needs casts and `Protected` bounds have no half-open path at all — which gives the inclusive trait a standing reason to keep attracting callers.

**Fix:** either deprecate `BoundedRng::next_bounded` with the same note and drop the new `usize` impl, or add `next_below` for `usize` / `Protected<u32>` (e.g. on the trait) and make the inclusive form a deprecated default method over it.

### 4. Fisher–Yates still runs `i == 0`, pinning a wasted draw into the seeded contract
[`packages/permutation/src/key.rs#L97`](https://github.com/cipherstash/vitaminc/blob/f81411a666a8931121940f65fd1e734471bfd7ba/packages/permutation/src/key.rs#L97)

The fold iterates `(0..N).rev()`, so the last step calls `next_below(1)` — always `0`, `swap(0, 0)` is a no-op — consuming one `u64` of ChaCha keystream per key for zero entropy. The new replay test and the re-pinned README values now bake that dead draw into `PermutationKey::from_seed`'s deterministic output.

**Why now:** this PR already changes every seeded key, which is the one free moment to fix the loop bound. Dropping the `i == 0` step yields the *identical* key for every seed (the dead draw is last and always `0`), so the README values are unaffected; doing it later is a second breaking change to deterministic keys.

**Fix:** `(1..N).rev()`, and update the replay test to match. Also: `let mut j` + `j.zeroize()` on a `usize` that has already been used as a swap index gains nothing — if intentional, the comment should say why.

### 5. The inclusive reduction is copy-pasted four times, plus a test-only fifth
[`packages/random/src/bounded.rs#L22`](https://github.com/cipherstash/vitaminc/blob/f81411a666a8931121940f65fd1e734471bfd7ba/packages/random/src/bounded.rs#L22)

`below_u64(self, u64::from(max) + 1) as u32` appears in the `u32` impl (L22), the `Protected<u32>` impl (L28), the deprecated `next_bounded_u32` (`safe_rand.rs` L53), and `TestBoundedRand`'s impl (L92), with a fifth variant (`checked_add` + `next_u64()` fallback) in the `usize` impl. The `bounded.rs` fixed-draw tests exercise the *test* copy, not `SafeRand`'s.

**Failure:** dropping the `+ 1` in `impl BoundedRng<u32> for SafeRand` keeps `min_draw_maps_to_zero` / `max_draw_maps_to_the_inclusive_bound` / `max_of_u32_max_covers_the_whole_word` green, because they route through L92.

**Fix:** one `pub(crate) fn inclusive_u64(rng, max: u64) -> u64` (owning the `checked_add` fallback) called by every impl and by `TestBoundedRand`; `Protected<u32>` becomes `max.map(|m| self.next_bounded(m))`; `next_bounded_u32` delegates to `BoundedRng::<u32>::next_bounded`. Alternatively make the primitive take a `u128` range so `usize::MAX` needs no special case and `below_u32` (a one-line wrapper with one caller) can go.

### 6. The deprecation note suggests a non-equivalent replacement
[`packages/random/src/safe_rand.rs#L49`](https://github.com/cipherstash/vitaminc/blob/f81411a666a8931121940f65fd1e734471bfd7ba/packages/random/src/safe_rand.rs#L49)

`#[deprecated(note = "use next_below(n) for a value in 0..n …")]` invites the mechanical rewrite `next_bounded_u32(max)` → `next_below(max)`, which silently drops the top value. The equivalent is `next_below(max + 1)` (with `u32::MAX` needing `next_u32()`), and the note doesn't say so.

**Fix:** e.g. `note = "inclusive 0..=max; use next_below(max + 1), or next_below(n) when you have a length n"`.

### 7. The bias bound `c4a427c` added to the trait doc was dropped by `f81411a`
[`packages/random/src/bounded.rs#L16`](https://github.com/cipherstash/vitaminc/blob/f81411a666a8931121940f65fd1e734471bfd7ba/packages/random/src/bounded.rs#L16)

`c4a427c` deliberately put the `range / 2⁶⁴` statistical-distance term on `BoundedRng::next_bounded` and `next_bounded_u32` "so protocol authors reading docs.rs" see it. `f81411a`'s doc rewrite removed it from both; it now survives only on `next_below` and the unrendered `pub(crate)` helper. Combined with finding 2, "A uniformly distributed value" with no bias term is materially misleading.

**Fix:** restore the one-sentence bound on the trait method and on `next_bounded_u32`.

### 8. The password character-table fix is bundled into a `fix(random)` commit
[`packages/password/src/lib.rs#L6`](https://github.com/cipherstash/vitaminc/blob/f81411a666a8931121940f65fd1e734471bfd7ba/packages/password/src/lib.rs#L6)

`'c','e','e','f'` → `'c','d','e','f'` (generated passwords could never contain `d` and had `e` at double weight) is a separable, user-visible `password` fix carried in `3cbf3f5` (`fix(random)!:`), against CLAUDE.md's "scope commits to the affected crate when possible". The `password` changelog will show only "Breaking: replace rejection sampling with fixed-count Lemire draw" and never mention `d`.

**Fix:** split into its own `fix(password): restore 'd' to STANDARD_CHARS` commit.

---

**Checked and not flagged:** Lemire math and the documented bounds for the `u32` API; `u32::MAX` inclusive edge; chi-squared critical values (18.47 @ 4 dof, 85.35 @ 49 dof); the fixed-draw test vectors; password table bounds 94/62/52; the raw ChaCha stream pin; all deleted tests have replacements; no other workspace caller of `next_bounded*`; clippy (incl. `wasm32-unknown-unknown`), rustdoc `-D warnings`, and tests on the three crates are clean. Efficiency: the `u64`-width draw costs ~1.4× the old average keystream per call; that is the right trade for a 2⁻⁵¹ permutation bias bound and there is no hot path in the workspace. `Password::into_unprotected_string` remaining `unimplemented!()` is noted in the PR as out of scope and was not flagged.

---
_Generated by [Claude Code](https://claude.ai/code)_

coderdan:
All eight findings from the review above are addressed at `084d7ec` (force-pushed; the branch was rebuilt as three commits). Posted by Claude Code on model Claude Fable 5.1, on request by @coderdan.

| # | Finding | Resolution |
|---|---|---|
| 1 | Stale `BREAKING CHANGE` footer | The five commits are squashed into one `fix(random)!` whose body and footer describe the shipped API (seeded keys differ; power-of-two `max` reachable; `BoundedRng<usize>` removed; raw stream unchanged). `a15ac0a`'s stale body is gone with it. |
| 2 | `BoundedRng<usize>` bias for large `max` | The `usize` impl is removed (it was `unimplemented!()` on `main`). |
| 3 | Undeprecated inclusive trait + index-shaped `usize` impl | `next_below` is now the trait's primary method for `u32` and `Protected<u32>`; `next_bounded` is `#[deprecated]` on the trait; `SafeRand::next_below` is generic over the trait so no import is needed and `Protected` bounds have a half-open path. |
| 4 | Fisher–Yates `i == 0` dead draw | Loop is `(1..N).rev()`; the `mut j` / `zeroize()` is gone; replay test updated. Seeded keys (and the README values) are unchanged, as predicted. Own `refactor(permutation)` commit. |
| 5 | Reduction copy-pasted 4× + test-only copy | One `upto_u32` helper owns the inclusive form; every impl and `next_bounded_u32` call `below_u32`/`upto_u32`; the fixed-draw tests call those same helpers (the `TestBoundedRand` trait impl is gone), plus a seeded test that pins all three entry points to them. |
| 6 | Deprecation note suggests `next_below(max)` | Note now reads: inclusive `0..=max`; use `next_below(max + 1)`, or `next_below(n)` when you have a length `n`. |
| 7 | Bias bound dropped from trait doc | Restored on the trait docs and on `next_bounded_u32`. |
| 8 | Password `'d'` fix bundled into `fix(random)` | Split out as `fix(password): restore 'd' to STANDARD_CHARS so every letter is reachable`. |

Verified locally on all three crates: `cargo fmt --check`, `clippy --all-targets -D warnings` (native and `wasm32-unknown-unknown`), tests including README doctests, `rustdoc -D warnings`. The middle commit compiles on its own.

Stacked PRs #282 and #283 have been rebased onto this head (`rebase --onto`, so only their own commits replay); #283 now carries just its one real commit instead of replaying stale copies of this PR's history.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/random/src/safe_rand.rs
```diff
@@ -109,49 +143,88 @@ mod tests {
...
         }
-        for max in [3u32, 5, 6, 100, 1000, (1 << 31) - 1] {
+    }
+
+    /// The deprecated inclusive form is `next_below(max + 1)` for every
+    /// `max`, including `u32::MAX`, where `max + 1` does not fit a `u32`.
+    #[test]
+    #[allow(deprecated)]
+    fn next_bounded_u32_is_the_inclusive_form_of_next_below() {
+        for max in BOUNDS.map(|n| n - 1) {
```
auxesis: [2 models: claude, codex] **The documented `u32::MAX` boundary is never driven through the public inclusive entry point.** `BOUNDS.map(|n| n - 1)` shifts every bound down, so the largest `max` exercised here is `u32::MAX - 1` — the exact input the new doc singles out ("for every `max` up to and including `u32::MAX`", where `max + 1` overflows a `u32`). `bounded.rs` covers `upto_u32(u32::MAX)` with two fixed draws only, not against a real stream, and not through `SafeRand::next_bounded_u32`.

Cheapest fix: append `u32::MAX` to the list this test iterates rather than adding a test. Or:

```rust
#[test]
#[allow(deprecated)]
fn next_bounded_u32_accepts_u32_max() {
    let mut safe = SafeRand::from_seed(SEED);
    let mut reference = ChaCha20Rng::from_seed(SEED);
    for _ in 0..256 {
        assert_eq!(
            safe.next_bounded_u32(u32::MAX),
            reference_below(&mut reference, u64::from(u32::MAX) + 1),
        );
    }
}
```

Expected: passes as written; fails (overflow panic or wrong value) if the inclusive path ever stops widening to `u64` before adding one.

packages/permutation/src/key.rs
```diff
@@ -93,13 +93,18 @@ where
...
-                let mut j = rng.next_bounded_u32(i as u32) as usize;
+        let key = KeyInner::<N>::generate(identity).map(|mut key| {
+            // Fisher–Yates: step `i` needs `j` uniform in `0..=i`, so the
+            // half-open bound is `i + 1`. `j == i` (no swap) must be as likely
+            // as any other choice or the permutation is not uniform. The loop
+            // stops at `i == 1`: the `i == 0` step could only draw `j == 0`
+            // and swap an element with itself, so it would spend a draw for
+            // no entropy.
+            for i in (1..N).rev() {
+                let j = rng.next_below(i as u32 + 1) as usize;
```
auxesis: [single-source: claude] **`j.zeroize()` was silently dropped in the `fold` → `for` rewrite.** The last Fisher–Yates swap index — a value derived from the key stream — is now left on the stack when `random` returns instead of being wiped.

The residual leak is small (the final iteration is `i == 1`, so the surviving `j` is one bit, and the compiler was always free to keep earlier `j`s in registers where `zeroize` never reached them). But this crate's whole thesis is wiping secret intermediates — see the `KeyInner` drop-glue note at the top of this file — and the commit subject is scoped to "stop spending a draw on the no-op step", so this reads as collateral rather than a decision.

```rust
for i in (1..N).rev() {
    let mut j = rng.next_below(i as u32 + 1) as usize;
    key.swap(i, j);
    j.zeroize();
}
```

Either restore it, or say in the comment that it was dropped deliberately as ineffective, so the next reader doesn't re-add it.

packages/random/src/bounded.rs
```diff
@@ -2,41 +2,96 @@ use crate::SafeRand;
...
+
     fn next_bounded(&mut self, max: u32) -> u32 {
-        next_bounded_u32(self, max)
+        upto_u32(self, max)
     }
 }
 
 impl BoundedRng<Protected<u32>> for SafeRand {
+    fn next_below(&mut self, n: Protected<u32>) -> Protected<u32> {
+        n.map(|n| below_u32(self, n))
```
auxesis, line 84: [single-source: claude] **A zero `Protected<u32>` bound panics, and it panics *after* the secret has been unwrapped.** `next_below(Protected::new(0))` aborts, so a `Protected` value decides control flow — the observable panic tells an attacker the secret bound was zero, which is precisely what `Protected` exists to prevent.

It is worse than a plain panic because of where it fires. `Controlled::map` is `Output::init_from_inner(f(self.risky_unwrap()))` (`packages/protected/src/controlled.rs:92`): the inner value is unwrapped *before* the closure runs, so `assert!(range > 0)` panics with the bound already out of its wrapper, and the raw value unwinds without being zeroized. Same shape applies to the inclusive impl below.

Give the `Protected` bound a fallible entry point rather than a panicking one:

```rust
fn try_next_below(&mut self, n: Protected<u32>) -> Option<Protected<u32>> {
    // check before unwrapping, so a zero bound never reaches `map`
    (n.risky_unwrap() != 0).then(|| n.map(|n| below_u32(self, n)))
}
```

or take a `Protected<NonZeroU32>` so the bound cannot be zero by construction. At minimum, document the panic on this impl — the trait-level `# Panics` note doesn't say the panic is observable on a secret.

Coverage note: this is also the only entry point whose zero-bound panic is untested. `an_empty_half_open_range_panics` hits `below_u32` and `next_below_zero_panics` hits the plain `SafeRand` path, but nothing drives a zero bound through `Protected::map`. Whatever behaviour you settle on, pin it.

packages/password/src/lib.rs
```diff
@@ -98,38 +98,89 @@ impl<const N: usize> Generatable for AlphaPassword<N> {
...
+        }
+        assert!(
+            seen.iter().all(|&s| s),
+            "some standard character never drawn"
+        );
+    }
 
-        Ok(())
+    #[test]
+    fn alphanumeric_passwords_contain_only_alphanumerics() {
```
auxesis, line 181: [single-source: claude] **The two narrow character sets are tested for membership but not for reachability.** These tests check that every drawn character is *in* the class, but never that the whole class is *reachable* — so an off-by-one that shrank the bound (`next_below(61)` / `next_below(51)`, dropping `'9'` / `'z'`) passes both. `Password<16>` gets the reachability half (`passwords_draw_from_exactly_their_character_set`); the narrower sets get only the membership half.

```rust
#[test]
fn narrow_sets_reach_their_last_character() {
    let mut rng = SafeRand::from_seed(SEED);
    let mut alnum = [false; 62];
    let mut alpha = [false; 52];
    for _ in 0..SAMPLES {
        let v: AlphaNumericPassword<16> = Generatable::random(&mut rng).expect("random");
        for c in chars(v.0) {
            alnum[STANDARD_CHARS.iter().position(|&s| s == c).expect("in set")] = true;
        }
        let v: AlphaPassword<16> = Generatable::random(&mut rng).expect("random");
        for c in chars(v.0) {
            alpha[STANDARD_CHARS.iter().position(|&s| s == c).expect("in set")] = true;
        }
    }
    assert!(alnum.iter().all(|&s| s), "alphanumeric set not fully reachable");
    assert!(alpha.iter().all(|&s| s), "alpha set not fully reachable");
}
```

Expected: passes on the current bounds (512 × 16 draws over 62 slots makes a miss astronomically unlikely at the fixed seed); fails if either bound is reduced by one, or if `AlphaPassword` is ever pointed at a slice excluding `'z'`.

packages/random/src/bounded.rs
```diff
@@ -2,41 +2,96 @@ use crate::SafeRand;
...
+/// [`BoundedRng`] for the contract, including the `range / 2⁶⁴` bias bound;
+/// that bound is why this helper is not exposed for ranges near `2⁶⁴`, where
+/// a single 64-bit draw is no longer close to uniform.
+///
+/// # Panics
+///
+/// Panics if `range` is zero.
+pub(crate) fn below_u64<R: CryptoRng>(rng: &mut R, range: u64) -> u64 {
+    assert!(range > 0, "range must be non-zero");
+    ((u128::from(rng.next_u64()) * u128::from(range)) >> 64) as u64
```
auxesis, line 115: [single-source: codex, downgraded] **Multiply-high without Lemire's threshold rejection is biased for non-divisor ranges — this is intentional, but `next_below`'s own doc line doesn't say so.**

The original filing asked for the rejection loop:

```rust
let threshold = range.wrapping_neg() % range;
loop {
    let product = u128::from(rng.next_u64()) * u128::from(range);
    if (product as u64) >= threshold {
        return (product >> 64) as u64;
    }
}
```

That should **not** land here. It directly contradicts the stated design: a data-dependent retry count leaks through timing when the generator is seeded from secret material, which is exactly the permutation-key case. And the trait docs above already state the bias bound (`n / 2^64`, ≤ 2^-32 at `n = u32::MAX`) and tell protocols needing exact uniformity to account for it. This helper's own doc repeats it.

What does survive: `BoundedRng::next_below` (line 27) and `SafeRand::next_below` open with an unqualified "A uniformly distributed value in `0..n`". A reader landing on the method via rustdoc search sees the promise without the caveat — which is how this got filed as request-changes in the first place. Suggest "A value in `0..n`, uniform to within the bias bound documented on [`BoundedRng`]" on both method docs, and consider `#[doc(alias)]`-free cross-links so the qualification travels with the claim.

packages/random/src/bounded.rs
```diff
@@ -2,41 +2,96 @@ use crate::SafeRand;
...
+
+    /// A uniformly distributed value in `0..=max`.
+    ///
+    /// Deprecated in favour of [`next_below`](BoundedRng::next_below); see
+    /// the trait docs for why. The equivalent call is `next_below(max + 1)`
+    /// (for `max == u32::MAX` that is the whole word: use
+    /// [`Rng::next_u32`](rand::Rng::next_u32)), or `next_below(n)` when the
+    /// caller has a length `n` rather than a maximum.
+    #[deprecated(
+        note = "inclusive `0..=max`; use `next_below(max + 1)`, or `next_below(n)` when you have a length `n`"
```
auxesis: [single-source: claude] **`next_bounded` is deprecated but still a *required* trait method,** so anyone implementing `BoundedRng<T>` for their own RNG is forced to write the method the crate tells them not to call. There is no migration path that ends with the deprecated method gone from their impl.

A generic default body isn't possible (`max + 1` needs `T: Add` plus the `u32::MAX` widening), but moving the inclusive form out of the required surface is:

```rust
pub trait BoundedRng<T> {
    fn next_below(&mut self, n: T) -> T;
}

#[deprecated(note = "inclusive `0..=max`; use `BoundedRng::next_below`")]
pub trait BoundedRngInclusive<T> {
    fn next_bounded(&mut self, max: T) -> T;
}
```

Removing the deprecated form later then costs a trait deletion rather than a second breaking change to `BoundedRng`. Since this PR is already `fix(random)!`, it is the cheapest time to do it.

auxesis (commented):
## Verdict: approve with follow-ups

The core change is correct and worth landing. `below_u64` is textbook Lemire multiply-high; the two edges that mattered are handled: a power-of-two `n` divides `2^64` so the reduction is *exactly* uniform there (the case the old `% max` branch got wrong), and routing the inclusive form through `u64::from(max) + 1` makes `max == u32::MAX` work instead of panicking in `next_power_of_two`. The real bug fixed — `next_bounded_u32(94)` returning `94` and indexing one past the end of `STANDARD_CHARS` — is a reachable OOB panic. The `'d'`/`'e'` table fix and the new chi-squared/reference tests are solid.

One reviewer filed the missing Lemire rejection loop as request-changes. That does not hold as a defect: the constant-draw-count design is deliberate and documented (a rejection loop's retry count would leak through timing when the generator is seeded from secret material), and the bias bound `n / 2^64` is stated on both the trait and `SafeRand::next_below`. It survives only as a doc-consistency nit, filed inline as such.

No blocking findings. Six follow-ups inline.

## Review stats

| Source | Model | Type | Raw | Kept |
|---|---|---|---|---|
| claude | claude-opus-5 | test-gap | 3 | 3 |
| claude | claude-opus-5 | rust | 3 | 3 |
| codex | gpt-5.6-terra | test-gap | 1 | 1 (merged) |
| codex | gpt-5.6-terra | rust | 1 | 1 (downgraded) |

After de-duplication: **6 distinct findings kept**, all posted inline. **1 kept finding was corroborated across models** (the `u32::MAX` boundary gap, raised independently by claude and codex). Two claude sources raised the `Protected` zero-bound issue from different angles (coverage vs. design) — merged, but counted as single-model.

Nothing was dropped as unsubstantiated; one finding was re-severitied (codex's uniformity claim, see above).

## Additional findings not posted inline

- **`packages/permutation/src/key.rs`** — nothing asserts directly that a generated key is a *valid permutation* (each of `0..N` present exactly once). It holds transitively via `key_is_fisher_yates_over_next_below` (single seed, `N == 16`) and the invert/complement round-trips, but an explicit invariant check across `N` in `[8, 16, 32, 64, 128]` would fail loudly rather than through a round-trip if the loop bounds regress.
- **`packages/random/src/bounded.rs`** — `below_u64` is `pub(crate)` and today only ever reached with `range <= 2^32`, so behaviour above `u32::MAX` is untested. Fine while the only callers are `below_u32` / `upto_u32`; worth a test the moment anything passes a genuinely 64-bit range.
- **`packages/password/src/lib.rs:81` and `:92`** — `62` and `52` are magic indices whose correctness depends entirely on `STANDARD_CHARS`' class ordering. The new `standard_chars_are_distinct_and_ordered_by_class` test does pin that ordering, so this is covered — but named consts (`ALPHA_LEN`, `ALPHANUMERIC_LEN`) shared by the generators and the test would make the coupling explicit rather than incidental.
- **Changelog / release notes** — three things deserve explicit `**Breaking:**` lines, since `fix(random)!` alone won't tell downstream what to do:
  - `impl BoundedRng<usize> for SafeRand` is removed. Its body was `unimplemented!()`, so no *working* code breaks, but it is a compile break.
  - `#[deprecated]` is a warning, not an error: a caller relying on the old (accidentally exclusive) behaviour at power-of-two `max` now silently gets an extra reachable value rather than a compile failure.
  - Every bounded draw now pulls a full `u64` instead of a `u32`, changing the ChaCha20 stream position for anyone interleaving bounded and unbounded draws off a fixed seed. That is why the two `README.md` expected-output constants changed; downstream fixtures pinned to a seeded `PermutationKey` need the same treatment.

auxesis (approved):
Thanks for this @coderdan! 

A few comments left about improvements, but not blocking.

packages/random/src/safe_rand.rs
coderdan: Done in 2052472: `next_bounded_u32_is_the_inclusive_form_of_next_below` now iterates `BOUNDS.map(|n| n - 1).into_iter().chain([u32::MAX])`, so `u32::MAX` is driven through the public inclusive entry point against the reference stream.

packages/permutation/src/key.rs
coderdan: Restored in 07682f5. `j.zeroize()` is back after the swap, with a comment saying why so it is not dropped again. Also added `key_is_a_permutation_case` (every value in `0..N` exactly once, for `N` in 8/16/32/64/128) from the review body.

packages/random/src/bounded.rs
coderdan: Done in 2052472. The zero check now runs on `risky_ref` *before* `Controlled::map`, so an unwind drops the bound still wrapped and zeroized rather than as a bare `u32` on the stack. The panic itself stays (a fallible variant would leak the same bit through `Option`), and both `BoundedRng::next_below` and the `Protected<u32>` impl now document that it is observable on a secret. Pinned by `a_zero_protected_bound_panics_before_it_is_unwrapped`, which also checks no draw is spent on the failed call.

packages/password/src/lib.rs
coderdan: Added `narrow_sets_reach_their_last_character` in adecb7d, plus named `ALPHANUMERIC_LEN` / `ALPHA_LEN` consts shared by the generators and the ordering test (the magic-index note from the review body).

packages/random/src/bounded.rs
coderdan: Done in 2052472. `BoundedRng::next_below`, `SafeRand::next_below` and the README now say "uniform to within the bias bound documented on `BoundedRng`" rather than an unqualified "uniformly distributed"; the helper docs use the same wording.

packages/random/src/bounded.rs
coderdan: Done in 2052472. `next_bounded` moved to a separate `#[deprecated] trait BoundedRngInclusive<T>`; `BoundedRng` now carries only `next_below`. Re-exported from the crate root so existing `use vitaminc_random::BoundedRng` callers get a clear "moved" signal rather than a silent break. Called out under Breaking in the PR description.

PR #282, opened 2026-08-16
feat(permutation)!: oblivious key generation via sort-by-random-key

### LINKED ISSUE #280: Constant-time oblivious permutation generation: sort-by-random-key over a Batcher network
## Summary

Split out of #198. That issue covers the bounded-draw fix (rejection sampling → fixed-count Lemire reduction), which closes the *per-draw* timing channel but keeps the Fisher–Yates structure: `key.swap(i, j)` with secret `j` is still a secret-indexed memory access, and the "constant-time note" in #198 flagged it as unresolved.

This issue replaces the key-generation algorithm entirely with a **constant-time oblivious shuffle**: sort an array of random keys through a fixed sorting network. Instruction trace, memory trace, and per-instruction latency become functions of `N` only — never of the secret permutation or the random draws. Same construction as djbsort / NTRU Prime permutation generation.

## Construction

Sort-by-random-key over a Batcher odd-even mergesort network:

1. **Pack**: build `[u64; N]` where `w[i] = (rng.next_u64() << 8) | i`. High 56 bits are the random sort key; low 8 bits carry the index as payload. One word per element halves memory traffic and the sort carries the payload for free.
2. **Sort**: run the Batcher odd-even merge network for `N` — a fixed, data-independent schedule of compare-exchange gates (543 gates at N=64), generated at compile time by a `const fn` (Knuth 5.3.4 iterative form).
3. **Collision check**: OR-accumulate `(w[j] >> 8).ct_eq(&(w[j+1] >> 8))` over adjacent pairs of the sorted array. Any collision → discard the whole batch and redraw. Whole-batch restart is required for both exact uniformity (a tie broken by packed index would bias toward identity order) and obliviousness (patching only colliding keys would leak which positions collided). Timing then depends only on *discarded* randomness — independent of the output, same argument that makes rejection sampling benign. With 56 random bits, collision probability ≈ N²/2⁵⁷ (≈ 2⁻⁴³ at N=128); restarts essentially never happen.
4. **Strip**: low bytes of the sorted array *are* the permutation (gather form).

### Constant-time compare-exchange

```rust
let swap = a.ct_gt(&b);
w[i] = u64::conditional_select(&a, &b, swap);
w[j] = u64::conditional_select(&b, &a, swap);
```

Both locations read and written unconditionally on every gate; comparison outcome absorbed into a `subtle` select mask (`cmov`/`csel`). Gate indices come from the public schedule — no memory address is ever derived from a secret.

## Scope for vitaminc

- New private module in `vitaminc-permutation` (candidate for crate extraction later).
- `const fn` schedule generation for all `IsPermutable` sizes (8, 16, 32, 64, 128 — all powers of two), gate count via a counting pass, verified against the closed form `(k² − k + 4)·2^(k−2) − 1`.
- `PermutationKey::random` switches from Fisher–Yates to the shuffle. `invert`/`permute`/`depermute` already use hardened `ConditionallySelectable` scans and are unchanged.
- Working state wrapped in `Zeroizing`.

## Verification

- [ ] Structural: unit-test the const generator against closed-form gate counts (1, 5, 19, 63, 191, 543, 1471).
- [ ] Zero-one principle: exhaustively verify the network sorts all binary inputs at N=8 (256 cases) and N=16 (65 536 cases); randomized u64 sort tests at larger N.
- [ ] Distribution: chi-squared uniformity test over positions with a seeded RNG; output always a valid permutation for every `IsPermutable` size.
- [ ] Asm inspection (follow-up): CI check that the hot path contains `cmov`/`csel` and no conditional branches on mask values.
- [ ] Empirical leakage (follow-up): `dudect-bencher` timing-distribution comparison.
- [ ] Stretch: symbolic trace verification (crux-mir / haybale-pitchfork) — Rust analogue of djbsort's angr proof.

## Non-goals (for now)

- SIMD (bitonic variant maps onto NEON/AVX-512; scalar is likely sub-microsecond already).
- Arbitrary non-power-of-two `N` (pad with `u64::MAX` sentinels if ever needed).
- Password generation (`STANDARD_CHARS[secret]` secret-indexed lookup) — single-cache-line argument applies there; separate decision.

coderdan:
Closes #280. Stacked on #281 (merge that first; this PR targets its branch). The `vitaminc-random` / `vitaminc-password` changes that came out of reviewing this PR now live in #336, a sibling on #281; this PR carries only the permutation change.

## What

Replaces Fisher–Yates in `PermutationKey::random` with a **constant-time oblivious shuffle**: sort an array of random keys through a fixed Batcher odd-even mergesort network (the djbsort / NTRU Prime construction).

Even with #281's fixed-count draw, Fisher–Yates still executes `key.swap(i, j)` where `j` derives from the secret-seeded RNG — a secret-dependent memory access pattern. `permute`/`depermute` were already hardened with `ConditionallySelectable` scans; key generation was the remaining gap flagged in #198's "constant-time note".

## Construction (`packages/permutation/src/shuffle.rs`)

1. **Pack** `w[i] = (rng.next_u64() << 8) | i` — high 56 bits random sort key, low byte carries the index.
2. **Sort** through the Batcher network: a fixed, data-independent gate schedule generated by a `const fn` (Knuth 5.3.4 iterative form), one `&'static [(u8, u8)]` per `IsPermutable` length (8–128, hung off the sealed trait so every length is compile-time guaranteed a schedule). Compare-exchange is branchless `subtle` (`ct_gt` + `conditional_select`): both slots read and written on every gate, gate indices always public.
3. **Collision check** over the random bits, accumulated into a `Choice` mask. Exactly one batch is attempted: a collision returns `RandomError::SeedRejected` and the seed must be discarded. Retrying from the same stream would make runtime reveal whether the *retained* seed collided (a predicate of the secret), and such a seed can never derive a stable key anyway. Entropy-seeded callers build a fresh `SafeRand` and call again. Whole-batch rejection keeps exact uniformity (index-broken ties bias toward identity) and obliviousness (patching would leak which positions collided). p ≈ N²/2⁵⁷, ≤ 2⁻⁴³ at N=128.
4. **Strip** low bytes → the permutation (gather form).

Working state is wrapped in `Zeroizing` and the permutation is written straight into the key's `Protected` slot. Instruction trace, memory trace, and per-instruction latency are functions of `N` only, except the single accept/reject branch on the collision predicate.

## Verification

- Gate counts match the closed form `(k² − k + 4)·2^(k−2) − 1` for n = 2…128
- **Zero-one principle** exhaustively at N=8 (256 cases) and N=16 (65 536 cases)
- Randomized u64 sort checks at N=64/128 against `sort_unstable`
- Output validity for every `IsPermutable` size incl. new N=128 invert/complement coverage
- Seed determinism; position-matrix chi-squared uniformity (from #281) now exercises this path
- Schedule sanity: every gate ordered and in bounds

Follow-ups tracked in #280: asm inspection in CI (`cmov`/`csel`, no branches on masks), `dudect` timing tests, symbolic trace verification.

## Breaking changes

- `PermutationKey::from_seed` / `from_controlled_seed` derive different keys from the same seed (algorithm change). Stacked on #281 so seed-derived keys break once, not twice.
- `PermutationKey::random` / `from_seed` can return `RandomError::SeedRejected`; `RandomError` gains that variant, so exhaustive matches need an arm.


coderdan (commented):
Security review follow-up

The initial review analysis was performed by GPT-5.6 Sol. After discussion with @coderdan, the collision-timing claim has been refined to distinguish whole-batch rejection from the current seeded architecture.

1. **Zeroize the extracted permutation buffer** (`packages/permutation/src/shuffle.rs`, around lines 138–142)

   The plain `out: [u8; N]` temporary contains the complete permutation. For `PermutationKey<128>` in an optimized x86-64 build, it can remain on the stack after being copied into protected storage; the later wipe only clears `w`. This leaves an unwiped copy even after the protected key is dropped. Wrap the extraction buffer in `Zeroizing` and transfer it using a zeroing replacement, or construct it directly in protected storage.

2. **Refined collision-handling finding** (`random_permutation` / `PermutationKey::from_seed`)

   Discarding an entire colliding batch does **not** reveal the eventual accepted permutation. With independent random batches, the rejection count is independent of the accepted permutation, so whole-batch rejection itself is sound.

   The narrower issue is that the current implementation does not discard the seed. `from_seed` initializes one deterministic RNG, and `random_permutation` advances that same RNG and sorts another batch after a collision. Runtime therefore reveals whether that retained seed caused an initial collision: a predicate of the seed, although not of the final permutation. This event is extraordinarily rare (about `2^-43` for `N = 128`) and is not plausibly exploitable in the usual setting, but it conflicts with a strict constant-time-with-respect-to-seed claim.

   The intended seed lifecycle discussed with @coderdan removes that concern:

   - generate a fresh independent seed;
   - attempt exactly one batch;
   - on collision, destroy the seed and generate another;
   - retain only a seed whose first batch succeeds.

   Under that architecture, provisioning time reveals only how many independent rejected seeds preceded the accepted seed. It reveals nothing about the retained seed beyond the already-known fact that it belongs to the collision-free set. To enforce this lifecycle, seeded derivation should return a collision error after one batch rather than retrying from the same RNG stream; the seed-generation layer can then retry with a new seed. Rejected seeds must not be reused, while caller-supplied fixed seeds should fail on collision.

   **Conclusion:** treat this as an architecture/API mismatch, not a general objection to whole-batch rejection. Once the fresh-seed rejection lifecycle is enforced, fixed-work collision handling is unnecessary for this threat model.

coderdan:
Both findings from the security review follow-up are addressed in bfa83d9:

1. **Zeroization** — `permutation_from_words` now builds the extracted permutation inside `Zeroizing<[u8; N]>` end to end; the only remaining copy is the unavoidable move into the `Protected` container (same baseline as the rest of the crate).

2. **Seed lifecycle** — `random_permutation` attempts exactly one batch. A collision returns the new `RandomError::SeedRejected` instead of drawing more words from the same RNG stream. `PermutationKey::from_seed` docs now state the contract explicitly: derivation is deterministic; `SeedRejected` means the seed can never derive a key and must be discarded; only seeds whose first derivation succeeds may be retained. Retry belongs to the seed-generation layer, where timing reveals only the count of independent, discarded seeds — nothing about the retained one.

Also from the Claude review pass: the random-word sorting test now covers every supported length (N=32 previously had no sorting-correctness coverage at all).

coderdan:
Rebased onto the re-based #281 (`fix/198-lemire-bounded-draw` at f81411a) with `rebase --onto`, so only this PR's four commits replay. One conflict, in `PermutationKey::random`: #281 now draws Fisher-Yates with `next_below(i + 1)`, this PR replaces the body with the oblivious shuffle. Resolved to the shuffle. The Fisher-Yates replay test that #281 added (`key_is_fisher_yates_over_next_below`) describes an algorithm this PR removes, so it is dropped in the first commit; #281's chi-squared position-matrix test stays and now exercises the shuffle. Workspace green locally (fmt, clippy, tests, docs), first rebased commit compiles on its own.

coderdan:
CI was red on the first Rust-gate run this branch has ever had (it predates #319). Fixed in 6a6aff5, all in `shuffle.rs` tests plus one mutants exclusion:

- **CRAP**: `schedule` was 0% covered because production only calls it inside `const` blocks. The bounds test now calls it at runtime and a `should_panic` test covers the unsupported-length arm. CRAP 56 to 7.
- **Mutants**: inner guard `i + j + k < n` was unexercised because every shipped size is a power of two; a zero-one test over every n from 2 to 16 now reaches it. The pack line uses `+` instead of `|` because `^` was an equivalent mutant. The outer merge-loop bound has a genuinely equivalent mutant (one extra no-op iteration); rewritten as `n > j + k` so it has a distinct name and excluded by that name in `.cargo/mutants.toml` with the reasoning.

Local: 100 mutants on `shuffle.rs`, 58 caught, 42 unviable, 0 missed; CRAP clean workspace-wide.

packages/permutation/src/shuffle.rs
```diff
@@ -0,0 +1,360 @@
...
+    // The payload (and the schedule's gate indices) are single bytes.
+    const { assert!(N <= 256, "permutation length must fit in u8") };
+    let mut w: Zeroizing<[u64; N]> = Zeroizing::new([0; N]);
+    for (i, slot) in w.iter_mut().enumerate() {
+        // The shift clears the low byte, so adding the index is the same as
+        // or-ing it in. `+` is used because `|` and `^` are indistinguishable
+        // here, which left the packing untestable by mutation.
+        *slot = (rng.next_u64() << 8) + i as u64;
+    }
+    permutation_from_words(&mut w).ok_or(RandomError::SeedRejected)
```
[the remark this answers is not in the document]
coderdan: Deliberate, see commit d2d21ad and the `random_permutation` doc: in-stream retry would leak whether the retained seed collided, and such a seed can never derive a stable key. PR description updated to match; #280 step 3 is superseded by this.

packages/permutation/src/shuffle.rs
```diff
@@ -0,0 +1,360 @@
...
+//!    fresh one generated (see [`random_permutation`] for the lifecycle
+//!    argument). A tie broken by the packed index would bias the output
+//!    toward identity order, and patching only the colliding keys would leak
+//!    *which* positions collided.
+//! 4. **Strip**: the low bytes of the sorted array *are* the permutation.
+//!
+//! Instruction trace, memory trace, and per-instruction latency are functions
+//! of `N` only: gate indices come from the public compile-time schedule, and
+//! the comparison outcome is absorbed into a `subtle` select mask. No memory
+//! address is ever derived from a secret value.
```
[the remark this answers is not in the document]
coderdan: Fixed in 0f820ea: the module doc now names the accept/reject branch as the one exception to the N-only trace claim.

packages/password/src/lib.rs
```diff
@@ -72,36 +72,35 @@ impl<const N: usize> AlphaPassword<N> {
     }
 }
 
+/// `N` characters drawn independently and uniformly from `set`, each with
+/// one fixed-count draw in `0..set.len()`, so the index is always in bounds
+/// and every character of the set is reachable.
```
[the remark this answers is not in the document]
coderdan: Fixed in 873335f. The doc now states the `set.len() / 2⁶⁴` bound from `BoundedRng` and the exact figure: total variation distance is `r(n − r) / (n · 2⁶⁴)` with `r = 2⁶⁴ mod n`, at most 2⁻⁶⁰ per character for the 94/62/52-character sets.

packages/random/src/generatable.rs
```diff
@@ -26,16 +26,15 @@ pub trait Generatable: Sized {
 }
 
 impl Generatable for NonZeroU16 {
+    /// Uniform over `1..=u16::MAX` in exactly one bounded draw: an offset
+    /// in `0..u16::MAX` is drawn with [`SafeRand::next_below`] and added to
+    /// one. Drawing a raw `u16` and retrying on zero would consume a
+    /// data-dependent number of words from the stream, which is the
+    /// in-stream retry that [`RandomError::SeedRejected`] exists to forbid
+    /// for a secret-seeded generator.
```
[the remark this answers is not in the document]
coderdan: Fixed in 873335f. Doc now says uniform to within `u16::MAX / 2⁶⁴` (exactly 2⁻⁶⁴ here, since `2⁶⁴ mod 65535 = 1`), and gives the real reason for the fixed-count draw: a retry on zero is equally deterministic, but its draw count leaks through timing on a secret-seeded generator. The `SeedRejected` reference is gone. Permutation generation itself packs raw `next_u64` words with no reduction, so accepted batches stay exactly uniform.

coderdan:
## Local assembly inspection of the shuffle path

Copilot's last pass deferred on the constant-time claims pending the #280 follow-ups. As an interim check I inspected the release codegen for the whole `from_seed` path (branch tip 873335f) on two targets. This is not the CI gate from #280, just a manual read of the output.

**Method.** A throwaway example with two `#[inline(never)]` wrappers calling `PermutationKey::<8>::from_seed` and `PermutationKey::<128>::from_seed`, built `--release` for `aarch64-apple-darwin` and `x86_64-apple-darwin`, dumped with `cargo asm --everything`. Everything below `from_seed` inlines into the wrapper, so the full path is visible in one function.

**Sort loop** (N = 128: loop counter runs 2942 bytes / 2 = 1471 gates, matching the closed form):
- Gate indices are loaded from the static schedule table; both slots are loaded and both stored on every gate.
- `ct_gt` compiles to the shift/or cascade followed by `bl subtle::black_box`. No compare-and-branch on the words.
- The swap is subtle's xor-mask form: `sub x8, x8, w0, uxtb` (arm64) / `neg rax` (x86-64) builds the all-ones mask, then `eor`/`and`/`eor`. There is no `csel`/`cmov` because nothing is selected, only masked. Same shape at N = 8 (fully unrolled words, looped gates).

**Collision scan.** LLVM rewrote `(a >> 8).ct_eq(&(b >> 8))` as `(a ^ b) < 256` and emitted `cmp; cset lo` (arm64) / `cmp; setb` (x86-64). Flag-to-register, no branch, accumulated through `black_box`.

**Every conditional branch in the function**, both targets:
- Bounds check and the `split_at_mut` ordering check on the gate indices (public table data).
- Loop counters (gate loop, scan loop, wipe loops).
- The single `tst w19, #0xff; b.ne` / `test bl, bl; jne` after the scan: the documented accept/reject branch.

**Confirmed baseline caveat.** On the accept path the permutation is `memcpy`'d (128 bytes) from the local key's `Protected` slot into the returned `Result`; the destination is wiped on drop, the source slot is not, because it was moved out of. On the reject path the slot is wiped. This is the "unavoidable move into the Protected container" already noted above, not a new finding.

Net: codegen matches the module doc's claim (trace is a function of N only, except the accept/reject branch) on both targets at the current tip. Still worth landing the #280 CI check so this holds across rustc upgrades.

packages/random/src/bounded.rs
```diff
@@ -21,16 +32,23 @@ use vitaminc_protected::{Controlled, Protected};
 /// below 2⁻⁵⁵ for `n ≤ 256` and still at most 2⁻³² at `n = u32::MAX`. A
 /// protocol that needs exact uniformity must account for that term.
 pub trait BoundedRng<T> {
+    /// The type of the value drawn. This is the bound's own type for `u32`
+    /// and [`Protected<u32>`], and [`Protected<u32>`] for a
+    /// [`Protected<NonZeroU32>`] bound, since `0` is a valid draw.
+    type Output;
```
freshtonic: This associated type breaks every downstream `BoundedRng` implementation, but issue #280 is scoped to oblivious permutation generation and does not require reshaping this public trait. Please split the associated-output and `Protected<NonZeroU32>` API into its own issue/PR so this security fix has a reviewable compatibility boundary.

packages/random/src/generatable.rs
```diff
@@ -26,16 +26,17 @@ pub trait Generatable: Sized {
...
-        let mut buf: [u8; 2] = [0, 0];
-        use rand::RngExt;
-        rng.fill(&mut buf);
-        if let Some(value) = NonZeroU16::new(u16::from_be_bytes(buf)) {
-            Ok(value)
-        } else {
-            // Because a 0 would be an invalid value we must try again (rejection sampling)
-            Self::random(rng)
-        }
+        let offset = rng.next_below(u32::from(u16::MAX)) as u16;
```
freshtonic: This changes fixed-seed output and stream consumption for `NonZeroU16`; the accompanying password changes are likewise independent of the oblivious shuffle requested by #280. Please move these behavior changes to a separately specified layer rather than coupling another seed-compatibility break to permutation generation.

packages/password/src/lib.rs
```diff
@@ -72,36 +72,39 @@ impl<const N: usize> AlphaPassword<N> {
     }
 }
 
+/// `N` characters drawn independently from `set`, each with one fixed-count
+/// draw in `0..set.len()`, so the index is always in bounds and every
+/// character of the set is reachable. Each draw is uniform to within the
```
freshtonic: This correctly qualifies the reduction as only approximately uniform, but `packages/password/README.md` still promises under “Unbiased Character Selection” that every character has equal probability. Please update that public security claim with the same quantified bias bound (and adjust the heading) so the README matches the implementation.

freshtonic (changes requested):
The oblivious permutation implementation itself matches amended #280: the fixed network, branchless compare-exchange, collision handling, protected output path, supported sizes, and verification are convincing. This layer also contains independently breaking RNG APIs and fixed-seed behavior changes that are not required by #280 and should be split into their own compatibility boundary. The password documentation must also be brought into line with the newly documented bounded-reduction bias.

packages/password/src/lib.rs
coderdan: It is unbiased character selection for all practical purposes.

packages/random/src/bounded.rs
coderdan: Split out to #336 (sibling on #281) together with the `Protected<NonZeroU32>` impl. This branch is rebased without commit ddacd18; `bounded.rs` is untouched here now.

packages/random/src/generatable.rs
coderdan: Moved to #336 with the password draw-loop refactor and the bias-bound docs. This branch no longer touches `generatable.rs` or `vitaminc-password`; the only `vitaminc-random` change left is the `RandomError::SeedRejected` variant the shuffle needs.

packages/password/src/lib.rs
coderdan: Fixed in #336 (0257d58): the README section is now "Near-Uniform Character Selection" and states the same `set.len() / 2⁶⁴` bound, at most 2⁻⁶⁰ per character for the shipped sets. Lives there because the `lib.rs` doc commit it matches moved with the split.

freshtonic (changes requested):
Re-review at head `873335f`: the previously requested split is not present in the PR's merge diff. Despite the updated body and replies saying the RNG/password work moved to #336, this head still includes the breaking `BoundedRng::Output`/`Protected<NonZeroU32>` changes, the `NonZeroU16` fixed-seed change, and the password refactor/docs; the password README still makes the conflicting equal-probability claim. The permutation implementation itself remains sound, but please push/rebase the advertised permutation-only branch before re-review.

freshtonic (changes requested):
Re-review at the still-current head `873335f`. The exact merge diff continues to include the breaking RNG-trait, `NonZeroU16`, and password changes that the PR body says were moved to #336, and the password README mismatch remains. The permutation work itself is ready, but the advertised permutation-only rebase has not reached this PR. Please update the remote head before requesting another re-review.

coderdan:
Pushed: head is now 1362ac8, permutation-only on top of #281. The four random/password commits and the README fix are in #336.

freshtonic (approved):
The rebase resolves the prior scope blockers: this merge diff is now limited to the oblivious permutation implementation, its documentation/dependency cleanup, and the required `SeedRejected` error. The Batcher schedule, constant-time compare-exchange, collision handling, protected output path, zeroized working state, and verification remain intact. No remaining spec or repository-standards concerns; CI is green.

PR #287, opened 2026-08-21
feat(aead): derive macros for Encrypt and Decrypt

### LINKED ISSUE #292: aead: hand-written Encrypt for a struct can leave two same-typed fields swappable in stored ciphertext — add Encrypt/Decrypt derive macros
## Background

`vitaminc-aead` is the crate in this workspace that defines how a Rust value gets encrypted. It works the way serde does, with two traits: a type implements `Encrypt` to describe how it turns itself into ciphertext, and `Decrypt` to describe how it reads itself back (`packages/aead/src/encrypt/mod.rs:9`, `packages/aead/src/decipher/mod.rs:252`). A "cipher" is the backend that does the actual AES work; the trait implementation just tells it what shape the value has — a single value, a sequence, or a map.

Shape matters here because of AAD — associated data. AAD is extra context that is mixed into the encryption and must be supplied again, identical, at decrypt time. It is not secret, and it is not stored in the ciphertext; it is how a ciphertext gets tied to *where it lives*. When a value is encrypted as a map entry, the cipher derives that entry's AAD from the map's AAD plus the entry's key (`Aad::for_map_entry`, `packages/aead/src/aad/mod.rs:79`). When a value is encrypted as a sequence element, the derived AAD carries no position — sequence order and length are deliberately not authenticated (see #252).

## Problem

Today, every struct that needs to be encrypted has its `Encrypt` and `Decrypt` implementations written by hand. That is a lot of repetitive code, and one of the choices buried in it is security-relevant: whether the struct is encoded as a map or as a sequence.

Encoding a struct as a sequence compiles cleanly, passes tests, and round-trips correctly. It is also wrong. Because element AAD carries no positional component, two fields of the same type are freely interchangeable in stored ciphertext. Concretely: a `User { name: String, email: String }` encoded as a sequence, with both values the same length, can have its two stored ciphertext blobs transposed by anyone with write access to the database. Decryption succeeds. The application now reads the email as the name and the name as the email, with no error, no warning, and nothing in the ciphertext to indicate tampering. Encoding the same struct as a map keyed by field name binds each value to its own key, and the transposed blobs fail to decrypt.

Nothing catches the sequence choice today. It is not a type error, the tests pass, and the difference only shows up under an attack nobody writes a test for. The safe encoding is available but it is not the default, and each new struct is another chance to pick the unsafe one.

There is a second, smaller obstacle. The decrypt side of a map is read through `MapAccess::next_entry::<T>()` (`packages/aead/src/decipher/mod.rs:248`), which fixes the value type `T` before the key is known. A struct has different types on different fields, so this method cannot decode one at all. Any per-field decoder needs to read the key first, decide the type from it, and then read the value.

Related open work: #215 (per-element AAD on the decrypt side is enforced by documentation rather than by the type system) and closed #269 (map keys were made part of the AAD binding in the first place, which is what makes the map shape safe).

## Proposal

1. Add `#[derive(Encrypt)]` and `#[derive(Decrypt)]` in a new `vitaminc-aead-derive` crate, re-exported from `vitaminc_aead`. This follows the existing `protected` / `protected-derive` convention in the workspace. Add it to the release group so it releases with the other crates.

2. Make the derive always emit the map shape, keyed by field name, so every encrypted field gets its own `Aad::for_map_entry` binding. Tuple struct fields use their decimal index as the key. Newtype structs stay transparent — they encrypt exactly as their inner type, so wrapping an existing type is not a wire-breaking change (this mirrors serde, and matches how `Protected<T>` and `Equatable<T>` already behave).

3. Reject enums with a compile error. A ciphertext carries no authenticated variant marker, so any encoding the macro could pick would either leak the variant in the clear or leave it forgeable. Better to refuse than to pick badly on the user's behalf.

4. Make decoding strict: a missing field, an unknown key, or a duplicate key is an error, not a default or a skip. A field whose value is skipped is a field whose AAD binding was never checked.

5. Split `MapAccess` into `next_key()` and `next_value::<T>()`, with `next_entry` defaulted over the two so existing homogeneous consumers such as `HashMap` are untouched. Reading key-first also makes decoding order-independent, which matters because entry order in stored ciphertext is not authenticated either.

6. Give field names an escape hatch. Field names become part of the ciphertext contract — renaming a field breaks compatibility with data already stored — so provide `#[aead(rename = "...")]` to pin the old wire key, and `#[aead(crate = "...")]` for use through a re-export.

7. Support cleartext fields with `#[aead(passthrough)]`. A derived struct maps onto a database row, and a row usually has columns that are not secret and must stay queryable — an id to join on, a tenant to filter by, a timestamp to sort on. Encrypting those makes the row useless to every query that does not hold the key. Passthrough is already a core capability of the cipher layer (`MapCipher::passthrough_entry`), so the derive should expose it: a passthrough field is stored as a cleartext map entry, deliberately unencrypted and unauthenticated — no tag covers it and its key is not bound — and the docs must say so plainly. The macro's anti-swap guarantee applies to encrypted fields; a passthrough field is by definition outside it, the same way the column would be if it lived beside the ciphertext instead of inside it. Two shapes are compile errors: an all-passthrough struct (nothing to seal) and `passthrough` on a transparent newtype (no map entry to hold it). On the decrypt side this needs a type-erased read (`MapAccess::next_passthrough`, mirroring `Cipher::passthrough_boxed`) with a rejecting default so a format with no passthrough representation refuses rather than invents one — and it must refuse an encrypted node, so a sealed value cannot be read back with its tag unchecked.

8. Cover the swap case explicitly in tests: two equal-length `String` fields transposed in the stored ciphertext must fail to decrypt. That is the property the whole map shape exists for, so it should be the test that fails loudest if the shape ever regresses.

coderdan:
Closes #292

Adds `#[derive(Encrypt)]` / `#[derive(Decrypt)]` for `vitaminc-aead`, and moves the `Decrypt` trait into its own module so the decrypt side mirrors the encrypt side.

## Why

Hand-writing `Encrypt`/`Decrypt` for a plain struct is boilerplate with a security-relevant detail buried in it: values have to be sealed as a **map**, because `MapCipher` binds each entry's key into the AAD its value is sealed against. Encoding a struct as a sequence instead compiles fine and round-trips fine — and leaves two same-typed fields freely interchangeable in stored ciphertext, since element AAD carries no positional component. The derive makes the safe shape the default.

## Commits

**`refactor(aead)`** — `Decrypt` and its impls move to `packages/aead/src/decrypt/`, opposite `encrypt`, leaving `decipher` as the mirror of `cipher`. Previously `Decrypt` (what a plaintext type implements) sat alongside `Decipher` (what a cipher backend implements), which the encrypt side already keeps apart. Public paths are unchanged — everything is re-exported flat from the crate root as before.

**`feat(aead)`** — the derive itself, in a new `vitaminc-aead-derive` crate re-exported from `vitaminc_aead`, following the `protected` / `protected-derive` convention. Added to the release group, so it releases with the other crates.

## Wire shape

```rust
use vitaminc_aead::{Decrypt, Encrypt};

#[derive(Encrypt, Decrypt)]
struct User {
    name: String,
    age: u32,
}
// ciphertext: Map { "name": …, "age": … }
```

- **Map keyed by field name.** Each field gets its own `Aad::for_map_entry` binding, so a stored field cannot be renamed or moved onto another key undetected. Interchangeable with the equivalent `HashMap<String, _>` ciphertext.
- **Field names are part of the ciphertext contract** — renaming a field breaks compatibility with existing data. `#[aead(rename = "...")]` pins the old wire key.
- **Newtype structs are transparent** — they encrypt and decrypt exactly as the inner type, so wrapping an existing type is not a wire-breaking change. Mirrors serde, and matches how `Protected<T>` / `Equatable<T>` already behave.
- **Tuple structs** of 2+ fields are keyed by decimal index (`"0"`, `"1"`, …), so they get the same per-field binding.
- **Unit / field-less structs** encrypt to the authenticated empty-map marker.
- **Decoding is strict**: a missing field, an unknown key, or a duplicate key is rejected rather than defaulted or skipped.
- **Enums are a compile error.** A ciphertext carries no authenticated variant discriminator, so any encoding the macro could pick would either leak the variant in the clear or leave it forgeable.
- `#[aead(crate = "::vitaminc::aead")]` on the container for use through a re-export.

## `MapAccess` gains a key/value split

`next_entry::<T>()` fixes the value type *before* the key is known, so it cannot decode a struct with heterogeneous field types at all. `MapAccess` now has:

```rust
fn next_key(&mut self) -> Result<Option<String>, Self::Error>;
fn next_value<T: Decrypt<'c> + 'c>(&mut self) -> Result<T, Self::Error>;
```

with `next_entry` **defaulted** over the two — so `HashMap`'s existing impl, and any homogeneous consumer, is untouched. `AesMapAccess` is the only implementor of the trait in the workspace.

Reading key-first also makes the decode order-independent, which matters because entry order in a stored ciphertext is not authenticated. Both misuse paths are refused rather than tolerated: a second `next_key` without consuming the pending value, and `next_value` with no pending key. A skipped value is one whose AAD binding is never verified.

## Tests

- 26 units in `vitaminc-aead-derive`: 7 shape tests — enum/union rejection, duplicate-key rejection, newtype vs named-single-field, tuple index keys, field-less shapes — plus 19 expansion tests that pin the generated code itself: per-field key binding, `rename`, tuple index keys, newtype transparency, the empty-map marker, the `'__c` bounds, the visitor's `PhantomData`, `#[aead(crate = "...")]` redirection, and the attribute-error paths. Expansions are compared as tokens, not text — the expected fragment is rendered through `quote!` too, so an assertion is written as ordinary Rust and cannot fail on `TokenStream` spacing.

  This puts the derive crate under both quality gates rather than exempting it: `decrypt::derive` goes from CRAP 90.0 at 0% covered to 9.0 at 100%, `encrypt::derive` from 56.0 to 7.0, and `cargo mutants` reports 14 caught / 0 missed. Only the two `#[proc_macro_derive]` shims in `lib.rs` stay uncovered — they run at compile time, and at CC 1 they score under the threshold anyway.
- 13 end-to-end tests in `packages/encrypt/tests/derive.rs` against real AES-256-GCM — round-trips for every shape, `HashMap` wire compatibility, rename, missing/unknown field rejection, order independence, wrong-AAD failure, and a **field-swap tamper test**: two equal-length `String` fields transposed in the stored ciphertext must fail to decrypt, which is the property the map shape exists for.
- 4 `MapAccess` contract tests in `packages/encrypt/tests/map_access.rs`.

Full workspace green; `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean.

https://claude.ai/code/session_011kjxjgxWmT4yi23ZqufSEc

### DIFF
--- a/packages/aead-derive/src/encrypt.rs
+++ b/packages/aead-derive/src/encrypt.rs
@@ -82,3 +82,181 @@ fn empty_body(krate: &syn::Path) -> TokenStream {
         #krate::MapCipher::end(#krate::Cipher::encrypt_map(__cipher, __aad))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_support::{assert_contains, assert_lacks};
+
+    fn expand(input: DeriveInput) -> String {
+        derive(input).expect("expansion should succeed").to_string()
+    }
+
+    /// Every field is sealed through `encrypt_entry`, which binds the key into
+    /// the AAD — that binding is what stops a stored field being renamed or
+    /// swapped with another of the same type.
+    #[test]
+    fn each_field_is_sealed_against_its_own_key() {
+        let out = expand(parse_quote! {
+            struct User {
+                name: String,
+                age: u32,
+            }
+        });
+
+        assert_contains(
+            &out,
+            quote!(::vitaminc_aead::Cipher::encrypt_map(__cipher, __aad)),
+        );
+        assert_contains(
+            &out,
+            quote!(::vitaminc_aead::MapCipher::encrypt_entry(
+                __map, "name", self.name
+            )),
+        );
+        assert_contains(
+            &out,
+            quote!(::vitaminc_aead::MapCipher::encrypt_entry(
+                __map, "age", self.age
+            )),
+        );
+        assert_contains(&out, quote!(::vitaminc_aead::MapCipher::end(__map)));
+    }
+
+    /// Field names are part of the wire contract, so `rename` has to reach the
+    /// key the value is bound to — not just the source-level name.
+    #[test]
+    fn rename_replaces_the_map_key() {
+        let out = expand(parse_quote! {
+            struct User {
+                #[aead(rename = "yrs")]
+                age: u32,
+            }
+        });
+
+        assert_contains(
+            &out,
+            quote!(::vitaminc_aead::MapCipher::encrypt_entry(
+                __map, "yrs", self.age
+            )),
+        );
+        assert_lacks(&out, quote!("age"));
+    }
+
+    /// A tuple struct of two or more fields keys on the decimal field index, so
+    /// its fields get the same per-key binding a named struct's do.
+    #[test]
+    fn tuple_fields_are_keyed_by_index() {
+        let out = expand(parse_quote!(
+            struct Pair(String, u32);
+        ));
+
+        assert_contains(
+            &out,
+            quote!(::vitaminc_aead::MapCipher::encrypt_entry(
+                __map, "0", self.0
+            )),
+        );
+        assert_contains(
+            &out,
+            quote!(::vitaminc_aead::MapCipher::encrypt_entry(
+                __map, "1", self.1
+            )),
+        );
+    }
+
+    /// A newtype adds nothing to the ciphertext: no map is opened, so wrapping
+    /// an existing type does not break its stored values.
+    #[test]
+    fn newtype_delegates_to_its_inner_value() {
+        let out = expand(parse_quote!(
+            struct Wrapper(String);
+        ));
+
+        assert_contains(
+            &out,
+            quote!(::vitaminc_aead::Encrypt::encrypt_with_aad(
+                self.0, __cipher, __aad
+            )),
+        );
+        assert_lacks(&out, quote!(::vitaminc_aead::Cipher::encrypt_map));
+    }
+
+    /// A field-less struct still opens and ends a map: the empty-map marker is
+    /// authenticated, so even this binds its AAD.
+    #[test]
+    fn empty_struct_still_binds_its_aad() {
+        let out = expand(parse_quote!(
+            struct Marker {}
+        ));
+
+        assert_contains(
+            &out,
+            quote!(::vitaminc_aead::MapCipher::end(
+                ::vitaminc_aead::Cipher::encrypt_map(__cipher, __aad)
+            )),
+        );
+        assert_lacks(&out, quote!(::vitaminc_aead::MapCipher::encrypt_entry));
+    }
+
+    /// Serde-style: bounding the parameter rather than each field's type lets a
+    /// field of type `Vec<T>` pick up `Vec<T>: Encrypt` from the blanket impl.
+    #[test]
+    fn type_parameters_gain_an_encrypt_bound() {
+        let out = expand(parse_quote! {
+            struct Holder<T> {
+                value: T,
+            }
+        });
+
+        assert_contains(&out, quote!(impl<T: ::vitaminc_aead::Encrypt>));
+        assert_contains(&out, quote!(for Holder<T>));
+    }
+
+    /// The `crate` attribute has to redirect *every* generated path, or the
+    /// expansion half-references a crate the caller may not depend on.
+    #[test]
+    fn crate_attribute_redirects_every_path() {
+        let out = expand(parse_quote! {
+            #[aead(crate = "::vitaminc::aead")]
+            struct User {
+                name: String,
+            }
+        });
+
+        assert_contains(&out, quote!(::vitaminc::aead::Encrypt for User));
+        assert_contains(&out, quote!(::vitaminc::aead::MapCipher::encrypt_entry));
+        assert!(
+            !out.contains("vitaminc_aead"),
+            "expansion still references the default crate path:\n{out}"
+        );
+    }
+
+    /// Attribute typos are rejected rather than ignored: a silently dropped
+    /// `#[aead(...)]` option is a setting the author believes is in effect.
+    #[test]
+    fn unknown_container_attribute_is_rejected() {
+        let err = derive(parse_quote! {
+            #[aead(bogus = "x")]
+            struct User {
+                name: String,
+            }
+        })
+        .expect_err("unknown container attribute should be rejected");
+
+        assert!(err.to_string().contains("unsupported container attribute"));
+    }
+
+    #[test]
+    fn unknown_field_attribute_is_rejected() {
+        let err = derive(parse_quote! {
+            struct User {
+                #[aead(bogus = "x")]
+                name: String,
+            }
+        })
+        .expect_err("unknown field attribute should be rejected");
+
+        assert!(err.to_string().contains("unsupported field attribute"));
+    }
+}
--- a/packages/aead-derive/src/encrypt.rs
+++ b/packages/aead-derive/src/encrypt.rs
@@ -1,0 +1,84 @@
+//! Expansion of `#[derive(Encrypt)]`.
+
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::{parse_quote, DeriveInput, Result};
+
+use crate::{
+    attrs::ContainerAttrs,
+    shape::{FieldInfo, Shape},
+};
+
+pub(crate) fn derive(input: DeriveInput) -> Result<TokenStream> {
+    let attrs = ContainerAttrs::parse(&input.attrs)?;
+    let krate = &attrs.krate;
+    let shape = Shape::parse(&input)?;
+    let name = &input.ident;
+
+    // Bound every type parameter, serde-style: a field of type `Vec<T>` then
+    // picks up `Vec<T>: Encrypt` from the blanket impl in `vitaminc_aead`.
+    let mut generics = input.generics.clone();
+    for param in generics.type_params_mut() {
+        param.bounds.push(parse_quote!(#krate::Encrypt));
+    }
+    let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
+
+    let body = match &shape {
+        Shape::Newtype(field) => newtype_body(krate, field),
+        Shape::Map(fields) => map_body(krate, fields),
+        Shape::Empty => empty_body(krate),
+    };
+
+    Ok(quote! {
+        #[automatically_derived]
+        impl #impl_generics #krate::Encrypt for #name #ty_generics #where_clause {
+            fn encrypt_with_aad<'__aead_a, __C, __A>(
+                self,
+                __cipher: __C,
+                __aad: __A,
+            ) -> ::core::result::Result<__C::Ok, __C::Error>
+            where
+                __C: #krate::Cipher,
+                __A: #krate::IntoAad<'__aead_a>,
+            {
+                #body
+            }
+        }
+    })
+}
+
+/// A newtype adds nothing to the ciphertext — it encrypts exactly as its
+/// inner value, so wrapping a type is not a wire-breaking change.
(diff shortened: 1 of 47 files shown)


packages/aead-derive/src/encrypt.rs
```diff
@@ -0,0 +1,313 @@
...
+    quote! {
+        #krate::Encrypt::encrypt_with_aad(self.#member, __cipher, __aad)
+    }
+}
+
+fn map_body(krate: &syn::Path, fields: &[FieldInfo]) -> TokenStream {
+    let entries = fields.iter().map(|field| {
+        let key = &field.key;
+        let member = &field.member;
+        if field.passthrough {
```
freshtonic, line 63: **Spec / security contract:** #292 requires the derive to “always emit the map shape, keyed by field name, so every field gets its own `Aad::for_map_entry` binding.” This branch deliberately bypasses that binding. Two same-typed passthrough values can be swapped or retargeted between keys and still produce a successfully decoded struct—the exact attack class the derive is meant to rule out. The warning is clear, but it changes the macro’s guarantee. Please remove this extension from #292, or give it a separately specified API whose unauthenticated semantics cannot be confused with strict derived `Decrypt`.

packages/aead/src/decipher/mod.rs
```diff
@@ -246,27 +284,101 @@ pub trait MapAccess<'c> {
...
+    ///
+    /// Unlike [`next_value`](MapAccess::next_value), nothing here is verified.
+    /// A passthrough entry carries no tag, and — unlike an encrypted entry's
+    /// key — its key is not bound into anything either, so a stored passthrough
+    /// entry can be edited, retargeted at another key, added, or removed with
+    /// no effect on whether the rest of the map decrypts. That independence is
+    /// the point: it is what lets a passthrough value be a plain database
+    /// column that other queries read and write on their own. Treat what comes
+    /// back as untrusted input, exactly as you would treat that column.
+    fn next_passthrough(&mut self) -> Result<Box<dyn Any + Send + 'static>, Self::Error>;
```
freshtonic: **Spec / compatibility:** the requested `MapAccess` change defaults `next_entry` over `next_key`/`next_value` so homogeneous consumers remain untouched. Making `next_passthrough` another required method forces every external `MapAccess` implementation to support unrelated functionality, including formats with no passthrough representation. If this feature remains, please provide a rejecting default or isolate it in an extension trait; preferably move it to its own scoped change.

freshtonic (changes requested):
## Standards

The implementation follows the repository’s documented conventions, keeps the derive logic well factored, documents the security tradeoffs clearly, and adds strong unit, UI, integration, mutation, and documentation coverage. I found no blocking standards violation.

## Spec

The requested derive/re-export/release integration, map and tuple shapes, transparent newtypes, enum rejection, strict missing/unknown/duplicate handling, key-first order-independent decoding, rename/crate attributes, defaulted `next_entry`, and explicit field-swap tamper test are all present.

Changes are required because the added passthrough facility conflicts with #292’s defining guarantee: the issue says the derive should always key-bind every field, whereas `#[aead(passthrough)]` deliberately emits fields with neither encryption nor authentication and permits undetectable retargeting. It also introduces a third required `MapAccess` method unrelated to the requested compatibility-preserving key/value split. Please move passthrough to a separately specified PR, or redesign it so the authenticated derive contract and existing external `MapAccess` implementations remain intact.

packages/aead-derive/src/encrypt.rs
coderdan: Invalid. Passthrough, by definition isn't checked against an AAD - its not encrypted so there is no AAD.

packages/aead/src/decipher/mod.rs
coderdan: Rejecting default is sensible but bear in mind that this code has never been released. There are no external consumers to be aware of.

packages/aead-derive/src/encrypt.rs
freshtonic, line 63: Re-review: this remains unresolved. I agree that passthrough, by definition, has no AAD; that is exactly the conflict with #292’s explicit requirement that the derive key-bind **every field**. The feature may be useful, and the warning is good, but it is a new unauthenticated exception to the issue’s central guarantee. Please separate it into its own scoped contract or make the untrusted decoding boundary distinct from ordinary derived `Decrypt`.

packages/aead/src/decipher/mod.rs
```diff
@@ -236,8 +244,43 @@ pub trait SeqAccess<'c> {
...
 pub trait MapAccess<'c> {
     /// The error type returned by [`next_entry`](MapAccess::next_entry).
-    type Error;
-    /// Returns the next decrypted `(key, value)` entry, or `None` when the map is exhausted.
+    ///
+    /// The [`From<Unspecified>`](crate::Unspecified) bound is what lets
+    /// [`next_passthrough`](MapAccess::next_passthrough) carry a default that
+    /// *refuses*: an implementation whose format has no passthrough
+    /// representation inherits the rejection rather than having to write one.
+    type Error: From<Unspecified>;
```
freshtonic, line 252: The rejecting default fixes the requirement to write `next_passthrough`, but moves the compatibility cost here: every `MapAccess` error type must now implement an unrelated conversion even when its format cannot represent passthrough. An extension trait (or separating passthrough from this PR) would preserve the requested key/value split without imposing this bound globally.

freshtonic (changes requested):
## Standards

The new commit is clean, documented, tested, and reduces boilerplate by making unsupported passthrough reads refuse safely. No standards or maintainability blocker remains; current CI is fully green.

## Spec

The required-method concern is improved but only partially resolved: `next_passthrough` now has a rejecting default, while every `MapAccess` implementation still inherits a new `Error: From<Unspecified>` requirement solely for the out-of-scope passthrough feature.

The main blocker remains unchanged. #292 specifies that the derive always uses field-key AAD so every field receives the anti-swap guarantee. `#[aead(passthrough)]` is an explicit, well-documented opt-out, but it was not part of that issue and normal derived `Decrypt` still accepts those fields without a tag or key binding. The fact that passthrough has no AAD is precisely why it should be separated from this security-focused derive contract or exposed through a distinctly untrusted API.

Changes remain requested: split passthrough into its own specification/PR, or redesign the API boundary so #292’s derived contract remains true and `MapAccess` implementors do not acquire unrelated requirements.

packages/aead-derive/src/encrypt.rs
coderdan: The issue is incorrect. You can't have any guarantees around pass-through, and pass-through is a core component of the Cipher design.

packages/aead/src/decipher/mod.rs
coderdan: No, I disagree. As I said before, we haven't released this yet. This is still shaping up the design.

packages/aead-derive/src/encrypt.rs
coderdan: Spec updated: #292 now includes passthrough (item 7) and scopes the key-binding guarantee to encrypted fields (item 2). Passthrough is a core cipher capability (`MapCipher::passthrough_entry` predates the issue) that the derive exposes, not an exception to the derive's contract. See https://github.com/cipherstash/vitaminc/issues/292#issuecomment-5419323606

packages/aead/src/decipher/mod.rs
coderdan: Spec updated: #292 now includes passthrough (item 7) and scopes the key-binding guarantee to encrypted fields (item 2). Passthrough is a core cipher capability (`MapCipher::passthrough_entry` predates the issue) that the derive exposes, not an exception to the derive's contract. See https://github.com/cipherstash/vitaminc/issues/292#issuecomment-5419323606

freshtonic (approved):
My previous rejection reasons were invalid.

PR #289, opened 2026-08-23
feat(aead): Passthrough<T> wrapper for unencrypted fields

### LINKED ISSUE #298: Typed Passthrough<T> wrapper for unencrypted fields on the dynamic path
## Problem

A custom type's `Encrypt` / `Decrypt` impl is generic over every cipher, so it cannot name a specific cipher's `Cipher::Passthrough` payload type and has no way to call `MapCipher::passthrough_entry` / `SeqCipher::passthrough_next` with a value it owns (an impl cannot add bounds the trait lacks).

The type-erased hooks (`Cipher::passthrough_boxed` / `DecipherVisitor::visit_passthrough`) close the gap at the trait level, but every struct that wants a clear field today has to hand-roll the same `Encrypt` + visitor-downcast pair. The `hlist::Passthrough` wrapper only serves the static HList encoding, not the dynamic path.

## Proposal

Add `vitaminc_aead::Passthrough<T>`: a wrapper that carries a value through the ciphertext container **without encrypting it**, as the typed front door to the existing boxed passthrough channel.

```rust
cipher
    .encrypt_map(aad)
    .encrypt_entry("id", Passthrough(self.id))?   // in the clear
    .encrypt_entry("email", self.email)?          // encrypted
    .end()
```

- `Encrypt`: hands the value to `passthrough_boxed`; AAD ignored.
- `Decrypt`: expects a passthrough node and downcasts to `T`; any other shape or payload type is `Unspecified`, never a panic.
- Documented contract: non-sensitive data only, unauthenticated.

## Context

Found while building the `stack-encrypt` target-directed layer in cipherstash-suite (cipherstash/cipherstash-suite#2147), whose `mixed_user` example currently carries a local copy of this wrapper.

coderdan:
## What

Adds `vitaminc_aead::Passthrough<T>`: a wrapper carrying a value through the ciphertext container **without encrypting it**, as the typed front door to the existing `Cipher::passthrough_boxed` / `DecipherVisitor::visit_passthrough` channel.

```rust
cipher
    .encrypt_map(aad)
    .encrypt_entry("id", Passthrough(self.id))?   // in the clear
    .encrypt_entry("email", self.email)?          // encrypted
    .end()
```

```rust
let (_, Passthrough(id)) = map.next_entry::<Passthrough<u32>>()?.ok_or(Unspecified)?;
```

## Why

A custom type's `Encrypt` / `Decrypt` impl is generic over *every* cipher, so it cannot name a specific cipher's `Cipher::Passthrough` payload type and has no way to call `MapCipher::passthrough_entry` / `SeqCipher::passthrough_next` with a value it owns (an impl cannot add bounds the trait lacks). The type-erased hooks close the gap at the trait level, but every struct that wants a clear field today has to hand-roll the same `Encrypt` + visitor-downcast pair. This wrapper is that pair, once — and the thing a future `#[derive(Encrypt)]` would emit for a `#[encrypt(passthrough)]` field.

Dynamic-path counterpart of `hlist::Passthrough` (which serves the static HList encoding only).

## Behaviour

- `Encrypt`: hands the value to `passthrough_boxed`; AAD ignored (nothing to authenticate).
- `Decrypt`: expects a passthrough node and downcasts to `T`. Any other node shape, or a payload of a different type, is `Unspecified` — never a panic.
- Contract documented prominently: **non-sensitive data only, unauthenticated** (value and, for map entries, key can be altered undetectably). Pinned by a tamper test.

## Tests

- `aead` unit tests (spy cipher / decipher): encrypt routes through the boxed channel; decrypt downcasts; foreign payload type errors.
- `encrypt` integration tests against real AES-256-GCM: mixed struct round-trip, `Vec` of mixed structs, top-level passthrough, passthrough readable in stored ciphertext, wrong type fails, tamper-undetected caveat.
- README gains a "Passing fields through in the clear" section (compiled as a doctest).

Closes #298

Found while building the `stack-encrypt` target-directed layer in cipherstash-suite (cipherstash/cipherstash-suite#2147), whose `mixed_user` example currently carries a local copy of this wrapper.

https://claude.ai/code/session_01T5iYiJc6xwzMcHDtMwCadG

### DIFF
--- a/packages/aead/README.md
+++ b/packages/aead/README.md
@@ -333,6 +333,65 @@ impl<'c> Decrypt<'c> for User {
 
 The visitor pattern keeps the cipher and the type independent: the cipher decides how the ciphertext is laid out and how AAD is enforced, while the type decides how its fields are reassembled.
 
+### Passing fields through in the clear
+
+Not every field of a record is sensitive. Wrap a field in [`Passthrough`] to carry it through the ciphertext container **unencrypted and unauthenticated** — routing or display data such as identifiers and schema versions. Because a custom type's impl is generic over every cipher, it cannot name a particular cipher's passthrough payload type; `Passthrough<T>` drives the type-erased channel for it, through the same `encrypt_entry` / `next_entry` calls as any encrypted field:
+
+```rust
+use vitaminc_aead::{
+    Cipher, Decipher, DecipherVisitor, Decrypt, Encrypt, IntoAad, MapAccess, MapCipher,
+    Passthrough, Unspecified,
+};
+
+struct User {
+    id: u32,       // stored in the clear
+    email: String, // encrypted
+}
+
+impl Encrypt for User {
+    fn encrypt_with_aad<'a, C, A>(self, cipher: C, aad: A) -> Result<C::Ok, C::Error>
+    where
+        C: Cipher,
+        A: IntoAad<'a>,
+    {
+        cipher
+            .encrypt_map(aad)
+            .encrypt_entry("id", Passthrough(self.id))?
+            .encrypt_entry("email", self.email)?
+            .end()
+    }
+}
+
+impl<'c> Decrypt<'c> for User {
+    fn decrypt_with_aad<'a, D, A>(decipher: D, aad: A) -> D::Ok<Self>
+    where
+        D: Decipher<'c>,
+        A: IntoAad<'a>,
+    {
+        struct UserVisitor;
+        impl<'c> DecipherVisitor<'c> for UserVisitor {
+            type Value = User;
+
+            fn visit_map<A: MapAccess<'c>>(self, mut map: A) -> Result<Self::Value, Unspecified> {
+                // Entries come back in encryption order; pull each with its type.
+                let (_, Passthrough(id)) = map
+                    .next_entry::<Passthrough<u32>>()
+                    .map_err(|_| Unspecified)?
+                    .ok_or(Unspecified)?;
+                let (_, email) = map
+                    .next_entry::<String>()
+                    .map_err(|_| Unspecified)?
+                    .ok_or(Unspecified)?;
+                Ok(User { id, email })
+            }
+        }
+        decipher.decrypt_map(UserVisitor, aad)
+    }
+}
+```
+
+A passthrough value — and, for map entries, its key — can be altered in the stored ciphertext without detection. Never route secret-bearing data through it; wrap it in `Protected` and encrypt it instead.
+
 ### Nonce Generation
 
 The crate provides nonce generation utilities for AEAD operations:
--- a/packages/aead/README.md
+++ b/packages/aead/README.md
@@ -335,7 +335,11 @@ The visitor pattern keeps the cipher and the type independent: the cipher decide
 
 ### Passing fields through in the clear
 
-Not every field of a record is sensitive. Wrap a field in [`Passthrough`] to carry it through the ciphertext container **unencrypted and unauthenticated** — routing or display data such as identifiers and schema versions. Because a custom type's impl is generic over every cipher, it cannot name a particular cipher's passthrough payload type; `Passthrough<T>` drives the type-erased channel for it, through the same `encrypt_entry` / `next_entry` calls as any encrypted field:
+Not every column of a table needs encrypting. A record usually has a few fields that other queries select, filter, or update without holding the key — a display name, a schema version, a plain `id` column — beside the ones that must be sealed. Wrap such a field in [`Passthrough`] and it is stored as a cleartext entry of the same ciphertext container, so the record still round-trips as one unit while that field stays an ordinary column.
+
+**Passthrough provides no security guarantees whatsoever.** The value is not encrypted and not authenticated — it, and for map entries its key, can be read, edited, added, or removed in storage and every encrypted field beside it still decrypts. Treat what comes back as untrusted input: non-sensitive, non-security-deciding data only, never a field the program then trusts for authorization, tenancy, access control, or for choosing which encrypted record to trust. The full contract is documented once, under [`#[aead(passthrough)]`](Encrypt#aeadpassthrough); `Passthrough<T>` is the hand-written equivalent of that attribute.
+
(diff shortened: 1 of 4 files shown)


packages/aead/src/passthrough.rs
```diff
@@ -0,0 +1,338 @@
...
+//! so it cannot name a specific cipher's
+//! [`Cipher::Passthrough`](crate::Cipher::Passthrough) payload type and has
+//! no way to call [`MapCipher::passthrough_entry`](crate::MapCipher::passthrough_entry)
+//! or [`SeqCipher::passthrough_next`](crate::SeqCipher::passthrough_next)
+//! with a value it owns. The type-erased hooks close that gap at the trait
+//! level; this wrapper packages them so a field can be declared "stored in
+//! the clear" with one type annotation and driven through the ordinary
+//! [`encrypt_entry`](crate::MapCipher::encrypt_entry) /
+//! [`encrypt_next`](crate::SeqCipher::encrypt_next) /
+//! [`next_entry`](crate::MapAccess::next_entry) calls like any other field:
```
coderdan: This comment explains why the pass-through named struct or named type is needed, but not why somebody would want to pass through in general.

I think that should be covered first, and then subsequently how the name type differs to calling the Cipher-related functions.

packages/aead/src/passthrough.rs
```diff
@@ -0,0 +1,370 @@
+//! [`Passthrough`], a wrapper that carries a value through the ciphertext
+//! container **without encrypting it** — the typed front door to the
+//! [`Cipher::passthrough_boxed`] / [`DecipherVisitor::visit_passthrough`]
+//! channel.
+//!
+//! # Why carry a field in the clear
+//!
+//! An encrypted record still has to work as a record. The identifier a row is
+//! fetched by, the tenant it is scoped to, the schema version that says how to
```
freshtonic: **Security documentation:** tenant scope and the identifier used to fetch a row are often authorization- or trusted-selection inputs. Because passthrough values and map keys are attacker-modifiable, recommending them here without qualification can turn the wrapper into an access-control/confused-record vulnerability. Please carry over the existing `aead-derive` rule: passthrough is for non-security-deciding data only and must never influence authorization, tenancy, access control, or selection of trusted encrypted data. Prefer an innocuous motivating example such as display metadata or a schema hint.

packages/aead/README.md
```diff
@@ -333,6 +333,65 @@ impl<'c> Decrypt<'c> for User {
...
+    where
+        D: Decipher<'c>,
+        A: IntoAad<'a>,
+    {
+        struct UserVisitor;
+        impl<'c> DecipherVisitor<'c> for UserVisitor {
+            type Value = User;
+
+            fn visit_map<A: MapAccess<'c>>(self, mut map: A) -> Result<Self::Value, Unspecified> {
+                // Entries come back in encryption order; pull each with its type.
```
freshtonic: **Incorrect map contract:** stored map order is not authenticated, so a reader cannot rely on encryption order. This visitor also ignores the returned keys and fixes the expected value type positionally; reordering the entries makes a valid record fail, and the example teaches the opposite of `MapAccess`’s key-first heterogeneous decoding contract. Please loop over `next_key`, match the key, then call `next_value::<Passthrough<u32>>()` or `next_value::<String>()`, with strict missing/unknown/duplicate handling.

freshtonic (changes requested):
## Standards

Changes requested for two security-relevant documentation problems. The new docs recommend passthrough for tenant/routing identifiers without carrying over the existing rule that unauthenticated fields must never drive authorization or selection of trusted encrypted data. The README also teaches positional heterogeneous map decoding and states that entries return in encryption order, conflicting with the crate’s documented contract that map order is unauthenticated.

There is also a minor commit-standard issue: `6343b84 doc: Updated Passthrough doc to explain motivation` should use the conventional `docs(<scope>): <why>` form required by `CLAUDE.md`. The independently repeated passthrough safety narratives have already begun to diverge; consider linking or single-sourcing a canonical contract.

## Spec

The implementation otherwise satisfies #298. `Passthrough<T>` is publicly re-exported, routes `Encrypt` through `passthrough_boxed` while ignoring AAD, routes `Decrypt` through `decrypt_any` and a fallible `Any` downcast, rejects wrong node shapes/types with `Unspecified`, and works through top-level, map, and sequence paths. Its bounds match the type-erased channel, the tamper behavior is tested, and CI is green.

Please correct the safety guidance and order-dependent example before merging.

packages/aead/src/passthrough.rs
coderdan: Agreed — fixed in ad8048d. The motivation now leads with the actual use case (a table where only some columns need encrypting: an `id` column, a display name, a schema version) and drops tenant as the motivating example. The warning section is renamed "No security guarantees whatsoever" and carries the `#[aead(passthrough)]` rule over verbatim in spirit: non-sensitive, non-security-deciding data only; never authorization, tenancy, access control, or selection of trusted encrypted data.

On single-sourcing: both the module docs and the README now link the `#[aead(passthrough)]` reference as the canonical contract rather than restating it a third time.

packages/aead/README.md
coderdan: Good catch — the example was wrong, not just the comment. Fixed in ad8048d: the visitor now loops `next_key`, matches on the key, and calls `next_value::<Passthrough<u32>>()` / `next_value::<String>()` from the arm, with duplicate, unknown, and missing keys all returning `Unspecified`. Same shape the derive emits. The "encryption order" claim is gone; the comment now states that stored entry order is unauthenticated and why key-first is required.

The `doc:` commit is also reworded to `docs(aead): ...`.

freshtonic (approved):
## Standards

The remediation resolves all prior findings. The README and module docs now state that passthrough provides no security guarantees and must never drive authorization, tenancy, access control, scope, or trusted-record selection. The example decodes maps key-first, is order-independent, and rejects duplicate, unknown, and missing keys. The safety contract points to one canonical reference, and the rewritten `docs(aead):` commits follow the repository’s commit convention. No standards or maintainability findings remain.

## Spec

`Passthrough<T>` continues to satisfy #298: encryption uses the boxed passthrough channel while ignoring AAD; decryption accepts only the passthrough visitor arm, downcasts fallibly, and returns `Unspecified` for wrong types or node shapes. It works through ordinary top-level, map, and sequence paths, and the documentation accurately describes its unauthenticated behavior.

Approved. All current CI checks are green.

PR #297, opened 2026-08-28
feat(protected): NonEmpty context wrapper checked once at construction

### LINKED ISSUE #291: Type-level non-empty AAD / PRF context, without losing the plain-string call site
## The problem

An AEAD associated data value and a PRF context can both be empty, and for some callers an empty one is a security bug rather than a degenerate case.

In `stack-encrypt`, one per-field context value becomes **both** the ciphertext's AAD and the index terms' PRF context. An empty context there collapses per-field domain separation:

- equal plaintexts in different fields produce identical index terms (cross-field equality leakage);
- every field shares one ORE/OPE key, so values become mutually order-comparable;
- ciphertexts become transplantable between fields.

Today `stack-encrypt` catches this at runtime, in two places (`Error::EmptyContext`, `TermError::EmptyContext`), on every encrypt. That is a type-system gap, and it belongs here rather than downstream: the invariant is about vitaminc's own `Aad` and `PrfContext`, and any other caller wanting it would re-implement the same check.

Note the invariant is **not** wanted universally — an empty AAD is perfectly legitimate for general AEAD use. What is needed is a way for a caller to *demand* non-emptiness in a bound.

## Sketch, and why it is not good enough yet

The `NonZero` analogy: check once at construction, let the type carry the invariant.

`vitaminc-prf` and `vitaminc-aead` are siblings over `vitaminc-protected`, so the wrapper would live in `protected` with each sibling implementing its own conversion:

```rust
// vitaminc-protected
pub struct NonEmpty<T>(T);
impl<T: AsRef<[u8]>> NonEmpty<T> {
    pub fn new(v: T) -> Option<Self>;                     // runtime, checked once
}
impl NonEmpty<&'static str> {
    pub const fn from_static(s: &'static str) -> Self;    // const-panics on "" — compile error
}

// vitaminc-aead:  impl IntoAad for NonEmpty<T>
// vitaminc-prf:   impl IntoPrfContext for NonEmpty<T>
```

plus a sealed `NonEmptyContext: IntoAad + IntoPrfContext` marker implemented only for `NonEmpty`, so a downstream crate can write the bound without defining its own type.

**The objection, and it is the reason this is filed rather than built:** it wrecks the call site.

```rust
// today
let term: EqualityTerm = "alice".encrypt_into(&cipher, "users/email").await?;

// under the sketch
let term: EqualityTerm = "alice".encrypt_into(&cipher, ctx!("users/email")).await?;
```

Passing a plain static string is worth keeping. Trading it away for the invariant is a bad deal, so the sketch above should be treated as a starting point, not the design.

## What a better answer would have to do

> **Amended (2026-09-01, after review on #297).** The original first criterion — keep `encrypt_into(&cipher, "users/email")` compiling *with the literal checked at compile time* — is unimplementable on stable Rust. A bare `&'static str` argument's value is not part of its type (`""` and `"users/email"` are the same type), so no bound, blanket impl, or const trick can inspect it at compile time; that would require `adt_const_params` (`&'static str` const generics), which is unstable. The `[u8; N]` const-generic route only reaches `b"..."` byte-string literals, not `str` literals. #297's first attempt (a `TryIntoNonEmpty` bound accepting both `NonEmpty` and raw values) preserved the call-site *syntax* by silently downgrading literals to a runtime check — which review correctly rejected as promising more than it delivers. The fallback the issue itself anticipated ("accepting that literals and dynamic strings take different paths") is therefore the design, and the ergonomic objection to the `ctx!` sketch is withdrawn: the wrapper macro is honest about what is checked and when.

- Literals take `nonempty!("users/email")`, checked at compile time — an empty one fails to compile (const-item evaluation, visible to `cargo check`).
- Dynamic values take `NonEmpty::new(value)`, checked structurally exactly once, where the context is built — not on every encrypt.
- Not force the invariant on callers who legitimately want an empty AAD: `NonEmpty` is opt-in, and plain `IntoAad`/`IntoPrfContext` call sites are untouched.

## Downstream

`stack-encrypt` deletes `require_context`, `Error::EmptyContext` and `TermError::EmptyContext` once this lands, tightens its `EncryptContext`/`DecryptContext` bounds to take `NonEmpty` directly, and its call sites become `nonempty!("users/email")`. Not urgent — the runtime checks are correct, just weaker than they should be.

Context: cipherstash-suite#2147 review, and the parent Linear issue CIP-3764.

coderdan:
## Summary

Closes #291 (as amended — see below).

An AEAD associated-data value and a PRF context can both be empty; for callers that use one value to domain-separate fields (`stack-encrypt`), an empty one collapses that separation. Downstream currently rejects empties at runtime, per encrypt, by parsing vitaminc's PAE framing and mirroring private tag constants — which already produced a real false-rejection bug (cipherstash-suite#2147).

This moves the invariant upstream and into the type.

### What's added (`vitaminc-protected`)

- **`IsEmpty`** — structural emptiness, decided on the source value *before* encoding. No parsing, no mirrored constants. `()`, `""`, `b""`, `None`, `Some("")`, `("", "")` are empty; integers never are; a composite is empty only when it contributes no caller bytes at all (`("", "email")` is not empty).
- **`NonEmpty<T>`** — checked once at construction. Two paths, stated at the call site: `nonempty!("users/email")` / `nonempty_bytes!(b"...")` at compile time (an empty literal fails to compile), `NonEmpty::new(value)` at runtime for dynamic values.
- **`EmptyError`**.

### `vitaminc-prf` / `vitaminc-aead`

- `IntoPrfContext` / `IntoAad` for `NonEmpty<T>` — byte-transparent, so opting in changes nothing on the wire.
- `IntoAad` extended to all ten fixed-width integers (raw little-endian; `u64` wire format unchanged); `IntoPrfContext` for `Cow<[u8]>` — aligning conversion coverage across the two crates.

Empty AAD remains legal for anyone who does not opt in.

### The review finding, and the #291 amendment

The first version added a `TryIntoNonEmpty` blanket bound so `encrypt_into(&cipher, "users/email")` kept compiling unchanged. Review (correctly) rejected it: a bare `&'static str` cannot be value-checked at compile time on stable Rust — `""` and `"users/email"` are the same type, and inspecting the value would need `adt_const_params` — so the bound preserved the call-site syntax by silently downgrading literals to a runtime check. That is a weaker guarantee wearing the stronger one's name.

`TryIntoNonEmpty` is gone. APIs that require the invariant take `NonEmpty<T>` directly, and the call site states which path it is on. #291's acceptance criterion has been amended accordingly (with the reasoning recorded in the issue): the `ctx!`-objection was premised on a compile-time-checked bare literal being achievable; it is not, and the honest wrapper costs 10 characters.

The mutation-gate CI change that was bundled here has been split out to #302.

### Design notes

- `nonempty!` expands to a named `const` item rather than an inline `const {}` block: inline-const panics are post-monomorphisation errors that `cargo check` — and therefore trybuild — never see. It accepts any `&'static str` constant expression (`concat!`, `env!`, a named `const`), not just literals.
- Already-encoded `Aad`/`PrfContext` deliberately do **not** implement `IsEmpty`: framing makes most encoded empties non-empty as bytes, so the check belongs before encoding (`NonEmpty<T> where T: IntoAad`).
- `TryFrom<T> for NonEmpty<T>` was left out: it conflicts with std's blanket impl since a downstream crate may legally add `From<Local> for NonEmpty<Local>`.
- `NonEmpty` is `Debug`-transparent (contexts are normally public identifiers); wrap sensitive contexts in a redacting type before proving them non-empty.

### Downstream follow-up (`stack-encrypt`)

Tighten the `EncryptContext` / `DecryptContext` supertraits to take `NonEmpty`, convert once per target, then delete `require_context`, `is_degenerate_aad`, `is_degenerate_prf_context`, `parse_pae`, the mirrored `prf_framing` constants, and both `EmptyContext` error variants. Call sites become `nonempty!("users/email")`.

## Test plan

- [x] `cargo test -p vitaminc-protected -p vitaminc-prf -p vitaminc-aead --all-features` — unit tables mirroring downstream's pinned cases, quickcheck properties, compile-fail doctests for `nonempty!("")` / `nonempty_bytes!(b"")`, doctests
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `RUSTDOCFLAGS=-D warnings cargo doc --no-deps`
- [x] `cargo +1.85.1 check` (MSRV)
- [x] `cargo fmt --all -- --check`

https://claude.ai/code/session_01PWcS13jzUo9vjU7toeFhVY


coderdan:
## Code review findings

Ranked most-severe first.

### 1. Mutants gate still passes on other did-not-complete exits — `.github/workflows/mutants.yml:215`

The gate special-cases exit 4, but other did-not-complete exits (70 internal error, 5/6 filter-diff errors, 137 OOM-kill) still pass whenever `outcomes.json` exists. `outcomes.json` is written incrementally after the baseline, so a mid-run death false-passes: `missed.txt`/`timeout.txt` empty, exit != 0 and != 4, `outcomes.json` present → gate prints "No surviving mutants in the PR diff" and goes green with most mutants untested — the same silent-skip class this commit set out to close. Since exits 2/3 are already handled by the missed/timeout file checks above, failing on any exit outside {0,2,3} closes the gap without the `outcomes.json` probe. The comment also misattributes "the tool itself errored" to exit 4 — that is exit 70, per cargo-mutants 27.1.0 `exit_code.rs`.

### 2. `IsEmpty` for `Aad`/`PrfContext` judges encoded bytes, so `NonEmpty` certifies the degenerate value it exists to exclude — `packages/aead/src/aad/mod.rs:204`, `packages/prf/src/context.rs:87`

`Some("").into_aad()` / `("", "").into_aad()` carry zero caller bytes yet pass the check because of PAE framing — violating `IsEmpty`'s own documented contract ("carries no caller-supplied bytes"). A caller building AAD from optional per-field descriptors passes it through an API bounded on `NonEmpty`/`TryIntoNonEmpty`; `NonEmpty::new(aad)` returns `Ok`, the type falsely certifies non-emptiness, and every such field shares one constant AAD: identical derived index terms and transplantable ciphertexts across fields — the exact security collapse the `NonEmpty` rustdoc names as its motivation. The diff's own test demonstrates the behavior: `assert!(!IsEmpty::is_empty(&Some("").into_aad()))`. Consider not implementing `IsEmpty` for already-encoded types (forcing the structural pre-encoding check the module doc prescribes), or recording structural emptiness at `Aad`/`PrfContext` construction.

### 3. Doc table documents a `NonEmpty<T>: IsEmpty` impl that does not exist and cannot be added — `packages/protected/src/non_empty.rs:18`

The row `| NonEmpty<T> | never, by construction |` documents an impl that does not exist — and cannot: `impl IsEmpty for NonEmpty<T>` would make the blanket `impl<T: IsEmpty> TryIntoNonEmpty for T` overlap `impl<T> TryIntoNonEmpty for NonEmpty<T>`, and a downstream impl is orphan-rejected. A user composing a proven value — `NonEmpty::new(("tenant", nonempty!("email")))` or `Some(nonempty!("x")).try_into_non_empty()` — gets E0277 (verified by compile probe), contradicting the documented contract at the exact seam downstream crates are told to build on. Drop the table row or document that `NonEmpty` composes only at the outermost position.

### 4. `IsEmpty`/`IntoAad` coverage asymmetry: integers — `packages/aead/src/aad/mod.rs:277`

`IsEmpty` covers all ten fixed-width integers but `IntoAad` covers only `u64` (`IntoPrfContext` covers all ten). `("users/email", record_id_u32).try_into_non_empty()` compiles per the `IsEmpty` table, but passing the resulting `NonEmpty` to an AEAD API bounded on `IntoAad` fails with "`IntoAad` is not implemented for `u32`"; the same context works in the PRF crate, so code moved between the two crates breaks.

### 5. Same asymmetry, other direction: `Cow<[u8]>` — `packages/prf/src/context.rs:99`

`IsEmpty` and `IntoAad` both cover `Cow<'_, [u8]>`, but vitaminc-prf has no `IntoPrfContext` impl for `Cow`, so `NonEmpty<Cow<[u8]>>` works as AAD but is rejected by the new transparent `IntoPrfContext` impl's `T: IntoPrfContext` bound — forcing an `into_owned()`/`as_ref()` workaround that changes the type and can desynchronize the AAD and PRF context representations.

### 6. New trybuild snapshot pins rustc const-eval panic internals — `packages/protected/tests/ui/negative_space/nonempty_rejects_empty_literal.stderr:8`

The snapshot pins rustc's const-eval panic backtrace rendering, including a frame inside `$RUST/core/src/panic.rs`; unlike the existing trait-resolution negative-space snapshots, const-eval diagnostics and core's panic internals change across stable releases, and all three consuming workflows use `dtolnay/rust-toolchain@stable` which auto-advances. Next stable that rewords the E0080 note turns the test, CRAP, and (via this PR's own exit-4 fix) Mutants gates red simultaneously for every open PR until someone on a matching rustc re-blesses. Asserting only the primary E0080 line avoids the recurring re-bless tax.

### 7. `nonempty!` is narrower than the API it feeds — `packages/protected/src/non_empty.rs:655`

The `$value:literal` matcher rejects consts and `concat!`/`env!` compositions that `from_static` handles fine, and `from_static` accepts only `&'static str` while the sibling `Aad`/`PrfContext` APIs are byte-oriented — `nonempty!(b"tag")` fails with a mismatched-types error pointing inside the hidden const expansion. A user with a byte-string domain tag falls back to `NonEmpty::new(b"tag".as_slice()).unwrap()` — a runtime panic path for a compile-time-checkable value. A const `from_static_bytes` plus a macro arm (or `$value:expr`, which the const item keeps compile-checked) closes the gap.

### 8. `NonEmpty<T>` derives `Debug`, printing the inner value verbatim — `packages/protected/src/non_empty.rs:536`

Debug-transparent in a crate whose wrapper convention (`Protected`, `OpaqueDebug`, `Redacted`) is to redact; safe for `NonEmpty<Protected<..>>` (delegates to the redacted impl) but not for bare `String`/`&str` contexts reaching logs via `{:?}` on an error path. Low severity since AAD/PRF contexts are usually public; deserves either a doc note that `NonEmpty` is Debug-transparent or a manual redacting impl.

### 9. rust-src hand-synced across three workflows — `.github/workflows/test.yml:61`

rust-src is now hand-added to three workflows with three cross-referencing comments; no composite action exists, and these workflows already triplicate the toolchain/LocalStack/binstall boilerplate. A future suite-affecting toolchain change applied to test.yml but not mutants.yml makes the mutants baseline fail every `packages/**` PR with an opaque "did not complete (exit 4)". A `.github/actions/setup-rust` composite with a `components` input makes it a single fact.

### 10. `IsEmpty` for `Cow` narrower than needed — `packages/protected/src/non_empty.rs:73`

Implemented only for `Cow<'_, [u8]>`; a generic `impl<T: IsEmpty + ToOwned + ?Sized> IsEmpty for Cow<'_, T>` is the same single impl and also covers `Cow<str>` (and future referents) for free.

---

Verified clean: trybuild runner picks up the new UI case automatically; Cargo wiring (thiserror workspace dep, aead/prf → protected deps) correct; inherent `is_empty` delegation has no recursion; the exit-4 gate change itself is strictly stricter than before (correct, just incomplete); `cargo fmt --check` passes.

coderdan:
All 10 findings above addressed in 88b20d0…13df8c6:

1. **Mutants gate** — now fails on any exit outside {0,2,3}; `outcomes.json` probe dropped (it is written incrementally, so it could not prove completion).
2. **`IsEmpty` for encoded types** — removed from `Aad` and `PrfContext`; non-emptiness is provable only pre-framing via `NonEmpty<T> where T: IntoAad / IntoPrfContext`.
3. **Doc table** — the `NonEmpty<T>` row is gone; docs now state it composes only at the outermost position (an impl would overlap the blanket `TryIntoNonEmpty`).
4. **Integer coverage** — `IntoAad` now covers all ten fixed-width integers (raw little-endian, u64 wire format unchanged).
5. **`Cow` coverage** — `IntoPrfContext` for `Cow<[u8]>` added, same typed encoding as `&[u8]`/`Vec<u8>`.
6. **Fragile snapshot** — const-eval trybuild case deleted; the `compile_fail` doctest plus the runtime panic-message test keep the negative space covered without pinning rustc internals.
7. **`nonempty!` breadth** — takes any `&'static str` const expression now; `NonEmpty::from_static_bytes` + `nonempty_bytes!` cover byte-string tags at compile time.
8. **Debug transparency** — documented on `NonEmpty` with guidance to wrap sensitive contexts in a redacting type first.
9. **rust-src triplication** — moot: deleting the const-eval snapshot removed the need, so rust-src is gone from all three workflows (no composite action required).
10. **Generic `Cow` emptiness** — `impl<T: IsEmpty + ToOwned + ?Sized> IsEmpty for Cow<'_, T>` replaces the `[u8]`-only impl, covering `Cow<str>` for free.

Verified locally: tests for protected/aead/prf/encrypt/aead-value green, `cargo clippy --all-targets --all-features -D warnings` clean, `cargo doc -D warnings` clean, `cargo fmt --check` clean.

packages/protected/src/non_empty.rs
```diff
@@ -0,0 +1,582 @@
...
+    /// Converts `self` into a [`NonEmpty`].
+    ///
+    /// # Errors
+    ///
+    /// Returns [`EmptyError`] if `self` is a raw value that is
+    /// [empty](IsEmpty). A `NonEmpty` input never fails.
+    fn try_into_non_empty(self) -> Result<NonEmpty<Self::Inner>, EmptyError>;
+}
+
+impl<T> TryIntoNonEmpty for T
```
freshtonic: **Spec:** this blanket implementation keeps the plain-string call compiling, but it cannot distinguish a literal from any other `&str`; both call runtime `NonEmpty::new`. Issue #291 explicitly requires the unchanged bare-literal call to be checked at compile time, while `nonempty!(...)` changes the call site in the same way as the rejected `ctx!(...)` sketch. Please implement that acceptance criterion or explicitly revise the issue before merging.

freshtonic (changes requested):
## Standards

No blocking standards or maintainability findings. The API is well documented, checks structural emptiness before encoding, avoids falsely certifying framed `Aad`/`PrfContext` values, preserves wire encodings through `NonEmpty<T>`, and keeps the invariant opt-in. Commit messages follow the repository convention, CI fails closed for incomplete mutation sweeps, and all current checks are green.

## Spec

The implementation satisfies most of #291: dynamic values can be checked once; already-proven values convert without rechecking; empty AAD remains legal for callers that do not opt in; composite/option/Cow semantics are coherent; and the explicit `nonempty!`/`nonempty_bytes!` forms genuinely fail during const evaluation.

One central requirement is still missing. #291 requires the unchanged `encrypt_into(&cipher, "users/email")` call to compile **with the literal checked at compile time**. Here a bare string satisfies the blanket `TryIntoNonEmpty` implementation and is checked by `NonEmpty::new` at runtime. Compile-time checking requires `nonempty!(...)`, recreating the wrapper-style call site the issue explicitly rejects. Please either provide the promised unchanged compile-time-checked literal path, or amend #291’s acceptance criterion and document that bare literals are runtime-checked.

The mutation-gate policy change is sensible but unrelated to the context type and would be cleaner as a separate PR.

packages/protected/src/non_empty.rs
coderdan: I was intentionally trying to keep the static string version because I like the ergonomics of it. But I guess you're right, this could be replaced with an empty string, which is annoying. That being the case, it's not even worth having the trait now, to be honest. It was only needed to avoid having an additional wrapping function or macro for the context.

coderdan: Resolved in 56132f8 by taking the second option, with the issue amended rather than quietly weakened:

- **`TryIntoNonEmpty` is deleted** — the trait's only purpose was preserving the bare-string call site, and it did so by silently downgrading literals to a runtime check. A bare `&'static str` cannot be value-checked at compile time on stable Rust (the value is not part of the type; it would need `adt_const_params`), so the criterion as written in #291 was unimplementable, not merely unimplemented.
- APIs that require the invariant now take `NonEmpty<T>` directly: `nonempty!("users/email")` for literals (compile-checked), `NonEmpty::new(value)` for dynamic values (checked once). The call site states which path it is on.
- **#291 has been amended** with the reasoning recorded in the issue body: the objection to the `ctx!` sketch was premised on a compile-time-checked bare literal being achievable, and it is not.
- The mutation-gate change is split out to #302, per the review's other note.

freshtonic (approved):
## Standards

The remediation introduces no standards or maintainability findings. Removing `TryIntoNonEmpty` eliminates the misleading convenience layer, the docs now distinguish compile-time and runtime construction precisely, commit metadata follows the repository convention, and the unrelated mutation-gate change has been removed from this PR.

## Spec

The current implementation matches amended issue #291. `nonempty!(literal)` provides a genuine compile-time check; `NonEmpty::new(dynamic)` performs one fallible structural check; APIs requiring the invariant accept `NonEmpty<T>` directly; and ordinary callers remain free to use empty AAD or PRF contexts. The sibling conversions remain byte-transparent, and structural emptiness is correctly assessed before framing.

Approved. All current CI checks are green.

PR #302, opened 2026-09-01
ci(mutants): fail the gate on any exit that means the sweep did not complete

### LINKED ISSUE #303: Mutants gate false-passes when cargo-mutants dies mid-run
## The problem

The mutants gate can go green without the sweep having run to completion.

The gate's did-not-complete check probes for `outcomes.json`:

```yaml
if [ "${{ steps.mutants.outputs.exit_code }}" != "0" ] && [ ! -f "$out/outcomes.json" ]; then
```

But cargo-mutants writes `outcomes.json` **incrementally**, starting right after the baseline build. So any mid-run death — internal tool error (exit 70), `--in-diff` filter failures (exits 5/6), an OOM kill (137) — leaves the file on disk with most mutants untested. `missed.txt` and `timeout.txt` are empty because those mutants never ran, the exit code isn't 0 but the file exists, and the gate prints "No surviving mutants in the PR diff" and passes.

That is the same silent-skip class the check was added to close: a PR can merge with its diff effectively unswept.

## The fix

Only three cargo-mutants exits mean the sweep completed: 0 (all mutants caught), 2 (missed mutants), 3 (timeouts) — and 2/3 are already enforced upstream via `missed.txt`/`timeout.txt`. Fail the gate on every other exit instead of probing for `outcomes.json`, which cannot prove completion.

Found during review of #297; fix split out per review since it is unrelated to that PR's `NonEmpty` work.

coderdan:
Closes #303.

Split out of #297 per review — unrelated to the `NonEmpty` context work.

The mutants gate treated the presence of `outcomes.json` as proof the sweep ran, but cargo-mutants writes that file incrementally after the baseline, so a mid-run death (internal error 70, filter-diff failures 5/6, an OOM kill) left it on disk and the gate went green with most mutants untested — the same silent-skip class the original check set out to close.

Only three exits mean the sweep completed: 0 (all caught), 2 (missed) and 3 (timeouts), and 2/3 are already enforced via `missed.txt`/`timeout.txt`. The gate now fails on every other exit instead of probing for the file.

https://claude.ai/code/session_01XerA7ky6KzQ4hfBLmsdvkr

### DIFF
--- a/.github/workflows/mutants.yml
+++ b/.github/workflows/mutants.yml
@@ -203,10 +203,19 @@ jobs:
             [ -s "$out/timeout.txt" ] && cat "$out/timeout.txt"
             exit 1
           fi
-          # cargo-mutants failed before producing results (e.g. baseline tests
-          # failed) — surface that as a gate failure rather than a false pass.
-          if [ "${{ steps.mutants.outputs.exit_code }}" != "0" ] && [ ! -f "$out/outcomes.json" ]; then
-            echo "::error::cargo-mutants did not complete (exit ${{ steps.mutants.outputs.exit_code }}) — check the run logs."
-            exit 1
-          fi
+          # Only three exits mean the sweep actually completed: 0 (all mutants
+          # caught), 2 (missed mutants) and 3 (timeouts) — and 2/3 are already
+          # enforced via missed.txt/timeout.txt above. Everything else means it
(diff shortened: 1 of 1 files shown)


auxesis (approved):
@coderdan thanks for this! Excellent work.

PR #305, opened 2026-09-01
refactor(go): extract the shared vcffi transport codec module

### LINKED ISSUE #304: go: extract the shared FFI transport codec into a vcffi module
## Problem

The Go transport codec — the wire encoding that carries value and ciphertext trees across the wasm boundary, plus the `Encryptable`/`Encoder` extension point — is an unexported detail of `bindings/go/vcencrypt`. The stack-encrypt Go SDK (cipherstash-suite `docs/plans/stack-encrypt-go-bindings.md`, Phase 4) needs the same codec, and its plan explicitly forbids forking it: one codec, one hostile-input test suite, one fuzz corpus.

Two additional constraints:

- The ciphertext decoder hard-codes the `vcvalue` sealed-leaf types. A stack-encrypt sealed leaf is **not** decryptable by vitaminc-encrypt and must stay a distinct Go type — it must never scan or marshal where a `vcvalue.Sealed` belongs — so the leaf materialization has to be parameterised per binding.
- `vcvalue` must stay model-only (no wire format), and codec consumers must not inherit wazero.

## Proposal

A third Go module, `bindings/go/vcffi`, sitting between `vcvalue` (model) and the bindings:

- Move the transport codec + `Encryptable`/`Encoder` out of `vcencrypt`; export the entry points (`Marshal`, `Unmarshal`, `MarshalCipherText`, `UnmarshalCipherText`, `ErrMalformed`) now that the module boundary is the visibility line.
- Parameterise the ciphertext codec over a `LeafSet` (classify + construct), with a `VCValueLeaves` set for the vcvalue types.
- `vcencrypt` keeps its public surface via type aliases and internal shims.
- Codec/fuzz tests move with the code; CI gains vcffi lanes including the 32-bit GOARCH run where the u32-bound guards are load-bearing.

coderdan:
Closes #304.

## What

The Go transport codec — the wire encoding carrying value and ciphertext trees across the wasm boundary, plus the `Encryptable`/`Encoder` extension point — moves out of `vcencrypt` into a third module, `bindings/go/vcffi`, so the stack-encrypt Go SDK can consume the same implementation instead of forking it: one codec, one hostile-input suite, one fuzz corpus.

## Design

- **`LeafSet` parameterisation.** The ciphertext codec's structural nodes (`[]any`, `map[string]any`, `vcvalue.Plain`) are common to every binding, but the sealed-leaf marker types are deliberately not: each binding supplies a `LeafSet{Classify, Make}` mapping its own leaf types onto the framing. A stack-encrypt leaf is not decryptable by vitaminc-encrypt, and keeping the Go types distinct means one can never scan or marshal where the other belongs. `VCValueLeaves` is the set for the vcvalue model; a partial `LeafSet` is rejected up front rather than surfacing as a nil-function panic mid-tree. The wire bytes are identical across `LeafSet`s — a set changes materialization, not the encoding (pinned by test).
- **Exported entry points.** `Marshal`, `Unmarshal`, `MarshalCipherText`, `UnmarshalCipherText` and `ErrMalformed` are public now that the module boundary is the visibility line; error prefixes follow the package name (`vcffi:`).
- **`vcencrypt` API unchanged.** Its public surface survives via type aliases (`Encryptable`, `Encoder`, `SeqEncoder`, `MapEncoder`, `Encode`) and thin internal shims — a type implementing `vcencrypt.Encryptable` *is* a `vcffi.Encryptable`. The test-only export bridge is gone; the cross-language fixture test calls vcffi directly. *Error text is not part of that guarantee:* messages surfacing through `vcencrypt.Encrypt`/`Decrypt` now read `vcffi: ...` rather than `vcencrypt: ...`, and encoder depth exhaustion moved from a fresh `errors.New` to the `ErrTooDeep` sentinel. A consumer matching on error strings breaks; `errors.Is` is now the supported test. Pre-1.0, so no shim.
- **Dependencies.** `vcffi` depends only on `vcvalue` (local replace, no go.sum). `vcvalue` stays model-only; codec consumers do not inherit wazero.

## Tests / CI

- Codec, fuzz and 386-guard tests move with the code (git rename detection: 83–98% similarity on moved files).
- New coverage for the seam: a binding-specific leaf type round-trips as itself, the same buffer decodes to vcvalue types under `VCValueLeaves`, a foreign leaf is rejected as bare plaintext, and partial `LeafSet`s are refused.
- `go.yml` gains vcffi vet/test lanes, including the linux/386 run where the u32-count guards are load-bearing.
- Verified locally: vet + tests green for all three modules, the sqlite `userrecord` example runs end-to-end, gofmt clean, linux/386 cross-build ok.

## Docs

READMEs and package docs rewritten for the three-module layout. This also fixes the top-level `bindings/go/README.md`, which claimed the codec and encoder lived in `vcvalue` — they never did.

https://claude.ai/code/session_01P5YHK3w6Kj9ajTnmkaXCHW


bindings/go/vcffi/ciphertext.go
```diff
@@ -0,0 +1,316 @@
...
+		if err != nil {
+			return nil, err
+		}
+		s, err := r.take(n)
+		if err != nil {
+			return nil, err
+		}
+		leaf := make([]byte, len(s))
+		copy(leaf, s)
+		switch tag {
```
coderdan: **correctness — confirmed:** `decodeCipherText` trusts `LeafSet.Make` unconditionally: `Make` has no error return and its result is never nil-checked, so a `LeafSet` that does not cover all four leaf kinds silently injects untyped nil nodes into the decoded tree instead of erroring.

A binding defining a partial `LeafSet` exactly like the in-repo test fixture `otherLeaves` (returns nil for `LeafNone`/`LeafEmptySeq`/`LeafEmptyMap`) makes `UnmarshalCipherText` on wire bytes carrying tag `ctNone` (0x02) return `nil, nil` — a nil node with no error. Downstream type-switches hit nil, and re-encoding the accepted decode fails with "is not a ciphertext node". `check()`'s own comment claims partial LeafSets are rejected up front, but per-kind partiality is not.

**Fix:** treat a nil `Make` result as `ErrMalformed` (or give `Make` an error return).

coderdan (commented):
Code review findings (10, ranked by severity — correctness first). Each posted inline.

---

**bindings/go/vcvalue/go.mod:1** (file not in this diff, so noted here): **docs:** this header comment still claims the module contains the "transport codec" — the very thing this diff moved to vcffi and that the rewritten `vcvalue/doc.go` and README now explicitly say is NOT here. A future stack-encrypt SDK author reading go.mod imports vcvalue expecting the codec, finds none, and forks the wire encoding — the exact outcome vcffi's own docs say must never happen.

**Fix:** drop "transport codec" from the comment.

bindings/go/vcffi/ciphertext.go
```diff
@@ -0,0 +1,316 @@
...
+func decodeCipherText(leaves LeafSet, r *reader, depth int) (any, error) {
+	if depth > maxDepth {
+		return nil, ErrMalformed
+	}
+	tag, err := r.byteTag()
+	if err != nil {
+		return nil, err
+	}
+	switch tag {
+	case ctSingle, ctNone, ctEmptySeq, ctEmptyMap:
```
coderdan: **correctness:** the LeafKind↔wire-tag mapping is spelled three times (`leafTag`, this outer case list, the inner tag→kind switch below), and the inner switch's `default` arm maps any unmatched tag to `LeafSingle` — so a future fifth leaf tag added everywhere except the inner switch decodes silently as the wrong kind (through `VCValueLeaves.Make` that becomes `vcvalue.Sealed`, surfacing only as a confusing AEAD failure at decrypt).

The encode side already errors on unknown kinds; decode should match. A shared table or `leafKind(tag)` inverse helper makes the misdecode impossible and collapses the four `Make` calls to one.

bindings/go/vcffi/ciphertext.go
```diff
@@ -0,0 +1,316 @@
...
+//
+// [VCValueLeaves] is the set for the vcvalue model types.
+type LeafSet struct {
+	Classify func(v any) (LeafKind, []byte, bool)
+	Make     func(kind LeafKind, bytes []byte) any
+}
+
+// VCValueLeaves is the [LeafSet] of the vcvalue model: vcvalue.Sealed,
+// vcvalue.SealedNone, vcvalue.SealedEmptySeq and vcvalue.SealedEmptyMap.
+var VCValueLeaves = LeafSet{
```
coderdan: **api-design:** `VCValueLeaves` is an exported mutable package-level var read on every call from vcencrypt's codec path; any importer can reassign its `Classify`/`Make` fields at runtime and silently change how every `vcencrypt.Client` in the process frames sealed leaves (e.g. stripping the Sealed classification so leaves hit the bare-plaintext arm). The corruption is process-wide and silent.

**Fix:** a `func VCValueLeaves() LeafSet` accessor (or unexported var) closes it — and also stops the vcvalue-specific set being the path-of-least-resistance default inside the binding-neutral module.

bindings/go/vcffi/ciphertext.go
```diff
@@ -0,0 +1,316 @@
...
+		return nil, ErrMalformed
+	}
+	return ct, nil
+}
+
+func encodeCipherText(leaves LeafSet, out []byte, v any, depth int) ([]byte, error) {
+	if depth > maxDepth {
+		return nil, errors.New("vcffi: ciphertext is nested too deeply")
+	}
+	if kind, bytes, ok := leaves.Classify(v); ok {
```
coderdan: **correctness:** `encodeCipherText` consults `leaves.Classify` before the structural type switch, so a `LeafSet` whose `Classify` over-claims `[]any`, `map[string]any`, or `vcvalue.Plain` silently encodes an entire subtree as one leaf chunk with no error — the old closed type switch made this structurally impossible. The `LeafSet` doc only says unclaimed values "fall through"; nothing forbids or checks claiming the codec's own structural types.

**Fix:** doc constraint plus a cheap pre-check (structural types never offered to `Classify`).

bindings/go/vcffi/ciphertext.go
```diff
@@ -0,0 +1,316 @@
...
+	},
+	Make: func(kind LeafKind, bytes []byte) any {
+		switch kind {
+		case LeafNone:
+			return vcvalue.SealedNone(bytes)
+		case LeafEmptySeq:
+			return vcvalue.SealedEmptySeq(bytes)
+		case LeafEmptyMap:
+			return vcvalue.SealedEmptyMap(bytes)
+		default:
```
coderdan, line 123: **correctness:** `VCValueLeaves.Make` materializes ANY unrecognized `LeafKind` as `vcvalue.Sealed` via this default arm (`LeafSingle` has no explicit case), while its encode-side sibling `leafTag` errors on unknown kinds. `LeafKind` is an open byte type and `Make` is externally callable: `VCValueLeaves.Make(LeafKind(9), b)` returns `vcvalue.Sealed(b)` today; a future fifth kind would be silently minted as ordinary Sealed values, surfacing as a late AEAD authentication error far from the cause.

**Fix:** mirror `leafTag` — explicit `LeafSingle` case, default returns nil (which the decode-side nil guard then turns into an error).

bindings/go/vcffi/encoder.go
```diff
@@ -456,7 +456,7 @@ func marshal(v any) ([]byte, error) {
 // declaration order.
 //
 // Encode returns the first error recorded while writing enc, so it composes:
-// an Encryptable can delegate a sub-value with vcvalue.Encode(m.Field(k), x).
+// an Encryptable can delegate a sub-value with vcffi.Encode(m.Field(k), x).
 func Encode(enc Encoder, v any) error {
 	encodeAny(enc, v)
 	return enc.st.err
```
coderdan, line 465: **api-design:** every method on the newly public `Encoder` nil-panics on a zero-value `Encoder` — `Encoder{}` compiles outside the package and dereferences nil `enc.st` — and no doc states the zero value is unusable. The docs bless `Encoder` being "passed and stored by value", so a zero-valued struct field holding one is a realistic accident: `var e vcffi.Encoder; e.Null()` panics with an unattributable nil deref.

**Fix (proportionate):** one doc sentence stating a usable `Encoder` comes only from `Marshal`/`Elem`/`Field`/`Passthrough`.

bindings/go/vcffi/encoder.go
```diff
@@ -217,7 +217,7 @@ func (e Encoder) Bytes(b []byte) {
 func (e Encoder) Passthrough() Encoder {
 	if e.st.err == nil {
 		if e.depth+1 > maxDepth {
-			e.st.fail(errors.New("vcencrypt: value is nested too deeply"))
+			e.st.fail(errors.New("vcffi: value is nested too deeply"))
```
coderdan: **consistency — confirmed:** depth-exceeded errors are constructed inline six times (four in encoder.go, two in ciphertext.go) as bare `errors.New`, inconsistent with the `ErrMalformed` sentinel convention this same diff establishes by exporting it — callers of the newly public module can `errors.Is`-match malformed input but not depth exhaustion, and the six independent string literals drift on any rewording.

**Fix:** add an `ErrTooDeep` sentinel next to `ErrMalformed` in transport.go.

bindings/go/README.md
```diff
@@ -7,45 +7,56 @@ no shared libraries, no cross-compilation matrix.
...
+  SDK imports for the model.
   Module path: `github.com/cipherstash/vitaminc/bindings/go/vcvalue`.
 
+- **[`vcffi/`](vcffi)** — the shared FFI transport codec: the wire encoding
+  that carries value and ciphertext trees across a wasm boundary, plus the
+  `Encryptable`/`Encoder` extension point. Depends only on `vcvalue`. One
+  codec for every binding that owns such a boundary — the encoding must
+  never fork per binding. Its ciphertext decoder is parameterised over a
+  `LeafSet`, so each binding materializes its **own** sealed-leaf types (a
+  stack-encrypt leaf is not a `vcvalue.Sealed`).
```
coderdan, line 27: **docs:** the module-boundary docs pull in opposite directions on `vcvalue.Sealed`: this README says "a stack-encrypt leaf is not a vcvalue.Sealed", while the vcvalue description above — plus vcvalue/README.md, doc.go, and the CI comment — headline "Sealed and the marker leaves" as the model a stack-encrypt Go SDK imports. An author reading vcvalue's docs alone concludes their sealed values should be `vcvalue.Sealed`, silently defeating the LeafSet type-separation whose whole point is that they must not be.

**Fix:** one clarifying sentence in vcvalue/README.md and doc.go (the sealed family here is vitaminc-encrypt's materialization; other bindings define their own via `vcffi.LeafSet`).

bindings/go/vcencrypt/codec.go
```diff
@@ -0,0 +1,49 @@
...
+type MapEncoder = vcffi.MapEncoder
+
+// Encode records v into enc following encoding/json-style rules, consulting
+// [Encryptable] first — see [vcffi.Encode]. It composes: an Encryptable can
+// delegate a sub-value with vcencrypt.Encode(m.Field(k), x).
+func Encode(enc Encoder, v any) error {
+	return vcffi.Encode(enc, v)
+}
+
+func marshal(v any) ([]byte, error) {
```
coderdan: **simplification:** the unexported `marshal`/`unmarshal` wrappers are pure passthroughs to `vcffi.Marshal`/`vcffi.Unmarshal` binding nothing (unlike their siblings, which bind `VCValueLeaves`), and the package already uses both spellings — vcencrypt_test.go calls vcffi directly while client.go uses the wrappers. A future change added to the wrapper but not to a direct call site silently forks behavior between the client path and the test path.

**Fix:** inline them into client.go, or use one consistent spelling.

bindings/go/vcffi/ciphertext.go
coderdan: Resolved in faee16b: `decodeCipherText` now treats a nil `Make` result as an error naming the kind (`vcffi: LeafSet.Make returned no value for LeafKind %d`) — a wiring mistake in the binding, attributable, never a nil node. Pinned by `TestCipherTextMakeMustMaterializeDecodedKind` decoding a `ctNone` under the partial `otherLeaves` fixture.

bindings/go/vcffi/ciphertext.go
coderdan: Resolved in faee16b: single `leafTags` table is now the source of truth, with `leafTag` (encode) and `leafKind` (decode) as its two directions. The tripled case lists and the misdecoding `default` arm are gone; the four `Make` calls collapsed to one.

bindings/go/vcffi/ciphertext.go
coderdan: Resolved in faee16b: `VCValueLeaves` is now an accessor func returning the (unexported) set by value — importers can no longer reassign `Classify`/`Make` out from under every client in the process. All call sites updated.

bindings/go/vcffi/ciphertext.go
coderdan: Resolved in faee16b: `encodeCipherText` now runs the structural arms first and only offers non-structural values to `Classify` (moved into the type switch's default arm), so a LeafSet structurally cannot claim `[]any`/`map[string]any`/`vcvalue.Plain`. The `LeafSet` doc states the contract.

bindings/go/vcffi/ciphertext.go
coderdan: Resolved in faee16b: `Make` now mirrors `leafTag` — explicit `LeafSingle` case, `default` returns nil, which the decoder's new nil guard turns into an error instead of minting `Sealed`. Pinned by `TestCipherTextMakeMustMaterializeDecodedKind` (`Make(LeafKind(9), ...)` must return nil).

bindings/go/vcffi/encoder.go
coderdan: Resolved in faee16b: `Encoder`'s doc now states the zero value is unusable (methods panic) and that usable Encoders come only from `EncryptValue`, `Passthrough`, `Elem` and `Field`.

bindings/go/vcffi/encoder.go
coderdan: Resolved in faee16b: added `ErrTooDeep` sentinel next to `ErrMalformed` in transport.go; all six encode-side inline constructions replaced. Decoders deliberately keep folding depth into `ErrMalformed` (hostile input stays unspecific), which the sentinel's doc spells out. Depth test now asserts `errors.Is(err, ErrTooDeep)`.

bindings/go/README.md
coderdan: Resolved in faee16b: vcvalue/doc.go and vcvalue/README.md now state the `Sealed` family is vitaminc-encrypt's materialization of the model, and that other bindings define their own distinct leaf types via `vcffi.LeafSet` — a stack-encrypt leaf is never a `vcvalue.Sealed`.

bindings/go/vcencrypt/codec.go
coderdan: Resolved in faee16b: the pure-passthrough `marshal`/`unmarshal` wrappers are gone; client.go and the tests call `vcffi.Marshal`/`vcffi.Unmarshal` directly, one spelling per operation. The ciphertext pair keeps its wrappers (they bind the vcvalue leaves).

coderdan:
The review-body finding (bindings/go/vcvalue/go.mod:1 — header comment still claiming the transport codec lives in vcvalue) is also resolved in faee16b: the comment now says the codec lives in the sibling vcffi module. All 10 findings addressed in faee16b; e13e24e additionally adds the previously missing gofmt gate to CI.

bindings/go/README.md
```diff
@@ -7,45 +7,56 @@ no shared libraries, no cross-compilation matrix.
...
 frozen data model, which real applications (via stack-encrypt) will depend on
 directly. `vcencrypt` is a reference tool that proves the model works
 end-to-end through wasm. A hard module boundary is what guarantees the
 wazero/crypto dependencies of the reference tool can never leak into the layer
-applications ship. This is the sharpened form of the earlier "two layers"
-rationale: one of the two is not a binding, so it gets its own module.
+applications ship. `vcffi` sits between them for the same reason from the
+other side: the codec must be shared across bindings (one implementation, one
+hostile-input suite, one fuzz corpus) without pulling transport machinery
+into the model module or wazero into codec consumers.
```
auxesis, line 59: This is a good clarification.

bindings/go/vcffi/README.md
```diff
@@ -0,0 +1,43 @@
+# vcffi — the shared FFI transport codec
```
auxesis, line 1: General note about this README: 

It would be good to have a couple of code examples that show how to use it.

auxesis (approved):
@coderdan love your work, approved in advance. 

Good to merge once small bit of feedback is addressed: 

- I (the human) have left some comments about documentation
- @cipherstash-bot has added feedback about test coverage
- @cipherstash-bot has added feedback from a Golang-specific review

bindings/go/vcffi/README.md
coderdan: Good suggestion. I'll add.

bindings/go/vcffi/ciphertext.go
```diff
@@ -0,0 +1,346 @@
...
+// kind cannot be wired into one direction and missed in the other.
+var leafTags = [...]byte{
+	LeafSingle:   ctSingle,
+	LeafNone:     ctNone,
+	LeafEmptySeq: ctEmptySeq,
+	LeafEmptyMap: ctEmptyMap,
+}
+
+// leafTag maps a classified kind to its wire tag.
+func leafTag(kind LeafKind) (byte, error) {
```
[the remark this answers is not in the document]
coderdan: Resolved in 9f97cbe: no ciphertext tag is 0x00, so zero is now the "not wired" sentinel — `leafTag` refuses `leafTags[kind] == 0` alongside the bounds check, and `leafKind` refuses tag 0 before the scan. `TestLeafTagsHasNoUnwiredKinds` pins both directions plus the table itself having no holes, and `0x00` is rejected on the wire.

bindings/go/vcffi/ciphertext.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+var vcValueLeaves = LeafSet{
+	Classify: func(v any) (LeafKind, []byte, bool) {
+		switch n := v.(type) {
+		case vcvalue.Sealed:
+			return LeafSingle, n, true
+		case vcvalue.SealedNone:
+			return LeafNone, n, true
+		case vcvalue.SealedEmptySeq:
```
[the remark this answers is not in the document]
coderdan: Resolved in 9f97cbe: `TestCipherTextRoundTripsAllLeafKinds` carries all four kinds through a round trip, so no `Classify`/`Make` arm is unexercised, and `FuzzUnmarshalCipherText` gains the `ctEmptySeq`/`ctEmptyMap` seeds.

bindings/go/vcffi/transport.go
```diff
@@ -66,11 +66,22 @@ func eagerCap(n int) int {
 	return min(n, maxEagerCapacity)
 }
 
-var errMalformed = errors.New("vcencrypt: malformed transport bytes")
+// ErrMalformed is returned (sometimes wrapped) by the decoders for any
```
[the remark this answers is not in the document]
coderdan: Resolved in f2c7f7d: dropped the qualifier and said so outright — `ErrMalformed` is returned bare, never wrapped, so `errors.Is` and `==` are equivalent today (with `errors.Is` still the one to prefer).

bindings/go/vcffi/doc.go
```diff
@@ -0,0 +1,44 @@
...
+// binding supplies a [LeafSet] mapping its own leaf types onto the framing,
+// so a leaf sealed by one cipher is a distinct Go type from a leaf sealed by
+// another and can never scan or marshal where the other belongs.
+// [VCValueLeaves] is the set for the vcvalue model types.
+//
+// # Hostile input
+//
+// The decoders treat their input as untrusted: recursion depth, length
+// prefixes and item counts are bounded, strings are UTF-8 validated,
+// duplicate map keys are rejected, and every malformed shape fails with
```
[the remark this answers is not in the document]
coderdan: Resolved in f2c7f7d: "malformed input", plus a paragraph in both doc.go and the README stating that a partially wired `LeafSet` or a `Make` miss fails with its own attributable error — so `errors.Is(err, ErrMalformed)` is a test for bad input, not a total branch over decode failures. `ErrTooDeep` named there too.

bindings/go/vcffi/README.md
```diff
@@ -0,0 +1,43 @@
...
+## Transport, not storage
+
+These bytes are an FFI detail with **no compatibility commitment**. The only
+frozen format is the sealed leaf inside the AEAD envelope (see
+`vcvalue.Sealed` and the Rust crate `vitaminc-aead-value`). Nothing this
+package encodes should ever be persisted.
+
+## The `LeafSet` parameter
+
+The ciphertext codec's structural nodes (`[]any`, `map[string]any`,
```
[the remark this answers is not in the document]
coderdan: Resolved in f2c7f7d: the `LeafSet` section now says the seam stops at the leaves — structural nodes including `vcvalue.Plain` and the `vcvalue.Object` decode shape are shared and not parameterised, a binding brings its own leaves and not its own model — with the reasoning (`Plain` has no key binding, and a second model would fork the encoding this module exists to keep single).

bindings/go/vcffi/README.md
coderdan: Added in f2c7f7d: a Usage section with three worked examples — the `Marshal`/`UnmarshalCipherText` round trip with struct tags, an `Encryptable` controlling its own encoding through the `Encoder` channels, and a second binding's `LeafSet`. All three were compile- and run-checked against the package before landing.

PR #314, opened 2026-09-04
feat(protected): `NonEmpty::with` and `From<integer>` for `NonEmpty`

### LINKED ISSUE #313: Compose and convert `NonEmpty` without re-checking: `NonEmpty::with` and `From<integer>`
## Problem

`NonEmpty<T>` (#291) proves a context carries caller bytes, checked once at construction. Two things a consumer needs are not expressible without giving that proof up or re-checking:

1. **Extending a proven context.** A row-level API derives each field under a fixed context (`nonempty!("users/email")`) and wants the caller to *extend* it with a value known only at the call site — a record id — so the field is bound to its row as well as its column. The result is a pair, and `(A, B)` is empty only when both halves are, so a non-empty head makes the pair non-empty whatever the tail. But the only constructor is `NonEmpty::new`, which needs `IsEmpty` on the pair (so `T: IsEmpty` leaks into every bound) and returns a `Result` for a case that cannot fail. `NonEmpty<T>` deliberately does not implement `IsEmpty`, so nesting the proven head inside a pair does not help either.

2. **Integers.** `IsEmpty` already declares the ten integer types never empty, but an API taking `impl Into<NonEmpty<T>>` cannot accept a bare `42u64`: there is no `From<u64> for NonEmpty<u64>`, and a caller has to write `NonEmpty::new(42u64)?` — a `?` on an infallible check.

## Proposal

- `impl<T> NonEmpty<T> { pub fn with<U>(self, tail: U) -> NonEmpty<(T, U)> }` — pairs a proven head with any tail, checking nothing. Encodes exactly as the bare `(T, U)` would (`NonEmpty` is transparent to `IntoAad` / `IntoPrfContext`), so `nonempty!("users/email").with(42u64)` authenticates the same bytes as `("users/email", 42u64)`.
- `impl From<$int> for NonEmpty<$int>` for the ten types in `never_empty!`.

Both additive; no change to existing encodings.

## Consumer

cipherstash-suite `stack-encrypt`: the leaves are bound on `NonEmpty<T>` directly (no crate-local context marker), and `#[derive(EncryptFrom)]` rows extend their inferred contexts with the caller's via `NonEmpty::with`. The suite pins the `non-empty-with` branch until this ships.

coderdan:
Closes #313.

## What

- `NonEmpty::with<U: MaybeEmpty>(tail) -> NonEmpty<(T, U)>`: pairs a proven head with a tail and keeps the proof, checking nothing. A pair is empty only when both halves are, so a non-empty head makes the pair non-empty whatever the tail, `()` or `""` included. The tail must be `MaybeEmpty` for the same reason `NonEmpty::new` requires it, so an already-encoded `Aad`/`PrfContext` or a nested `NonEmpty` is rejected (pinned by a `compile_fail` doctest). The pair encodes exactly as the bare `(T, U)` would; a test in `vitaminc-aead` pins `nonempty!("users/email").with(42u64)` against `("users/email", 42u64)`.
- `From<$int> for NonEmpty<$int>` for the ten integer types `never_empty!` already covers: `NonEmpty::from(7u64)` / `7u64.into()` with no check.
- **Breaking:** `IsEmpty` is renamed `MaybeEmpty`. The old name read as a marker ("values of this type are empty") when it is a capability ("this value can be asked whether it is empty"); integers implement it and are never empty. The method `is_empty` is unchanged.
- Docs: the struct-level "two ways to build one / no third path" narrative is replaced with one rule (checked once where the type cannot prove non-emptiness, converted freely where it can) listing all four routes. `with` documents that chaining nests left (`a.with(b).with(c)` is `((a, b), c)`), that it only extends rightwards, and three things the tail does not inherit: it is unchecked, an empty tail still frames as a pair, and integer tails encode untagged in AEAD.

No encoding changes.

## Why

A row-level consumer (`stack-encrypt`) derives each field under a fixed context and lets the caller extend it with a record id, so a field is bound to its row as well as its column. Without `with` the only way to build that pair is `NonEmpty::new`, which returns a `Result` for a case that cannot fail. The integer conversion removes a `?` on an infallible check at every call site that passes an id.

The rename came out of review: the `U: MaybeEmpty` bound on `with` only makes sense once the trait name says what it provides rather than what it sounds like.

## Deferred

- `with_head` mirror for layouts with the fixed part on the right (`ContextTag::aad_with`).
- `NeverEmpty` marker trait replacing the hand-listed integer `From` impls.

Both non-breaking to add later.

https://claude.ai/code/session_019VRX1tXygj5bew3YhyK1B6

> **Stack:** #318 (`aad-pieces`) is based on this branch and retargets to `main` once this merges.

### DIFF
--- a/packages/protected/src/lib.rs
+++ b/packages/protected/src/lib.rs
@@ -34,6 +34,12 @@ pub use digest::ProtectedDigest;
 pub use ::digest::InvalidLength;
 pub use equatable::{ConstantTimeEq, Equatable};
 pub use exportable::{Exportable, SafeDeserialize, SafeSerialize};
+/// The name `MaybeEmpty` had before it was renamed: it read as a marker
+/// ("values of this type are empty") when it is a capability ("this value
+/// can be asked whether it is empty"). Bounds and impls written against
+/// `IsEmpty` keep compiling; switch to `MaybeEmpty` at your convenience.
+#[deprecated(note = "renamed to `MaybeEmpty`; the trait is unchanged")]
+pub use non_empty::MaybeEmpty as IsEmpty;
 pub use non_empty::{EmptyError, MaybeEmpty, NonEmpty};
 pub use protected::{flatten_array, Protected};
 pub use usage::{Acceptable, DefaultScope, Scope, Usage};
--- a/packages/protected/src/lib.rs
+++ b/packages/protected/src/lib.rs
@@ -34,7 +34,7 @@ pub use digest::ProtectedDigest;
 pub use ::digest::InvalidLength;
 pub use equatable::{ConstantTimeEq, Equatable};
 pub use exportable::{Exportable, SafeDeserialize, SafeSerialize};
-pub use non_empty::{EmptyError, IsEmpty, NonEmpty};
+pub use non_empty::{EmptyError, MaybeEmpty, NonEmpty};
 pub use protected::{flatten_array, Protected};
 pub use usage::{Acceptable, DefaultScope, Scope, Usage};
 
--- a/packages/protected/src/non_empty.rs
+++ b/packages/protected/src/non_empty.rs
@@ -304,7 +304,22 @@ impl<T> NonEmpty<T> {
     /// .with(42u64)` encodes to the same bytes as `("users/email", 42u64)`.
     /// Chaining nests to the **left**: `a.with(b).with(c)` is `((a, b), c)`,
     /// which encodes differently from `(a, (b, c))`. To match an existing
-    /// tuple layout, pass the whole tail at once: `a.with((b, c))`.
+    /// tuple layout, pass the whole tail at once: `a.with((b, c))`. For a
+    /// layout with the fixed part on the *right*, build the tuple and prove it
+    /// with [`NonEmpty::new`]; `with` only extends rightwards.
+    ///
+    /// Three things the tail does **not** get from the head:
+    ///
+    /// - It is not checked. If an empty tail would be a bug in your domain, an
+    ///   id that must be present say, validate it before pairing; integer ids
+    ///   need no validation because they are never empty.
+    /// - An empty tail does not vanish: `head.with(())` is still a pair and is
+    ///   framed as one, so it does not encode to the same bytes as `head`
+    ///   alone. Do not use an empty tail to mean "absent" — in AEAD `()`, `""`
+    ///   and an empty `Vec<u8>` all frame to the same zero bytes.
+    /// - Its type is not authenticated in AEAD: integers encode as raw
+    ///   little-endian bytes with no tag, so `with(42u64)` and `with(42i64)`
+    ///   produce the same AAD. See the integer note on `IntoAad`.
     pub fn with<U: MaybeEmpty>(self, tail: U) -> NonEmpty<(T, U)> {
         NonEmpty((self.0, tail))
     }
@@ -429,13 +444,13 @@ mod tests {
     }
 
     #[test]
-    fn with_keeps_the_proof_whatever_the_tail() {
+    fn with_pairs_unchecked_and_nests_left() {
         // The head is proven; the tail is not checked, and need not carry
-        // anything — a pair is empty only when both halves are.
+        // anything — a pair is empty only when both halves are. Chaining
+        // nests to the left, as the rustdoc promises.
         let head = nonempty!("users/email");
         assert_eq!(head.with(()).get(), &("users/email", ()));
         assert_eq!(head.with("").get(), &("users/email", ""));
-        assert_eq!(head.with(42u64).get(), &("users/email", 42u64));
         assert_eq!(
             head.with(String::from("acme")).with(7u32).into_inner(),
             (("users/email", String::from("acme")), 7u32)
--- a/packages/protected/src/non_empty.rs
+++ b/packages/protected/src/non_empty.rs
@@ -458,6 +458,20 @@ mod tests {
     }
 
     #[test]
+    fn with_needs_no_bound_on_the_head() {
+        // The point of `with` for a generic consumer: a `NonEmpty<C>` can be
+        // extended without `C: MaybeEmpty` leaking into the caller's bounds.
+        // Adding `T: MaybeEmpty` to `with` would stop this compiling.
+        fn bind_row<C>(column: NonEmpty<C>, row: u64) -> NonEmpty<(C, u64)> {
+            column.with(row)
+        }
+        assert_eq!(
+            bind_row(nonempty!("users/email"), 42).into_inner(),
+            ("users/email", 42u64)
+        );
+    }
+
+    #[test]
     fn integers_convert_without_a_check() {
         assert_eq!(NonEmpty::from(0u8).into_inner(), 0u8);
         assert_eq!(NonEmpty::from(-1i64).into_inner(), -1i64);
--- a/packages/protected/src/non_empty.rs
+++ b/packages/protected/src/non_empty.rs
@@ -26,11 +26,12 @@ use std::borrow::Cow;
 /// `Some("")` and `Some(None)` are. This is the definition a downstream crate
 /// would otherwise have to reconstruct by parsing the encoded bytes.
 ///
-/// [`NonEmpty<T>`] deliberately does **not** implement `MaybeEmpty`, so it
-/// composes only at the outermost position: wrap the whole composite
-/// (`NonEmpty::new(("tenant", "email"))`), not the parts — a proven part
-/// nested inside a larger value would force the outer check to be re-derived
-/// anyway.
+/// [`NonEmpty<T>`] deliberately does **not** implement `MaybeEmpty`, so
+/// [`NonEmpty::new`] checks a composite as a whole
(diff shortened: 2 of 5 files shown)


packages/protected/src/non_empty.rs
```diff
@@ -243,6 +252,28 @@ impl<T> NonEmpty<T> {
...
+    ///
+    /// let column = nonempty!("users/email");
+    /// let row: NonEmpty<(&str, u64)> = column.with(42u64);
+    /// assert_eq!(row.get(), &("users/email", 42u64));
+    /// ```
+    ///
+    /// The pair encodes exactly as the bare `(T, U)` would (`NonEmpty` is
+    /// transparent to the context traits), so `nonempty!("users/email")
+    /// .with(42u64)` authenticates the same bytes as `("users/email", 42u64)`.
+    pub fn with<U>(self, tail: U) -> NonEmpty<(T, U)> {
```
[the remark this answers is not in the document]
coderdan: Split in two, since the branch has moved since this review.

**Tail (`U`):** now bounded on purpose, `with<U: MaybeEmpty>` (b9e97f1). Every other constructor requires the wrapped type to be `MaybeEmpty`, and the rustdoc says wrap the whole composite, not the parts. An unbounded tail was the one route to a `NonEmpty<X>` whose `X` could never have passed `NonEmpty::new` (an already-encoded `Aad`, or a nested `NonEmpty`). A `compile_fail` doctest pins the bound.

**Head (`T`):** stays unbounded, and you are right that nothing pinned it. Added `with_needs_no_bound_on_the_head` in 92bb086: a generic `fn bind_row<C>(NonEmpty<C>, u64) -> NonEmpty<(C, u64)>` that fails to compile if `T: MaybeEmpty` is ever added to `with`.

packages/protected/src/non_empty.rs
```diff
@@ -104,6 +104,15 @@ macro_rules! never_empty {
                 false
             }
         }
+
+        /// An integer is never empty, so it converts without a check: an
+        /// API taking `impl Into<NonEmpty<T>>` accepts a bare integer where
+        /// a string would need [`NonEmpty::new`] or [`nonempty!`](crate::nonempty).
```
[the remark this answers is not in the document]
coderdan: Fixed in b9e97f1. Module header and struct docs no longer say two ways / no third path; the section is now "Building one" with one rule (checked once where the type cannot prove non-emptiness, converted freely where it can) listing literal, dynamic, integer and `with` routes. README qualifies "no implicit conversion" to strings and byte slices.

packages/protected/src/non_empty.rs
```diff
@@ -243,6 +270,59 @@ impl<T> NonEmpty<T> {
...
+    ///   id that must be present say, validate it before pairing; integer ids
+    ///   need no validation because they are never empty.
+    /// - An empty tail does not vanish: `head.with(())` is still a pair and is
+    ///   framed as one, so it does not encode to the same bytes as `head`
+    ///   alone. Do not use an empty tail to mean "absent" — in AEAD `()`, `""`
+    ///   and an empty `Vec<u8>` all frame to the same zero bytes.
+    /// - Its type is not authenticated in AEAD: integers encode as raw
+    ///   little-endian bytes with no tag, so `with(42u64)` and `with(42i64)`
+    ///   produce the same AAD. See the integer note on `IntoAad`.
+    pub fn with<U: MaybeEmpty>(self, tail: U) -> NonEmpty<(T, U)> {
```
freshtonic: **Spec:** issue #313 explicitly proposes `with<U>` with any tail and no check. This bound excludes `Aad`, `PrfContext`, another `NonEmpty`, and arbitrary downstream context types, although the head’s proof already makes `(T, U)` non-empty and this method never evaluates `U::is_empty`. Please remove the tail bound and add a generic no-tail-bound regression test, or amend #313 to specify and justify the narrower API.

packages/protected/src/lib.rs
```diff
@@ -34,7 +34,7 @@ pub use digest::ProtectedDigest;
 pub use ::digest::InvalidLength;
 pub use equatable::{ConstantTimeEq, Equatable};
 pub use exportable::{Exportable, SafeDeserialize, SafeSerialize};
-pub use non_empty::{EmptyError, IsEmpty, NonEmpty};
+pub use non_empty::{EmptyError, MaybeEmpty, NonEmpty};
```
freshtonic, line 43: **Spec / scope:** #313 says both changes are additive. Replacing the public `IsEmpty` export makes every downstream bound and implementation rename, which is an unrelated breaking change. The new name may be clearer, but it should be split into a separately scoped change with a compatibility/deprecation decision, or explicitly added to the originating issue before this PR merges.

freshtonic (changes requested):
## Standards

No documented-standard or maintainability findings. The rename is mechanically consistent and correctly marked as breaking; `MaybeEmpty` better communicates the trait’s capability semantics. Integer conversions are generated from one authoritative list, and `with` thoroughly documents left nesting, transparent encoding, unchecked tails, and AEAD’s untagged integer caveat. CI is green.

## Spec

Changes are required because the implementation diverges from issue #313 in two material ways.

First, #313 specifies `with<U>` for any tail, with no check. Requiring `U: MaybeEmpty` excludes encoded contexts, nested proven contexts, and arbitrary downstream context types even though the method never calls `is_empty` and the proven head alone establishes the result’s invariant.

Second, #313 explicitly says both proposed changes are additive. Removing the public `IsEmpty` name is a separate breaking API change, regardless of whether `MaybeEmpty` is a better name. Please keep the existing name in this PR or split/propose the rename with an appropriate compatibility plan; alternatively amend the originating issue to make the changed contract explicit.

The ten integer conversions and tuple-encoding transparency otherwise match the issue.

tobyhede:
Reviewed — no correctness findings. Two non-blocking notes:

**`with` has no `#[must_use]`** (`packages/protected/src/non_empty.rs:324`)

`with` is builder-style: consumes `self`, returns a new value, mutates nothing. Usually the move catches a discarded call, but `NonEmpty<T>` is `Copy` (line 242), so for `Copy` payloads — including the `&'static str` from `nonempty!` — a bare statement compiles silently:

```rust
let ctx = nonempty!("users/email");
ctx.with(row_id);           // compiles, does nothing
let aad = Aad::from(ctx);   // AAD binds only the field name, not the row
```

Silently narrower AAD than intended, no warning. Not a regression from this PR — nothing else in the file carries `#[must_use]` either — so it's really a case for adding it across `new` / `from_static` / `with` in a follow-up.

**`IsEmpty` → `MaybeEmpty` ships with no transitional alias** (`packages/protected/src/lib.rs:37`)

The old name is gone from the public API entirely, so downstream code at 0.2.0 (`use vitaminc_protected::IsEmpty`, `impl IsEmpty for …`, `T: IsEmpty`) gets a hard "cannot find trait" with no pointer to the new name. One release of:

```rust
#[deprecated(since = "0.3.0", note = "renamed to `MaybeEmpty`")]
pub use non_empty::MaybeEmpty as IsEmpty;
```

would keep existing impls and bounds compiling with a warning that names the replacement. Defensible to skip at 0.x with the `refactor(protected)!` marker doing the changelog work — just flagging it so it's a deliberate call rather than a surprise in a downstream build.

tobyhede (approved):

packages/protected/src/non_empty.rs
coderdan: Agreed, and done in 8cd5146. The bound never evaluated anything and its practical effect was to make a downstream context type implement `MaybeEmpty` purely to be a tail it is never checked as; already-encoded contexts and nested `NonEmpty` values are framed once by the pair, so they are harmless tails. `with<U>` is unbounded again, matching #313. `MaybeEmpty` stays on `NonEmpty::new`, where the check runs.

Regression coverage: the `compile_fail` doctest is now a positive one with a `RecordId` tail that implements nothing, and `with_needs_no_bound_on_the_tail` pins an opaque type and a nested `NonEmpty` as tails. The head-side test from 28ad006 stays.

packages/protected/src/lib.rs
coderdan: Fair on scope. Resolved as additive plus deprecation in 4274a82 rather than by splitting: `pub use MaybeEmpty as IsEmpty` under `#[deprecated]`, so every 0.2.0 bound and impl keeps compiling with a warning that names `MaybeEmpty`, pinned by a test. The alias can be dropped in a later major. #313 is amended with both this and the `with` change so the issue matches what ships.

coderdan:
Both notes taken, in 8cd5146 and 4274a82. `#[must_use]` is on `with`, `new`, `from_static` and `from_static_bytes`, and the `Copy` case you described now warns with a message saying the receiver is unchanged. `IsEmpty` stays as a deprecated alias for `MaybeEmpty`, pinned by a test, so 0.2.0 code compiles with a pointer to the new name.

freshtonic (approved):
The remediation resolves both prior blockers. `NonEmpty::with<U>` now accepts any tail without checking or requiring `MaybeEmpty`, while preserving the tuple encoding, and the deprecated `IsEmpty` re-export keeps existing bounds and downstream impls compiling. The integer conversions remain complete and the additional `#[must_use]` annotations are appropriate. No remaining spec or repository-standards concerns.

PR #320, opened 2026-09-06
test(random): pin SafeRand to the ChaCha20 stream and next_bounded_u32 to its sampling

### LINKED ISSUE #344: random: SafeRand's generator output is untested (try_next_u32/u64 survive replacement with a constant)
Found by `cargo-mutants`. Listed as one line in the inventory (cipherstash/vitaminc#211, Tier 2: "`random/bounded` + `safe_rand` bounds/distribution"); this issue writes it up on its own so the fix has something to close.

## The gap

`packages/random/src/safe_rand.rs` — four mutants survive:

* `SafeRand::try_next_u32` replaced with `Ok(0)` and with `Ok(1)`
* `SafeRand::try_next_u64` replaced with `Ok(0)` and with `Ok(1)`

Every test that drew from `SafeRand` drew from entropy and then asserted only `value < max`, which any constant satisfies. Nothing checks that the generator produces the ChaCha20 stream, so a `SafeRand` that always returned the same number would pass the suite.

Scoping the same run to the file turned up seven more survivors in `next_bounded_u32`, in the `%`, `next_power_of_two` and comparison operators of the rejection loop — the sampling shape is untested for the same reason.

## Fix

Pin the output against a reference for a fixed seed rather than asserting a bound:

* compare `TryRng` word for word and byte for byte with `ChaCha20Rng` seeded identically;
* compare `next_bounded_u32`'s rejection branch with a straightforward re-implementation of the same sampling over the same seed;
* check that every value in the documented output range is reachable.

Two mutants will stay alive on purpose: the `%` replacements inside the rejection loop make it spin forever, so they are observable only as a timeout, not a failure.

Do not pin the two power-of-two-branch behaviours (`max` never returned; panic above `1 << 31`) — those are bugs, tracked in cipherstash/vitaminc#321 and cipherstash/vitaminc#198.

coderdan:
Resolves #344

## What

Tests only. `SafeRand`'s `TryRng` impl is compared word for word and byte for byte with `ChaCha20Rng` for a fixed seed; `next_bounded_u32`'s rejection branch is compared with a reference re-implementation of the sampling over the same seed; a range test checks every value in `0..=max` is reachable.

## Why

The whole-workspace `cargo mutants` sweep run by hand for #318 found four survivors in `safe_rand.rs`: `try_next_u32`/`try_next_u64` replaced with `Ok(0)` or `Ok(1)`. The only tests drew from entropy and asserted `< max`, which any constant satisfies. Scoping the same run to this file showed seven more survivors in `next_bounded_u32`.

Result for `safe_rand.rs`: 14 of 19 mutants caught, 3 unviable, 2 remaining are `%` replacements in the rejection loop that make it spin forever, observable only as a timeout.

## What is not pinned, on purpose

Writing the reference found two behaviours that are bugs, not behaviour to lock in. Filed as https://github.com/cipherstash/vitaminc/issues/321 since fixing them changes what three crates rely on:

- A power-of-two `max` reduces mod `max` and never returns `max`, while the doc and the other branch are inclusive.
- Any `max` above `1 << 31` overflows `next_power_of_two` and panics.

### DIFF
--- a/packages/random/src/safe_rand.rs
+++ b/packages/random/src/safe_rand.rs
@@ -82,6 +82,91 @@ impl SeedableRng for SafeRand {
 #[cfg(test)]
 mod tests {
     use super::SafeRand;
+    use rand::{rngs::ChaCha20Rng, Rng, SeedableRng, TryRng};
+
+    const SEED: [u8; 32] = [7u8; 32];
+
+    /// `SafeRand` is a transparent wrapper: for the same seed it must yield
+    /// exactly the ChaCha20 stream, word for word and byte for byte. Pins the
+    /// `TryRng` impl to real output rather than to "some value", which is
+    /// what a mutant returning a constant would otherwise pass as.
+    #[test]
+    fn try_rng_yields_the_chacha20_stream_for_the_seed() {
+        let mut safe = SafeRand::from_seed(SEED);
+        let mut reference = ChaCha20Rng::from_seed(SEED);
+
+        for _ in 0..4 {
+            assert_eq!(safe.try_next_u32().unwrap(), reference.next_u32());
+        }
+        for _ in 0..4 {
+            assert_eq!(safe.try_next_u64().unwrap(), reference.next_u64());
+        }
+        let mut got = [0u8; 40];
+        let mut want = [0u8; 40];
+        safe.try_fill_bytes(&mut got).unwrap();
+        reference.fill_bytes(&mut want);
+        assert_eq!(got, want);
+        assert_ne!(got, [0u8; 40], "fill_bytes must write the buffer");
+    }
+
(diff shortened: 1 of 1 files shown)


freshtonic (approved):
These tests cover the mutation gaps described by the PR: the `TryRng` methods are pinned to the underlying ChaCha20 stream in a shared mixed sequence, and the bounded reference pins modulus, rejection comparison, rejection consumption, and subsequent stream alignment for valid non-power-of-two bounds. The reachability check is deterministic for its fixed seed. The two known boundary bugs are appropriately excluded rather than encoded as expected behavior and are tracked separately in #321. CI is green; no blocking spec or scope concerns.

PR #323, opened 2026-09-07
fix(random)!: uniform bounded sampling; fix biased permutation keys and out-of-range password indices

### LINKED ISSUE #321: `next_bounded_u32` is inclusive on one branch and exclusive on the other; callers each rely on the wrong one
Found while writing the reference test in https://github.com/cipherstash/vitaminc/pull/320.

## The function

`SafeRand::next_bounded_u32(max)` (`packages/random/src/safe_rand.rs`) and the private duplicate in `packages/random/src/bounded.rs`:

```rust
if max.is_power_of_two() {
    self.0.next_u32() % max              // yields 0..max      (exclusive)
} else {
    let cap = max.next_power_of_two();
    let mut value = self.0.next_u32() % cap;
    while value > max { ... }            // yields 0..=max     (inclusive)
    value
}
```

The doc says "up to and including the given maximum". Only the rejection branch does that. Two further defects on the same lines:

- `max > 1 << 31`: `next_power_of_two` overflows. Debug panics; release returns 0 and the `% cap` then panics on division by zero.
- `bounded.rs` carries the same code (with `& (max - 1)` instead of `%`) and its tests pin the split: `next_bounded(10)` may return 9 (inclusive), `next_bounded(32)` is asserted below 32 (exclusive).

## Callers, each wrong on one branch

**`vitaminc-password`** treats `max` as an exclusive length and indexes `STANDARD_CHARS: [char; 94]`:

| call | branch | draws | effect |
|---|---|---|---|
| `next_bounded_u32(94)` (`Password`) | rejection, inclusive | `0..=94` | `STANDARD_CHARS[94]` is out of bounds: **panic** with probability 1/95 per character |
| `next_bounded_u32(62)` (`AlphaNumericPassword`) | rejection, inclusive | `0..=62` | index 62 is `'~'`: a symbol in an alphanumeric password, 1/63 per character |
| `next_bounded_u32(52)` (`AlphaPassword`) | rejection, inclusive | `0..=52` | index 52 is `'0'`: a digit in an alphabetic password, 1/53 per character |

Separately, `STANDARD_CHARS` reads `'c', 'e', 'e', 'f'`: no `'d'`, and `'e'` at double weight.

**`vitaminc-permutation`** (`key.rs:98`) treats `max` as inclusive, which Fisher-Yates requires: `j = next_bounded_u32(i)` for `i` from `N-1` down to 0. Permitted `N` are 8, 16, 32, 64, 128, so `i` passes through 1, 2, 4, 8, 16, 32, 64. At each of those the power-of-two branch draws `0..i` and element `i` can never stay in place; at `i = 1` the swap is forced. **The generated permutation is not uniform.** `PermutationKey` is key material, so this is the one that matters.

## Suggested fix

One implementation, one meaning. Exclusive is what every std and `rand` API does and what two of three call sites already assume:

```rust
/// Uniform in `0..n`. Panics if `n == 0`.
pub fn next_below(&mut self, n: u32) -> u32 {
    // widen so n == 1 << 31 and n == u32::MAX do not overflow
    let n = u64::from(n);
    let cap = n.next_power_of_two();
    loop {
        let v = u64::from(self.0.next_u32()) % cap;   // % on a power of two, no bias
        if v < n { return v as u32; }
    }
}
```

Then: `password` calls are already right; `permutation` becomes `next_below(i as u32 + 1)`; `bounded.rs` delegates or goes; `next_bounded_u32` is removed or kept as a deprecated inclusive shim over `next_below(max.checked_add(1)...)`. The reference test in https://github.com/cipherstash/vitaminc/pull/320 flips to the new definition with its full input range, including `u32::MAX`. Fix the `'d'` while in `STANDARD_CHARS`.

Renaming is a breaking change to a released crate; keeping the name and changing the power-of-two branch to inclusive instead would silently make `password` index out of range on every call, so the exclusive direction is the safe one.

coderdan:
Closes #321. Stacked on #320 (`test/safe-rand-pins-chacha-stream`); retargets to `main` once that merges.

## What

- **`vitaminc-random`.** One sampler: mask a 32-bit draw to the next power of two, reject while `>= n`, widened to `u64` so `n == 2^32` is expressible, capped at 128 attempts (probability below 2⁻¹²⁸ for a working generator; a CSPRNG that rejects 128 times is broken and should say so). New `SafeRand::next_below(n)`, uniform in `0..n`. `next_bounded_u32(max)` kept, **deprecated**, now inclusive on every branch as its doc always said. `BoundedRng::next_bounded` documented inclusive and delegates to the same sampler; its private duplicate is gone.
- **`vitaminc-permutation`.** Fisher-Yates draws `next_below(i + 1)`. New tests: exact replay of the algorithm on the same seed, and a uniformity histogram over 40,000 keys within 10% of expected per cell. Against the old bound that histogram shows element 0 landing at position 0 **zero** times. README bitwise example re-pinned.
- **`vitaminc-password`.** `next_below(len)` for all three generators; `'d'` restored to `STANDARD_CHARS` (it read `c, e, e, f`). Tests un-ignored and made real: every character in its own set, every standard character reachable, no duplicates, classes in order.

**Breaking:** for a power-of-two `max`, `next_bounded_u32(max)` and `BoundedRng::next_bounded(max)` can now return `max`. Seeded output changes for those inputs, so seeded `PermutationKey`s change.

## Verification

- Workspace green (fmt, clippy, tests, docs), `vitaminc-kms` excluded locally for LocalStack.
- PR-scoped `cargo mutants --in-diff`: 25 mutants, 17 caught, 4 unviable, 0 missed, 0 timeouts after the attempt cap. `safe_rand.rs` alone: 23 mutants, 20 caught, 3 unviable.
- The uniformity test was run against the original sampler body inlined and fails as described.

## Not done here

`Password::into_unprotected_string` is still `unimplemented!()`; `BoundedRng<usize>` is still `unimplemented!()`. Both pre-existing and out of scope.

### DIFF
--- a/packages/password/src/lib.rs
+++ b/packages/password/src/lib.rs
@@ -3,7 +3,7 @@ use vitaminc_random::{Generatable, RandomError, SafeRand};
 
 const STANDARD_CHARS: [char; 94] = [
     'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S',
-    'T', 'U', 'V', 'W', 'X', 'Y', 'Z', 'a', 'b', 'c', 'e', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l',
+    'T', 'U', 'V', 'W', 'X', 'Y', 'Z', 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l',
     'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', '0', '1', '2', '3', '4',
     '5', '6', '7', '8', '9', '~', '`', '!', '@', '#', '$', '%', '^', '&', '*', '(', ')', '_', '-',
     '+', '=', '{', '[', '}', ']', '|', '\\', ':', ';', '"', '\'', '<', ',', '>', '.', '?', '/',
@@ -67,7 +67,7 @@ impl<const N: usize> Generatable for Password<N> {
     fn random(rng: &mut SafeRand) -> Result<Self, RandomError> {
         let mut password: [char; N] = [0x00 as char; N];
         (0..N).for_each(|i| {
-            let char = rng.next_bounded_u32(STANDARD_CHARS.len() as u32);
+            let char = rng.next_below(STANDARD_CHARS.len() as u32);
             password[i] = STANDARD_CHARS[char as usize];
         });
         Ok(Password::new(password))
@@ -78,7 +78,7 @@ impl<const N: usize> Generatable for AlphaNumericPassword<N> {
     fn random(rng: &mut SafeRand) -> Result<Self, RandomError> {
         let mut password: [char; N] = [0x00 as char; N];
         (0..N).for_each(|i| {
-            let char = rng.next_bounded_u32(62);
+            let char = rng.next_below(62);
             password[i] = STANDARD_CHARS[char as usize];
         });
         Ok(Self(Password::new(password)))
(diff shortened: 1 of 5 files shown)


coderdan:
Superseded by #281, which fixes the same three bugs plus the timing side-channel from #198 via a fixed-count Lemire draw. Its `next_below` naming and deprecated inclusive `next_bounded_u32` shim are being carried over to #281.

PR #326, opened 2026-09-07
chore: update chacha20 to 0.10.2 because 0.10.0 was yanked

coderdan:
## Why

`chacha20 v0.10.0` in `Cargo.lock` is yanked on crates.io. Cargo warns about it on every `cargo package` and `cargo publish`, including in the release workflow.

## What

`cargo update -p chacha20`, bumping 0.10.0 to 0.10.2. Lockfile-only change. `chacha20` is a transitive dependency via `rand`'s ChaCha-based RNG; no vitaminc crate depends on it directly. `cargo check --workspace --all-features` passes.

https://claude.ai/code/session_01AthJDZs25JztNCpiYc9smm

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -619,9 +619,9 @@ checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
 name = "chacha20"
-version = "0.10.0"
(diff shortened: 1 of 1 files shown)


auxesis (approved):
Thanks for this @coderdan!

PR #336, opened 2026-09-11
fix(random)!: secret bounds checked at construction, and a fixed-count `NonZeroU16`

### LINKED ISSUE #345: random: a secret zero bound and NonZeroU16's retry loop still branch on secrets, and the password docs promise exact uniformity
## Summary

Reviewing #282 turned up four problems in `vitaminc-random` and `vitaminc-password` that have nothing to do with permutation generation. Two are secret-dependent control flow that the fixed-count draw from #198 was supposed to eliminate; two are public claims stronger than what the code delivers.

## 1. A secret bound of zero can only be rejected by branching on the secret

`packages/random/src/bounded.rs`:

```rust
impl BoundedRng<Protected<u32>> for SafeRand {
    fn next_below(&mut self, n: Protected<u32>) -> Protected<u32> {
        assert!(*n.risky_ref() != 0, "range must be non-zero");
        n.map(|n| below_u32(self, n))
    }
}
```

`0..n` is empty when `n == 0`, so a zero bound has to be rejected somewhere. With the bound wrapped in `Protected<u32>` the only place left is the call, and the check there is a branch on a secret whose outcome is observable as a panic. #198 removed the timing channel inside the draw; this one sits just outside it.

The type system can carry the proof instead. A `Protected<NonZeroU32>` bound moves the check to construction, where the caller decides how to handle a zero and the failure is not a panic on a secret. Every bound the type admits then takes the same path through the draw.

That impl cannot return the bound's own type — `0` is a valid draw and is not a `NonZeroU32` — so `BoundedRng` needs an associated output type, which is a breaking change to the trait.

## 2. `NonZeroU16::random` retries in-stream

`packages/random/src/generatable.rs`:

```rust
fn random(rng: &mut SafeRand) -> Result<Self, RandomError> {
    let mut buf: [u8; 2] = [0, 0];
    rng.fill(&mut buf);
    if let Some(value) = NonZeroU16::new(u16::from_be_bytes(buf)) {
        Ok(value)
    } else {
        // Because a 0 would be an invalid value we must try again (rejection sampling)
        Self::random(rng)
    }
}
```

The same shape #198 removed from `next_bounded_u32`, at p = 2⁻¹⁶ per draw: the number of words consumed from the generator depends on the values drawn. On a secret-seeded generator that count is observable through timing, and it desynchronises any caller that interleaves this with raw draws from a fixed seed.

One bounded draw over `0..u16::MAX` plus one gives the same distribution over `1..=u16::MAX` at a fixed cost of exactly one word.

## 3. The password docs promise uniformity the reduction does not give

`packages/password/README.md`, "Unbiased Character Selection":

> The crate uses bounded random number generation to ensure each character in the character set has an equal probability of being selected, preventing bias that could reduce password entropy.

Equal probability per character is exactly what a rejection-free multiply-high reduction does not deliver. `packages/password/src/lib.rs` has the same claim, and so do the `next_below` docs. The bias is far too small to exploit, but a protocol that needs exact uniformity has to know the term is there, and a crate that documents a guarantee it does not meet is the harder problem.

The figure needs care in both directions. `n / 2⁶⁴` is a valid ceiling but a loose one; the exact distance is `r(n − r) / (n · 2⁶⁴)` with `r = 2⁶⁴ mod n`. For the 94-character default set that is `72 × 22 / (94 × 2⁶⁴)` = 2⁻⁵⁹·⁹³, so a per-character ceiling of 2⁻⁶⁰ is one the shipped default set exceeds.

## 4. Three copies of one draw loop

`Password`, `AlphaNumericPassword` and `AlphaPassword` each have their own copy of the same per-character loop, differing only in which slice of `STANDARD_CHARS` they index. Three copies are three places for a bound to drift, and the narrow sets are the ones where an off-by-one silently shrinks the alphabet.

## Scope

Fixing 1 breaks `BoundedRng` (new associated type), so it wants to land with the rest of the bounded-draw churn from #198 rather than on its own later.

coderdan:
Closes #345.

Split out of #282 at review request (the shuffle PR should carry only the permutation break). Stacked on #281 (targets its branch). #282 will be rebased to sit on #281 without these commits; merge order is #281, this, #282, #283.

## What

Five commits that came out of reviewing #282 but change `vitaminc-random` and `vitaminc-password`, not permutation generation, plus one commit of review fixes on this branch's own work:

- **`BoundedRng` gains `type Output`, and a `Protected<NonZeroU32>` bound impl.** A `Protected<u32>` bound of zero can only be rejected by a check on the secret at the call site. `Protected<NonZeroU32>` moves that check to construction, so every bound the type admits takes the same path through the draw: nothing branches on the secret, and the non-zero assertion in the shared reduction cannot fire. The impl returns `Protected<u32>` (zero is a valid draw), which is why the trait needs an associated output type.
- **`NonZeroU16::random` is one bounded draw.** It used to draw a raw `u16` and recurse on zero: an in-stream retry whose word count depends on the stream, at p = 2⁻¹⁶ per draw. Now `next_below(u16::MAX) + 1`.
- **Password generators share one draw loop.** The three generators were three copies of the same loop; now one `draw_chars` with the character set injected.
- **Docs qualify "uniform" with the reduction's bias bound.** `next_below` is uniform to within `n / 2⁶⁴`; the `NonZeroU16` and password docs now say so instead of promising exact uniformity.
- **Password README matches.** "Unbiased Character Selection" promised equal probability per character; it now states the bound the implementation actually meets.
- **Review fixes on the above.** The per-character ceiling first went in as 2⁻⁶⁰, which the 94-character set exceeds: its exact distance is `72 × 22 / (94 × 2⁶⁴)`, or 2⁻⁵⁹·⁹³. Both the crate doc and the README now say 2⁻⁵⁹, and the README no longer derives that figure from `set.len() / 2⁶⁴` (which is nearer 2⁻⁵⁷) instead of from the tighter `r(n − r) / (n · 2⁶⁴)` form it comes from. Three overstated claims are pulled back to what the code does, and the test that checks the narrow sets now asserts membership by name rather than relying on an out-of-bounds index panic. `NonZeroU16::random`'s distance was also given as *exactly* 2⁻⁶⁴; `2⁶⁴ mod 65535` is 1, so it is `65534 / (65535 · 2⁶⁴)`, just under. The exact `r(n − r) / (n · 2⁶⁴)` form now lives on `BoundedRng`, so both figures derive from one stated formula.

## Breaking changes

- `BoundedRng` gains `type Output`; `next_below` returns `Self::Output`. Downstream implementors must add the associated type. Callers are affected only where the drawn value's type was left to inference: the result no longer has the bound's own type, so such a call may now need an annotation.
- `NonZeroU16::random` yields a different value for a given seed (one bounded draw instead of an in-stream retry).

## Known follow-up

The stream-consumption note on `BoundedRngInclusive` and `next_bounded_u32` ("every call now consumes exactly one 64-bit word, where it previously consumed one or more 32-bit words") documents #281's Lemire change rather than anything on this branch. It is accurate where it sits; moving it to #281 would mean rewriting that branch's history after approval, so it stays here unless you'd rather it moved.

https://claude.ai/code/session_013SXH62m7Jg4GiYbeqdKTR9


packages/password/README.md
```diff
@@ -187,9 +187,9 @@ Passwords are stored in `Protected<[char; N]>`, which ensures:
 - **No accidental leakage**: the opaque `Debug` impl prints only the type name (e.g. `Protected<[char; 16]>`), never the contents
 - **Constant-time operations**: Where applicable, to prevent timing side-channels
 
-### Unbiased Character Selection
+### Near-Uniform Character Selection
 
-The crate uses bounded random number generation to ensure each character in the character set has an equal probability of being selected, preventing bias that could reduce password entropy.
+Each character is one fixed-count bounded draw in `0..set.len()`, so every character of the set is reachable and the number of random words consumed never depends on the values drawn. The draw is not exactly uniform: `SafeRand`'s bounded reduction has a statistical distance from uniform of at most `set.len() / 2⁶⁴`, which for the 94-, 62- and 52-character sets in this crate is at most 2⁻⁶⁰ per character. That bias is far below anything an attacker could exploit, but a protocol that needs exact uniformity must account for it.
```
freshtonic: The `2^-60` ceiling is too strong for the 94-character set. Here `r = 2^64 mod 94 = 72`, so the exact total-variation distance is `r(94-r)/(94*2^64)`, approximately `2^-59.93`—slightly greater than `2^-60`. Please use the promised `2^-59` ceiling (and state the tighter formula); the looser `n/2^64` bound alone only supports less than `2^-57`.

packages/random/src/generatable.rs
```diff
@@ -26,16 +26,17 @@ pub trait Generatable: Sized {
 }
 
 impl Generatable for NonZeroU16 {
+    /// Exactly one bounded draw: an offset in `0..u16::MAX` is drawn with
+    /// [`SafeRand::next_below`] and added to one, so the result is uniform
+    /// over `1..=u16::MAX` to within the `u16::MAX / 2⁶⁴` bias bound
+    /// documented on [`BoundedRng`](crate::BoundedRng) (exactly 2⁻⁶⁴ here).
```
freshtonic: This is not exactly `2^-64`. For `n = 65535`, `2^64 mod n = 1`, so the exact total-variation distance is `(n-1)/(n*2^64)`, strictly below `2^-64`. Please describe it as an upper bound or give the exact formula.

packages/password/src/lib.rs
```diff
@@ -177,19 +180,11 @@ mod tests {
...
-            assert!(s.iter().all(|c| c.is_ascii_alphanumeric()), "{s:?}");
-        }
-    }
-
-    /// The membership tests above would still pass if a narrow set's bound
-    /// shrank by one (dropping `'9'` or `'z'`); this checks every character
-    /// of each narrow set is actually drawn.
+    /// Every character of each narrow set is drawn, and nothing outside it:
+    /// the position lookup is into the *whole* table, so a character past
+    /// the narrow set's end indexes past the tally array and panics. A
```
freshtonic: The PR body says the review fix asserts membership by name, but this comment confirms the test still relies on an out-of-bounds tally index to reject a character outside the narrow set. Please search the relevant narrow slice or assert membership explicitly so failures are intentional and diagnostic.

freshtonic (changes requested):
The fixed-count bounded-draw design and Protected<NonZeroU32> API look sound, but the current head still overstates the statistical guarantees. For the 94-character alphabet, the exact total-variation distance is slightly greater than 2^-60, and the NonZeroU16 distance is strictly less than—not exactly—2^-64. The narrow-set test also still relies on the bounds panic that the PR body says was replaced with an explicit membership assertion. Please correct these public security claims and make the test match the documented review fix.

packages/password/README.md
freshtonic: Non-blocking, on condition that there is a follow up to address it.

packages/password/src/lib.rs
freshtonic: Non-blocking, on condition that there is a follow up to address it.

packages/random/src/generatable.rs
freshtonic: Non-blocking, on condition that there is a follow up to address it.

freshtonic (approved):
Approved - but documentation accuracy issues should be addressed in a follow-up.

coderdan:
> Approved - but documentation accuracy issues should be addressed in a follow-up.

They were already addressed in https://github.com/cipherstash/vitaminc/pull/336/changes/4df5a8ed4a2c74e0665a794046cee9771fc6e41b

Issue #333, opened 2026-09-08
Share the guest ABI scaffolding as a crate with a native cdylib backend


coderdan:
## Decision (2026-09-20)

The second consumer now exists: the credential guest for `stack-profile` and `stack-auth` (ADR-0005, `packages/stack-encrypt/docs/adr/0005-a-separate-credential-guest-for-the-profile-and-auth.md`; sequencing in `docs/plans/stack-encrypt-go-bindings.md`, "Credential guest"). Extract **before** the second guest is written, as its own PR, so the memory-hygiene fixes from [CIP-4111](https://linear.app/cipherstash/issue/CIP-4111/go-bindings-key-hygiene-must-not-depend-on-close-running-lock-and-non) are made once.

* Location: `packages/stack-guest-abi`, a workspace member (so it gets the suite's clippy and fmt gates), `publish = false`.
* Contents: the allocator and `se_alloc`/`se_dealloc` ABI, the buffer registry, the status table, and the `cipherstash_transport` host import with its header micro-format. `WasiHostConnection` stays in the crypto guest; it is ZeroKMS-shaped.
* The status table is **one numbering for both guests**: existing codes keep their values and are still frozen; the credential guest's profile codes append after them. The Go side decodes it once, in the internal package [CIP-4115](https://linear.app/cipherstash/issue/CIP-4115/go-bindings-one-module-at-bindingsgo-with-an-internal-package-for-what) creates.
* The native cdylib backend this issue originally proposed is out of scope for this step; keep the crate shaped so it can be added.

The original text follows.

---

## Summary

The stack-encrypt WASI guest in cipherstash-suite documents itself as "same conventions as the vitaminc guest (`vc_*`)": the buffer registry with zeroize-on-free and registry-checked lengths, the packed `u64` result encoding, the handle table, the status codes, the `catch_unwind` at every export, the hostile-input posture. Today that is shared by convention and copied by hand. The native SDK plan (cipherstash-suite `docs/plans/stack-encrypt-ffi-sdks.md`) needs the same scaffolding to build as a native cdylib for C, C++, Python and the languages after them, and three parts of it are wasm-shaped:

* **Result packing** `ptr << 32 | len` in one `u64` assumes 32-bit pointers and breaks on a native 64-bit target.
* **The session / handle table** is a `thread_local! RefCell`, correct for single-threaded wasm and wrong for a native library called from many threads.
[...]

Issue #315, opened 2026-09-04
Integer AAD is untagged: `0u64` and `None` authenticate identical bytes


coderdan:
## Problem

`IntoAad` for the integer types encodes the raw little-endian bytes with no type tag (`integer_aad!`, deliberately, to match the original `u64` wire format). Two consequences a consumer trips over:

1. **Same-width type confusion.** `7u32.into_aad() == 7i32.into_aad()`. `IntoPrfContext` for the same types *is* tagged (`PrfEncoding::U32` vs `I32`), so a record sealed under a `u32` id and probed under an `i32` id decrypts fine (AAD identical) while every equality/ORE term derived under it silently never matches. Decrypt succeeding tells the caller the context is right, which is exactly wrong.

2. **Zero collides with an empty composite.** `0u64.into_aad()` is eight zero bytes, which is byte-identical to `None::<T>.into_aad()` — `pae([])`, a count of zero. So a ciphertext sealed under `("users/age", 0u64)` opens under `("users/age", None::<u64>)` and vice versa. Again the PRF channel does not collide, so a caller can observe a successful decrypt paired with terms that do not match.

`IsEmpty` says integers are never empty and `From<int> for NonEmpty<int>` (#314) converts them unchecked, so `NonEmpty` does not help here: the collision is in the encoding, not the emptiness.

## Proposal

Tag the integer AAD encoding the way the PRF context already is — a `PrfEncoding`-style label per width and signedness, PAE-framed — so the two channels agree on what distinguishes two contexts. Wire-format change for anyone sealing under a bare integer AAD; `u64` was the only integer context before 0.2.0 widened the roster, so the blast radius is that one type.

Until then stack-encrypt documents the collision (packages/stack-encrypt/src/target/mod.rs, "Contexts") and tells callers not to mix an integer id and an `Option` id under one prefix.

## Consumer

cipherstash-suite `stack-encrypt`: record contexts are `nonempty!("users/age").with(record_id)` with integer ids; the WASI guest accepts raw AAD bytes from Go, where eight zero bytes are now a valid (non-empty) context.

coderdan:
Superseded by #339, which tags integer AAD as part of moving to one canonical context encoding shared by `IntoAad` and `IntoPrfContext`.

Issue #223, opened 2026-06-23
Extract shared PAE (pre-authentication encoding) into a common crate


coderdan:
## Summary

The PASETO **PAE** encoder (`LE64(count) || (LE64(len) || piece)*`) — an injective encoding of a byte-piece sequence — now exists in **two** places:

- `packages/aead/src/aad/pae.rs` (`pub(crate) fn encode -> Aad`), used to frame multi-piece AAD.
- `packages/prf/src/pae.rs` (`pub fn encode -> Vec<u8>`), used to frame multi-field / domain-separated digest inputs.

`vitaminc-prf` reimplemented it rather than depending on `vitaminc-aead`, because the two are sibling crates (prf is the lower-level primitive; the index layer sits on both) and a `prf -> aead` dependency would invert the layering.

## Proposal

Once there's a clear third consumer (or when the searchable-index crate's home is decided), lift PAE into a shared low-level crate that both depend on — candidates: a new minimal `vitaminc-encoding` crate, or an existing shared base. Have `aead` and `prf` delegate to it (aead keeps its thin `-> Aad` wrapper). Move the injectivity test suite with it.

## Why not now

The format is a small, stable spec (PASETO PAE) and the duplication is ~10 lines, so the drift risk is low and it isn't worth blocking current work or churning the `aead` crate (which has in-flight PRs) for. This issue tracks doing it deliberately later.

## Acceptance

- Single PAE implementation + test suite in a shared crate.
- `aead` and `prf` both delegate; no behavioural change to either.

coderdan:
Superseded by #339: the PAE framing was never what drifted; the leaf and composite encoding rules above it were. #339 lifts the whole context encoding, not only the ten-line framer.

PR #338, opened 2026-09-11
feat(aead,prf)!: `AadPiece` is the identity of a context on both derivations

### LINKED ISSUE #335: aead/prf: make the parts view (`AadPiece`) the identity of a context on both derivations, so a runtime context needs no mirror type and `Option` cannot silently diverge
## Summary

`AadPiece` is vitaminc's runtime form of a context: every leaf kind, `List`, and the thing `Descriptor` renders. A consumer that receives a context as *data* rather than as a Rust type — cipherstash-suite's WASI guest today (cipherstash/cipherstash-suite#2210), the C and Python bindings next (#333) — needs three derivations from that one tree: AAD bytes, PRF context bytes, and the empty check to wrap it in `NonEmpty`. `AadPiece` provides one of the three, and the orphan rule stops any downstream crate adding the others. So the guest carries a seven-arm `ContextPart` mirror of `AadPiece`, restates the tuple emptiness rule by hand, and reproduces the PRF tuple framing by delegating leaves to the static impls and calling `PrfContext::pae` itself. It agrees with the static types only because its tests pin it against `nonempty!(..).with(..)`; every further binding would copy the mirror and the tests.

While writing that mirror we found the two derivations already disagree on shape for one type. `IntoAad for Option<T>` encodes `Some(x)` as the one-element list `pae([x])` (and `into_aad_piece` says so: "`Some(x)` the one-element list"). `IntoPrfContext for Option<T>` encodes `Some(x)` as `pae([OPTION_SOME_DOMAIN, x])`. So a Rust row sealed under `Some(7u64)` and a runtime context `List([U64(7)])` produce the same AAD, the same descriptor (`(7u64)`), and different index terms: the ciphertext opens from Go and every equality/range probe silently returns nothing. The parts view claims to be the identity of a context ("the bytes from `IntoAad` are its identity") and is not, on the PRF side. Documenting "don't use `Option` in a context Go must match" is a footgun a caller has to remember; the fix is to make it impossible.

## Proposal

Make the parts view the identity on **both** derivations, stated as a law and pinned by a test:

> For every context type `T`: `T.into_prf_context() == T.into_aad_piece().into_prf_context()`, and `T.into_aad() == T.into_aad_piece().into_aad()` (the latter already holds).

1. **`impl IntoPrfContext for AadPiece`.** Leaves use the same typed encodings the static impls use (`Text` as `str`, `Bytes` as `[u8]`, each integer width as itself); `List` is `PrfContext::pae` of its parts — which is what the `(A, B)` impl produces today. One place says "a list is a tuple".
2. **`impl MaybeEmpty for AadPiece`**: the existing pair/`Option` rule (a list is empty when every part is; integers never are), so `NonEmpty<AadPiece>` is constructible.
3. **`IntoPrfContext for Option<T>` follows the parts view**: `Some(x)` is `pae([x])`, `None` stays `pae([])`. The current `OPTION_SOME_DOMAIN` tag exists "so the crate has exactly one `Some` encoding" shared with `PrfValue`'s `Option` path. That consistency is between a *value* and a *context*, which are never compared, and it costs the parts identity. `PrfValue`'s `Option` encoding is untouched. **Breaking** for any term derived under an `Option` context (`!` + `BREAKING CHANGE:`); cipherstash-suite has no frozen fixture under one (one test constructs `NonEmpty::new(Some("users/email"))` and asserts nothing about its bytes).
4. **The law as a test**: a property over a generated tree of the standard context types (`&str`, `String`, bytes, every integer width, `Option`, pairs, `NonEmpty`) asserting both equalities against `into_aad_piece()`. A future impl that drifts fails here, not in a binding's parity tests.
5. Docs: `AadPiece` is described as the identity of a context on both derivations; `Option` and pairs are described by their parts view (`Some(x)` is the one-element list, `None` the empty one, `(a, b)` the two-element one). A flat n-ary `List` is a legitimate context a Rust caller reaches through `AadPiece::List` directly; `.with()` chains are left-nested (`((a, b), c)`), and both spellings are exact on both sides.

## What it buys

- The guest's `ContextPart` collapses to `FfiValue -> AadPiece`; its mirror, its `MaybeEmpty` restatement and its parity tests go. The guest's wire grammar (string | bytes | int | list) is unchanged.
- A Go plan context `["users/age", 7]` is `nonempty!("users/age").with(7u64)`; `[7]` is `Some(7u64)`; `[]` is `None` — by construction, not by convention. Nothing for a caller in either language to remember.
- C, Python and the languages after them (#333) inherit the same three derivations from the same type.

## Context

- cipherstash/cipherstash-suite#2210 (the guest mirror, and the review finding that surfaced the `Option` divergence).
- `packages/aead/src/aad/piece.rs` (`AadPiece`, `into_aad_piece` docs), `packages/aead/src/aad/mod.rs` (`Option` and pair `IntoAad`), `packages/prf/src/context.rs` (`Option` and pair `IntoPrfContext`, `for_option_some`, private `typed`).
- Siblings: #332 (codec contract), #333 (shared ABI crate).

coderdan:
## Summary

A *context* is the extra data a caller binds to an encrypted value, such as `("users/email", 7u64)`. It is used in two places. The AEAD authenticates it as associated data, so decryption fails under the wrong context. The PRF mixes it into key derivation for searchable index terms, so a probe under the wrong context matches nothing. Today the two encodings are computed separately, and for `Option` they disagreed: a row sealed under `Some(7u64)` and the same context arriving at runtime as the list `[7u64]` decrypted fine and then matched no index term.

This PR makes the parts tree of a context, `AadPiece`, the single source of truth for both. A context assembled at runtime from parts (for example one that crossed the Go or Python FFI boundary as plain data) now produces exactly the same AEAD bytes and the same PRF context as the static Rust value with those parts. The rule is stated in the docs and checked by quickcheck for every built-in context type.

## Changes

**`vitaminc-prf`** (breaking)
- `IntoPrfContext for Option<T>`: `Some(x)` is now `pae([x])`, matching what `IntoAad` already produced. The `OPTION_SOME_DOMAIN` tag is no longer used for contexts. How a `PrfValue` encodes an optional value is untouched.
- `refine` docs now say which collision its domain tag prevents.

**`vitaminc-aead`**
- `impl IntoPrfContext for AadPiece`: leaves use the standard type's own impl, `List` is `PrfContext::pae` of its parts.
- `impl MaybeEmpty for AadPiece`: same rule as the static types, so `NonEmpty<AadPiece>` can be built.
- New `AadPiece::Unit` variant for `()`. It encodes as no bytes for the AEAD and the empty context for the PRF, so `()` follows the rule too, including inside pairs, `Some`, and `.with()` chains. Previously `()` was an empty `Bytes` leaf, which the PRF side frames, so `("x", ())` derived different PRF bytes from its tree than from the value. The variant is additive (the enum is `non_exhaustive`) and changes no existing bytes.
- `Display`: `Unit` prints as `()`. The empty list now prints as `None` so the two stay distinguishable in a log. Existing byte encodings are unchanged.
- New dependency on `vitaminc-prf` (no cycle: prf depends only on protected).
- Docs: the rule is a runnable doctest on `AadPiece`. The type, trait, impl and README docs are rewritten in plainer language.

**Tests**
- Rule tests live in `given_<context>` modules, one expectation per test, with a message on every assert. AAD and PRF sides are checked together by one `is_identity` helper.
- The tree oracle for a list's PRF context frames bytes by hand so it does not share `PrfContext::pae` with the code under test.
- `()` is covered on its own, inside a pair on either side, inside `Some`, and appended with `with`. `[u8; N]` and `&[u8; N]` are covered on the PRF side.

## Breaking change

Any index term derived under an `Option` PRF context changes bytes. Re-derive stored terms whose context was an `Option`. Non-`Option` contexts and `PrfValue` derivations are unaffected. The `feat(prf)!` commit carries the `BREAKING CHANGE:` footer. cipherstash-suite has no frozen fixture under an `Option` context.

`Display` output for the empty list changed from `()` to `None`. This is log output only, not a wire format.

## Verification

Run on the first three commits by the original author, mirroring `.github/workflows/test.yml`: clippy (all features, aws-lc-rs, wasm32), `cargo fmt --check`, `cargo doc` with `-D warnings`, `cargo test` with `vitaminc-kms` against LocalStack, the RustCrypto and `hlist` variants, wasm32 check, `wasm-pack test --node`. All green.

Run after the review-fix commits (`ae22e79` through `210feda`):
- `cargo test --workspace --exclude vitaminc-kms`: all green. `vitaminc-aead` 135 unit tests and 14 doctests (9 ignored, pre-existing), `vitaminc-prf` 56 unit tests. The new `AadPiece` doctest runs and passes.
- `cargo test -p vitaminc-aead --features hlist`: green.
- `cargo clippy --workspace --all-features --all-targets -- -D warnings`: clean.
- `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`: clean.
- `cargo fmt --all --check`: clean.
- `cargo check -p vitaminc-aead --target wasm32-unknown-unknown`: clean.
- Not re-run after the fix commits: `wasm-pack test`, LocalStack KMS tests. CI covers them.

## Related

- Closes #335.
- Refs #339: one canonical context encoding shared by both derivations, superseding #223 and #315.
- Refs cipherstash/cipherstash-suite#2210: the WASI guest's `ContextPart` mirror collapses to `FfiValue -> AadPiece`.
- Refs #333: the C and Python bindings inherit the same three derivations from the same type.

## Review notes

- Start at `packages/aead/src/aad/piece.rs`: the type docs state the rule, the impls sit directly under it, and the `given_*` test modules check it per type.
- `AadPiece::Unit` came from the P2 review comment on the `()` exception. The alternative, changing what `().into_prf_context()` encodes, would change the default context everywhere.
- The `given_*` test layout is new to this repo and intended as the direction for touched tests, not a wholesale rewrite.

### DIFF
--- a/packages/aead/src/aad/piece.rs
+++ b/packages/aead/src/aad/piece.rs
@@ -29,33 +29,35 @@ use vitaminc_protected::MaybeEmpty;
 
 use super::{Aad, IntoAad};
 
-/// One part of a context, or a PAE-framed list of parts.
+/// One part of a context, or a list of parts.
 ///
-/// Built by [`IntoAad::into_aad_piece`]; encodes through [`IntoAad`]
-/// to the same bytes the source value does, and through
-/// [`IntoPrfContext`] to the same PRF context, so it can stand in for the
-/// value anywhere a context is taken (the law is [below](#the-parts-view-is-the-identity-of-a-context)).
-/// A list encodes in one allocation however deep the tree.
-/// [`Display`](fmt::Display) renders it injectively
-/// — `("users/email", 7u64)` — and [`leaves`](Self::leaves) walks the parts
-/// in encoding order for a consumer building its own rendering or binding.
+/// [`IntoAad::into_aad_piece`] builds one from any context. The tree is a
+/// stand-in for the value it was built from: it encodes to the same AAD
+/// bytes through [`IntoAad`] and to the same PRF context through
+/// [`IntoPrfContext`], so it can be passed anywhere a context is expected
+/// (see [the rule below](#a-parts-tree-is-the-same-context-as-the-value-it-came-from)).
+/// On the AAD side a list is written into a single buffer, however deeply
+/// it nests. [`Display`](fmt::Display) renders a tree so that different
+/// trees never print the same, for example `("users/email", 7u64)`, and
+/// [`leaves`](Self::leaves) walks the parts in encoding order for a caller
+/// that wants to render or bind them itself.
 ///
-/// Integers keep their source type as their variant, so every tree
-/// expressible here encodes without truncation or failure: `U8(7)` is one
-/// byte and `U64(7)` is eight, exactly as `7u8` and `7u64` are.
+/// Integers keep their source type as their variant, so every tree that
+/// can be built here encodes without truncation: `U8(7)` is one byte and
+/// `U64(7)` is eight, exactly as `7u8` and `7u64` are.
 ///
-/// `PartialEq` is tree identity, not encoding identity. `Text("ab")` and
-/// `Bytes(b"ab")` compare unequal, as do `U64(7)` and `I64(7)`, though each
-/// pair encodes to the same bytes. To compare what the AEAD authenticates,
-/// compare `into_aad().as_bytes()`.
+/// `PartialEq` compares trees, not encodings. `Text("ab")` and
+/// `Bytes(b"ab")` are unequal, as are `U64(7)` and `I64(7)`, even though
+/// each pair encodes to the same bytes. To compare what the AEAD
+/// authenticates, compare `into_aad().as_bytes()`.
 ///
-/// # The parts view is the identity of a context
+/// # A parts tree is the same context as the value it came from
 ///
-/// A context feeds two derivations: the AEAD's associated data
-/// ([`IntoAad`]) and a PRF's domain-separation context
-/// ([`IntoPrfContext`]). [`AadPiece`] implements both, and the law every
-/// context type upholds is that each derivation of the value equals the
-/// same derivation of its parts view:
+/// A context is used in two places: as the associated data an AEAD
+/// authenticates ([`IntoAad`]) and as the context a PRF mixes into its key
+/// derivation ([`IntoPrfContext`]). [`AadPiece`] implements both, and every
+/// built-in context type follows one rule: converting the value gives the
+/// same result as converting its parts tree, on both sides.
 ///
 /// ```rust
 /// use std::borrow::Cow;
@@ -69,7 +71,8 @@ use super::{Aad, IntoAad};
 /// );
 /// assert_eq!(x.into_aad_piece().into_prf_context(), x.into_prf_context());
 ///
-/// // So a list assembled at runtime *is* the static value with those parts.
+/// // A list built at runtime is therefore the same context as the
+/// // static value with those parts.
 /// let runtime = AadPiece::List(vec![
 ///     AadPiece::Text(Cow::Borrowed("users/email")),
 ///     AadPiece::U64(7),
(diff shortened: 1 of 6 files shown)


packages/aead/src/aad/piece.rs
```diff
@@ -44,6 +49,37 @@ use super::{Aad, IntoAad};
...
+/// across an FFI boundary, say — is the *same* context as the static Rust
+/// value with those parts, on both sides, and needs no type of its own:
+/// `Some(x)` is the one-element list, `None` the empty list, `(a, b)` the
+/// two-element list, and `nonempty!(a).with(b).with(c)` the left-nested
+/// `((a, b), c)`. A flat list of three or more parts is a context too,
+/// reachable from Rust through [`AadPiece::List`] directly. The one
+/// exception is `()`: its PRF context is the *empty* context by definition
+/// (no encoding at all), while its parts view is an empty `Bytes` leaf,
+/// which the PRF side encodes as typed empty bytes. `()` is never a
+/// context a caller proves non-empty, so nothing built from parts meets
```
coderdan: **[P2] Preserve unit identity inside nonempty composite contexts**

The `()` exception also affects valid nonempty contexts such as `NonEmpty::new(("users/email", ())).unwrap()` and `nonempty!("users/email").with(())`. Both constructions succeed, but the direct PRF encoding frames unit as zero bytes while `into_aad_piece().into_prf_context()` frames it as typed empty bytes. Their AAD remains identical, so this preserves the silent Rust/runtime index mismatch that #335 requires this PR to eliminate.

Represent unit distinctly in `AadPiece`, with empty AAD and PRF encodings and `MaybeEmpty::is_empty() == true`, and map `()` to that variant. This preserves existing non-Option PRF context bytes while keeping unit distinct from empty byte slices. Add regression coverage for unit inside pairs, options, and `NonEmpty`.

Verified at `5a9db03`: the existing `vitaminc-aead` and `vitaminc-prf` suites pass. Temporary regression tests fail PRF identity for the pair above, `.with(())`, and `.with(Some(()))`, after confirming that nonempty construction and AAD equality succeed. The empty-byte-slice control passes.

coderdan:
Reviewed `191775d`. No standards findings or implementation defects found; two documentation corrections:

- **Migration notice:** The PR description says non-`Option` contexts are unaffected. Tuples such as `("tenant", Some(7u64))` also change bytes. Include contexts containing `Some(...)` in the re-indexing guidance.
- **[context.rs:59](https://github.com/cipherstash/vitaminc/blob/191775d4012c125dc5a52d1b64a6cebf738391e0/packages/prf/src/context.rs#L59):** Removing `refine`'s domain produces two PAE parts; `for_map_entry` produces three, so the claimed collision cannot occur. Use the collision with `(parent, component).into_prf_context()` instead.

Validation: 185 unit tests, the compile-fail suite, 15 doctests, and reproductions confirming both findings passed. The documented `()` exception remains deferred to #339.

calvinbrewer (approved):

PR #341, opened 2026-09-12
feat(context)!: one canonical context encoding shared by AAD and PRF derivations

### LINKED ISSUE #339: aead/prf: one canonical context encoding shared by both derivations (supersedes #223, #315)
Supersedes #223 and #315.

## Summary

A context feeds two derivations, the AEAD's associated data (`IntoAad`) and the PRF's domain-separation context (`IntoPrfContext`), and they are produced by two unrelated trait hierarchies in two crates with no shared specification. #223 noticed the PAE framing is duplicated. That duplication is ten identical lines and has never drifted. What has drifted, three times now, is the layer above it: the rule for how a leaf is encoded and how a composite is framed.

| Shape | `IntoAad` | `IntoPrfContext` |
|---|---|---|
| `&str`, bytes | raw bytes | `pae([CONTEXT_VALUE_DOMAIN, tag, bytes])` (typed) |
| integers | raw little-endian bytes, untagged | typed, per width and signedness |
| `()` | zero bytes (parts view: `AadPiece::Unit` after #338) | `PrfContext::default()` |
| `Some(x)` | `pae([x])` | was `pae([OPTION_SOME_DOMAIN, x])`; `pae([x])` after #338 |
| `None`, `(a, b)` | `pae([])`, `pae([a, b])` | same |
| derived (`for_map_entry`, …) | `vitaminc/aead/…` domains | `vitaminc/prf/…` domains |

Every divergence found so far is a row in that table, and each presents the same way: decrypt succeeds, index terms silently miss.

- #315: integers untagged on the AAD side, so `7u32` and `7i32` authenticate the same bytes and `0u64` collides with `None`; the PRF side keeps them apart.
- #335 / #338: `Some` carried a domain tag on the PRF side only, so a runtime one-element list matched the AAD and descriptor of `Some(x)` and none of its terms.
- #338 review: `()` is the empty PRF context by definition but was an empty typed-bytes leaf in the parts view, so `("x", ())` (non-empty by the pair rule, so `NonEmpty`-provable) broke the parts-view law on the PRF side. #338 now fixes this with an `AadPiece::Unit` variant, but it is a fourth hand-written encoding rule that had to be discovered by review.

#338 states the law `x.into_prf_context() == x.into_aad_piece().into_prf_context()` and pins it by quickcheck over every built-in type. That catches the next drift at test time. It does not prevent it: any new leaf or composite type still gets two hand-written encodings that must be kept in agreement by a test someone remembers to write.

## Design

Settled 2026-09-11 after a design review. Make the law hold by construction: one canonical encoding of the parts tree, in one place, that both derivations are views of.

### 1. A new crate, `vitaminc-context`

Sits below `vitaminc-aead` and `vitaminc-prf` and depends only on `vitaminc-protected` (for `MaybeEmpty` / `NonEmpty`, which stay where they are). It owns everything that decides the bytes of a context:

- **PAE framing** `LE64(count) || (LE64(len) || piece)*` and its injectivity tests, moved from both crates.
- **`ContextPiece<'a>`**, the parts tree, which is today's `AadPiece` renamed. Leaves: `Text`, `Bytes`, `Unit`, one variant per integer width and signedness, `List`, plus one new leaf `Encoded` (see 4). `Display` stays injective; `Unit` renders `()`, an empty list renders `None`.
- **The typed leaf encoding**, one for both sides: `pae([VALUE_DOMAIN, type_tag, bytes])` with tags renamed from `vitaminc/prf/...` to neutral names (`vitaminc/context/value/v1`, `vitaminc/context/utf8/v1`, `vitaminc/context/u64-le/v1`, ...). Text, bytes and every integer type carry a tag. This is #315's fix.
- **`Context<'a>`**, the encoded bytes. `Aad` and `PrfContext` are the same `Cow<[u8]>` today and become this one type.
- **The derived-context methods**, one set on `Context`: `for_map_entry`, `for_leaf`, `for_sequence_element`, `for_empty_sequence`, `for_empty_map`, `for_none`, `for_option_some` and `refine`, with domains renamed to `vitaminc/context/...`. Today AAD and PRF each have their own set with their own domains; a derived context describes the plaintext's shape, not the primitive, so it gets one encoding like any other value. PRF's `for_map_entry` goes away and `for_option_some` becomes public next to `for_none`.

### 2. One implementable trait

```rust
pub trait IntoContext<'a> {
    fn into_context(self) -> ContextPiece<'a>;
}

pub trait IntoAad<'a>: IntoContext<'a> {
    fn into_aad(self) -> Aad<'a>;
}

impl<'a, T: IntoContext<'a>> IntoAad<'a> for T {
    fn into_aad(self) -> Aad<'a> {
        self.into_context().encode()
    }
}
// and the same blanket for IntoPrfContext
```

A context type is written once, as its parts. `IntoAad` and `IntoPrfContext` keep their names so existing bounds (stack-encrypt's `IntoAad + IntoPrfContext`) still compile, but the blanket impls mean nobody can implement either by hand, so no hand-written encoding can diverge again. The `into_aad_piece` provided method from #318 goes away; `into_context` is the only thing to write.

Rejected: a subtrait with an implementable `into_aad`. That would have the consumer write the tree and the bytes separately with nothing tying them together, which is the split #338 patched. Deferred: marker subtraits without a blanket, which would allow per-side opt-in; nothing needs that today and it is a small later change.

### 3. Byte-identical derivations

For every context `x`, `x.into_aad().as_bytes() == x.into_prf_context().as_bytes()`, and both equal the canonical encoding of the tree. No domain prefix at either boundary.

Why identical rather than domain-separated:
- There is no security reason to separate them. AAD is public input to the AEAD and the PRF context is input to a keyed function. Different primitives, different keys, so equal bytes create no cross-protocol risk.
- It makes "a parts tree is the same context as the value it came from" literally true at the byte level, which is what the docs already claim.
- Anything that holds only the AAD bytes can probe an index with them directly instead of rebuilding the tree first.
- A domain prefix would be 16 more bytes on every context and one more thing to explain, for nothing.

For the record, domain separation would not have broken index matching on its own. stack-encrypt always derives both the index term and the probe by calling `into_prf_context()` on the same typed context, so as long as both go through the shared encoder they match regardless of any prefix. The mismatches in #315, #335 and the `()` case came from two independently hand-written structure rules, which this design removes either way. Identical bytes is the simpler design, not a correctness requirement, and nobody should later read a domain prefix as a bug.

### 4. Raw bytes: two entry points, both explicit

- `Context::from_encoded(bytes)` replaces `from_slice` and `new_owned`. It means "bytes this encoder already produced" and exists for storage and FFI round-trips. An encoded `Context` implements `IntoContext` (the blanket impls need it, and stack-encrypt re-passes derived contexts as values) by becoming a `ContextPiece::Encoded` leaf, which writes its bytes verbatim, untagged, and is empty iff the bytes are. `Encoded` is the only untagged leaf and only `from_encoded` and the derived methods produce it.
- `Context::pae(&[&[u8]])` stays public. stack-encrypt frames its own domains with it (`stack-encrypt/leaf`, `stack-encrypt/sem/equality/v1`, ...) and those never start with our tag prefix, so they cannot collide with a typed leaf. Whether bindings may build raw frames is a #333 question.

Today's `from_slice(b"7")` and `"7".into_aad()` produce different bytes, so keeping `from_slice` would keep an untyped second encoding around. It goes.

### 5. `()` stays its own leaf

`ContextPiece::Unit` (from #338) encodes to zero bytes on both sides and `Context::default()` stays zero bytes as the "no context" value. Inside a composite it frames a zero-length piece. No collision is possible because typed leaves are never zero-length.

## Breaking

The wire format changes for every context on both sides: AAD gains typed framing and integer tags, PRF gets the renamed tags and the unified derived-context domains. No stored ciphertexts or index terms exist in vitaminc or cipherstash-suite; the only things sensitive to the change are test pins (`stack-encrypt/tests/term_bytes.rs`, `tests/frozen_bytes.rs`, `leaf_aad_bytes_are_pinned` in `cipher.rs`, descriptor doctests, and the vitaminc pins in `aad/mod.rs` and `prf/context.rs`). No compatibility path and no `SealedValue::FORMAT_VERSION` bump. Sequence this before the C and Python bindings in #333 copy the current split.

## Delivery

1. **One vitaminc PR.** New crate; `vitaminc-aead` and `vitaminc-prf` delegate to it; `Aad`, `PrfContext` and `AadPiece` kept as deprecated aliases so cipherstash-suite compiles against the release that ships the new bytes without also doing the rename. Released as one `feat!` in the version group.
2. **One cipherstash-suite PR.** Update pins; rewrite `WorkspaceId`'s hand-written `IntoAad` (today `Aad::new_owned(raw bytes)`) as an `IntoContext` impl yielding a `Text` leaf, and `UserCode` likewise; rename imports; the WASI guest's `FfiValue` mapping targets `ContextPiece` directly.
3. **Follow-up vitaminc release** removes the aliases.

## Acceptance

- One PAE implementation, one typed leaf encoding and one set of derived-context methods, all in `vitaminc-context`, with the injectivity and framing tests moved with them.
- `IntoContext` is the only trait a context type implements; `IntoAad` and `IntoPrfContext` are blanket impls over it, and a hand-written impl of either fails to compile.
- `x.into_aad().as_bytes() == x.into_prf_context().as_bytes()` for every built-in context type and for every generated tree, pinned by one quickcheck over `ContextPiece`.
- `x.into_context().into_aad() == x.into_aad()` (the #338 law) holds for every type by construction, not by a per-type quickcheck.
- #315's collisions (`7u32` vs `7i32`, `0u64` vs `None`) are gone on the AAD side.
- `Encoded` is the only untagged leaf; `from_slice` and `new_owned` no longer exist.
- cipherstash-suite's WASI guest maps `FfiValue` straight onto `ContextPiece`, with no mirror type.

## Context

- #223 (shared PAE crate, superseded), #315 (untagged integer AAD, superseded), #335 / #338 (parts-view law), #333 (shared ABI crate, sequenced after this).
- `packages/aead/src/aad/pae.rs`, `packages/aead/src/aad/piece.rs`, `packages/aead/src/aad/mod.rs`, `packages/prf/src/context.rs`.

coderdan:
Closes #339. Supersedes #223 and #315.

## What

A context is used in two places: as the associated data an AEAD authenticates (`IntoAad`) and as the context a PRF derives under (`IntoPrfContext`). Until now those were two unrelated trait hierarchies in two crates, and every built-in type carried two hand-written encodings that had to be kept in agreement by a test someone remembered to write. They drifted three times (#315, #335, the `()` leaf found in #338 review), and each time it presented as decrypt-succeeds-terms-miss.

This PR makes the law hold by construction, as designed in #339:

- **New crate `vitaminc-context`** (depends only on `vitaminc-protected`). It owns PAE framing, the parts tree `ContextPiece` (the old `AadPiece`, renamed, plus an untagged `Encoded` leaf), the typed leaf encoding, the `Context` bytes type (one type replacing both `Aad` and `PrfContext`), and the single set of derived-context methods (`for_map_entry`, `for_leaf`, `for_sequence_element`, the three markers, `for_option_some` now public, and `refine`) under `vitaminc/context/...` domains.
- **`IntoContext` is the only implementable trait.** `IntoAad` (in `vitaminc-aead`) and `IntoPrfContext` (in `vitaminc-prf`) are supertraits of it with blanket impls only. A hand-written impl of either no longer compiles. The `into_aad_piece` provided method from #318 is gone; `into_context` is the one thing to write.
- **Byte-identical derivations.** `x.into_aad().as_bytes() == x.into_prf_context().as_bytes()` for every context, with no domain prefix at either boundary. Pinned by one quickcheck over every `ContextPiece` tree (`packages/aead/src/aad/mod.rs`), using a new optional `arbitrary` feature on the context crate.
- **Typed leaves on both sides.** Text, bytes and every integer type encode as `PAE(b"vitaminc/context/value/v1", type_tag, bytes)`. `7u32` and `7i32`, and `0u64` and `None`, no longer collide as AAD. `()` stays its own zero-byte leaf.
- **Raw bytes are explicit.** `Aad::from_slice` and `new_owned` are gone. `Context::from_encoded` is the one entry point for bytes this encoder already produced (storage, FFI round-trips); it becomes the `Encoded` leaf inside a composite, written verbatim. `Context::pae` stays public for crates that frame their own domains.
- **Deprecated aliases** `Aad`, `AadPiece` and `PrfContext` (`since = "0.5.0"`) so cipherstash-suite compiles against this release before doing the rename. `vitaminc-aead` no longer depends on `vitaminc-prf` (it is now a path-only dev-dependency for the identity test).

## Breaking

The wire format changes for every context on both sides. No stored ciphertexts or index terms exist in vitaminc or cipherstash-suite; the sensitive things are test pins, which are updated here. No compatibility path, per the issue.

Call-site rule that fell out of the change: anywhere one side passed `&[u8]` and the other `Aad::from_slice` of the same bytes, both must now use `Context::from_encoded`, because a typed `Bytes` leaf and an `Encoded` leaf are different contexts. Two such sites existed (`encrypt/src/cipher.rs` element decrypt, `tests/map_access.rs`).

## Go binding

The WASI guest passes host AAD bytes through `Context::from_encoded` (the FFI round-trip entry point; whether bindings build `ContextPiece` trees directly is a #333 question). The committed `wasm/vitaminc_guest.wasm` and `testdata/cross_lang.bin` are regenerated per the README recipe, because the fixture is a map and the map-entry derivation changed. Go is not installed on my machine, so the Go tests are verified by CI only; the guest crate's own clippy (native and wasm32-wasip1), docs and tests pass.

## Verified locally

- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps`
- `cargo test --workspace --all-features`: 700 passed, 0 failed
- Guest crate: fmt, clippy (native + wasm32-wasip1), doc, test

## Follow-ups (not here)

1. cipherstash-suite PR: update pins, rewrite `WorkspaceId` / `UserCode` as `IntoContext` impls yielding `Text` leaves, rename imports, map the WASI guest's `FfiValue` straight onto `ContextPiece`.
2. A later vitaminc release removes the deprecated aliases.


coderdan:
Review by **gpt-6-astra**, via Codex, through to commit `e45e2f4c0d749efbf5bd2af059885b1ff0acd50a`.

1. **[P2] Restore the deprecated `Aad` re-export** — [packages/encrypt/src/lib.rs:13](https://github.com/cipherstash/vitaminc/blob/e45e2f4c0d749efbf5bd2af059885b1ff0acd50a/packages/encrypt/src/lib.rs#L13-L15). Removing this export breaks `vitaminc_encrypt::Aad` and `vitaminc::encrypt::Aad`, contrary to #339's deprecated-alias transition. A separate consumer probe importing `vitaminc_encrypt::Aad` fails with `E0432`; the workspace and CI pass because their callers already use `Context`. Restore the alias alongside `Context` and add a compiling doctest using the old import, with deprecation warnings allowed.

2. **[P2] Marker tests now pass for the wrong reason** — [packages/encrypt/src/cipher.rs:1629](https://github.com/cipherstash/vitaminc/blob/e45e2f4c0d749efbf5bd2af059885b1ff0acd50a/packages/encrypt/src/cipher.rs#L1629-L1639) and [packages/encrypt/src/cipher_static.rs:375](https://github.com/cipherstash/vitaminc/blob/e45e2f4c0d749efbf5bd2af059885b1ff0acd50a/packages/encrypt/src/cipher_static.rs#L371-L379). The fixtures use `Context::from_encoded(...)`, while verification uses typed text/bytes contexts. Authentication therefore fails before reaching the nonempty-payload check. A separate reproduction confirms that both fixtures reject even valid empty payloads. Derive fixture contexts from `"context".into_aad()` and `aad.into_aad()`, respectively, and include an empty-payload success assertion so the tests prove authentication succeeded.

No documented coding-standard violations found. Tests across `vitaminc-context`, `vitaminc-aead`, `vitaminc-prf`, `vitaminc-encrypt`, `vitaminc-hmac`, and `vitaminc-aead-value` passed with all features enabled. Both findings above were reproduced separately from that passing suite.

calvinbrewer (approved):

